PR #1, opened 2023-12-04
Init

raghukiran1224:


README.md
```diff
@@ -0,0 +1,104 @@
...
+This repo aims to create basic tuning scripts with support for specific models, assumes the use of Hugging Face `SFTTrainer` and PyTorch FSDP only. Our approach to tuning is:
+1. Models are loaded from Hugging Face `transformers` or the `foundation-model-stack` -- models are either optimized to use `Flash Attention v2` directly or through `SDPA`
+2. Hugging Face `SFTTrainer` for the training loop
+3. `FSDP` as the backend for training
+
+## Installation
+
+```
+pip install -r requirements.txt
+pip install -U datasets
```
nairbv, line 12: is there a reason datasets isn't just part of requirements.txt?

README.md
```diff
@@ -0,0 +1,104 @@
...
+        "Write a response that appropriately completes the request.\n\n"
+        "### Instruction:\n{instruction}\n\n### Response:"
+    ),
+}
+
+def format_alpaca_fn(example):
+    prompt_input, prompt_no_input = PROMPT_DICT['prompt_input'], PROMPT_DICT['prompt_no_input']
+    output = prompt_input.format_map(example) if example.get("input", "") != "" else prompt_no_input.format_map(example)
+    output = f"{output} {example['output']}"
+    return {"output": output}
```
nairbv, line 37: I'm not seeing how we handle `IGNORE_INDEX` / `-100` for the parts of the prompt that should be ignored in the backward pass

tuning/config/configs.py
```diff
@@ -0,0 +1,42 @@
+from dataclasses import dataclass, field
+from typing import Dict, Optional
+import transformers
+
+DEFAULT_CONTEXT_LENGTH=4096
+DEFAULT_OPTIMIZER="adamw_torch"
+
+IGNORE_INDEX = -100
```
nairbv, line 8: appears to be unused

tuning/config/configs.py
```diff
@@ -0,0 +1,42 @@
...
+DEFAULT_PAD_TOKEN = "[PAD]"
```
nairbv, line 9: is this a valid token for the tokenizer used by llama? if not maybe we should use the unk token?

tuning/config/configs.py
```diff
@@ -0,0 +1,42 @@
...
+class DataArguments:
+    data_path: str = field(default=None, metadata={"help": "Path to the training data in JSONL format."})
+    response_template: str = field(default=None, metadata={"help": "Response template, separator to train on completions only"})
+    dataset_text_field: str = field(default=None, metadata={"help": "Training dataset text field"})
+
+
+@dataclass
+class TrainingArguments(transformers.TrainingArguments):
+    cache_dir: Optional[str] = field(default=None)
+    optim: str = field(default=DEFAULT_OPTIMIZER)
```
nairbv, line 32: looks like this `optim` field is currently unused

README.md
raghukiran1224: yes, there was a version issue with caching. Replaced `requirements.txt` with the correct version.

README.md
raghukiran1224: This is being done in the HF codebase based on a `response_template`. Does the assumption of `IGNORE_INDEX` being the default `-100` pose issues? I do see that their `DataCollator` is exposing this in API with default of `-100`. We can fix this as a parameter based on the model:

Relevant loc: https://github.com/huggingface/trl/blob/main/trl/trainer/utils.py#L178

tuning/config/configs.py
raghukiran1224: to fix the above comment, I have added the `DataCollator` constructor to take the `ignore_index` parameter and using the default one now. We have to change this based on the model.

@lchu-ibm -- another potential model specific code.

README.md
```diff
@@ -0,0 +1,104 @@
...
+1. Models are loaded from Hugging Face `transformers` or the `foundation-model-stack` -- models are either optimized to use `Flash Attention v2` directly or through `SDPA`
+2. Hugging Face `SFTTrainer` for the training loop
+3. `FSDP` as the backend for training
+
+## Installation
+
+```
+pip install -r requirements.txt
+pip install -U datasets
+pip install -e .
```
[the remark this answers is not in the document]
raghukiran1224: oh, I had one.. adding one!

lchu-ibm:
@raghukiran1224 

added prompt tuning in https://github.com/foundation-model-stack/fms-hf-tuning/pull/1/commits/9f1c917be7c71397527196389f5a94a853c5f74e

README.md
```diff
@@ -0,0 +1,140 @@
+# FMS HF Tuning
+
+This repo aims to create basic tuning scripts with support for specific models, assumes the use of Hugging Face `SFTTrainer` and PyTorch FSDP only. Our approach to tuning is:
```
nairbv, line 3: some suggestions:
s/aims to create/provides a/
s/, assumes the use of/. The repo relies on/
s/ only//

README.md
```diff
@@ -0,0 +1,140 @@
...
+1. Models are loaded from Hugging Face `transformers` or the `foundation-model-stack` -- models are either optimized to use `Flash Attention v2` directly or through `SDPA`
```
nairbv, line 4: for `foundation-model-stack` might be better to either use the pypi name (ibm-fms) or link to the github repo

README.md
```diff
@@ -0,0 +1,140 @@
...
+
+
+## Supported Models
+
+Current supported and tested models are `Llama2` (7 and 13B configurations have been tested) and `GPTBigCode`.
+
+## Training
+
+### Single GPU
+```python
```
nairbv, line 57: this should be bash not python, below is getting syntax errors on `--` and `$`

README.md
```diff
@@ -0,0 +1,140 @@
...
+--logging_steps 1  \
+--include_tokens_per_second  \
+--packing False  \
+--response_template "\n### Response:"  \
+--dataset_text_field "output" 
+
+```
+
+### Multiple GPUs with FSDP
+```
```
nairbv, line 84: might want to add `bash` here too

README.md
```diff
@@ -0,0 +1,140 @@
...
+--fsdp_config tuning/config/fsdp_config.json \
+--include_tokens_per_second \
+--packing False \
+--response_template "\n### Response:" \
+--dataset_text_field "output"
+```
+
+The above is an example. We would need to tune parameters depending on the model size, data size. The above example has been validated on 8 x A100 80GB.
+
+For `GPTBigCode` models, Hugging Face has enabled Flash v2 and one can simply replace the `'LlamaDecoderLayer'` with `'GPTBigCodeBlock'` for proper sharding of the model.
```
nairbv, line 115: replace in `tuning/config/fsdp_config.json`? might want to specify where the change happens since I was expecting to see 'LlamaDecoderLayer' in the script above.

README.md
```diff
@@ -0,0 +1,140 @@
...
+--logging_steps 1 \
+--fsdp "full_shard auto_wrap" \ 
+--fsdp_config tuning/config/fsdp_config.json \
+--include_tokens_per_second \
+--packing False \
+--response_template "\n### Response:" \
+--dataset_text_field "output"
+```
+
+The above is an example. We would need to tune parameters depending on the model size, data size. The above example has been validated on 8 x A100 80GB.
```
nairbv, line 113: if the above is confirmed to work, maybe this should read "The user can tune the above hyperparameters based on their model size, data..." ? "We would need to" feels to me like it implies this example isn't validated.

README.md
```diff
@@ -0,0 +1,140 @@
...
+We can use [`lm-evaluation-harness`](https://github.com/EleutherAI/lm-evaluation-harness) from EleutherAI for evaluating the generated model. For example, for the Llama-13B model, using the above command and the model at the end of Epoch 5, we evaluated MMLU score to be `53.9` compared to base model to be `52.8`.
+
+How to run the validation:
+```
+pip install -U transformers
+pip install -U datasets
+git clone https://github.com/EleutherAI/lm-evaluation-harness
+cd lm-evaluation-harness
+pip install -e .
+python main.py
```
nairbv, line 128: I think you need `\`'s on these lines so user can copy/paste

requirements.txt
```diff
@@ -0,0 +1,14 @@
...
+aim==3.17.5
+sentencepiece
+tokenizers>=0.13.3
+tqdm
+trl
+ninja
+peft
+datasets>=2.15.0
+flash-attn
+fire
```
nairbv, line 14: should ibm-fms be a dependency?

tuning/config/peft.py
```diff
@@ -0,0 +1,21 @@
+from dataclasses import dataclass, field
+from typing import List
+
+
+@dataclass
+class lora_config:
+    r: int = 16
```
nairbv, line 7: why such a large default value for `r`?

llama-recipes uses 8. The original paper doesn't test 16 (when they explore "optimal rank r" they test 1,2,4,8, and 64) concluding "LoRA already performs competitively with a very small r"

examples/prompt_tuning_twitter_complaints/README.md
```diff
@@ -0,0 +1,65 @@
+## Prompt Tuning on Twitter Complaints
+
+This example follows [HF's Prompt Tuning example](https://huggingface.co/docs/peft/main/en/task_guides/clm-prompt-tuning)
+which demonstrates how to apply our tuning to any HF example.
+
+### Dataset
+
+The first thing is to make a `SFTTrainer`-competible dataset.
```
nairbv, line 8: typo: s/competible/compatible/

i might also suggest s/thing/step/

tuning/config/peft.py
lchu-ibm: @raghukiran1224 Any reason using 16?  I think I might copied this value from your hardcoded version.

tuning/sft_trainer.py
```diff
@@ -0,0 +1,149 @@
...
+            exit(-1)
+        
+        if data_args.dataset_text_field is None:
+            logger.error("Error, dataset_text_field is None, needs to be set for training")
+            exit(-1)
+        
+        #the below is specific to Llama since it uses sentencepiece tokenizer
+        # TODO: the below code is preparing data in a way that was specificed in  the example code from SFTTrainer for Llama.
+        # This is not accurate for GPTBigCode.
+        response_template_ids = tokenizer.encode(data_args.response_template, add_special_tokens=False)[2:]
```
nairbv, line 129: You can pass response_template directly, inside DataCollatorForCompletionOnlyLM it calls the tokenizer.encode if the response_template is of type str: https://github.com/huggingface/trl/blob/main/trl/trainer/utils.py#L97

or, we're dropping the first two tokens? assuming we need that, we should probably add a comment explaining it.

README.md
lchu-ibm: these are different commands, which we shouldn't use \ to make them one-line. it will error out I believe

tuning/sft_trainer.py
```diff
@@ -0,0 +1,149 @@
...
+        data_collator = DataCollatorForCompletionOnlyLM(response_template_ids, tokenizer=tokenizer, ignore_index=configs.IGNORE_INDEX)
+        trainer = SFTTrainer(
```
nairbv, line 131: IMO it might be simpler to `if packing: collator = None; else collator = ...` and then just call the SFTTrainer constructor once.

tuning/sft_trainer.py
```diff
@@ -0,0 +1,149 @@
...
+        model_max_length=training_args.model_max_length,
+        padding_side="right",
+        use_fast = True
+    )
+    if isinstance(tokenizer, LlamaTokenizer) or isinstance(tokenizer, LlamaTokenizerFast):
+        tokenizer.add_special_tokens({
+            "bos_token": "<s>",
+            "eos_token": "</s>",
+            "unk_token": "<unk>",
+            "pad_token": "<pad>",
```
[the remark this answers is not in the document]
lchu-ibm: the hardcoded in the training script is to make sure override user-provided/global-default values for llama, as this is preferred given the discussion we had yesterday.

But I am also included to change global-default as the default now is for llama.

README.md
nairbv: not the above, just this specific line (and the below lines) that are all parameters to `python main.py`

README.md
```diff
@@ -0,0 +1,140 @@
...
+The above is an example. We would need to tune parameters depending on the model size, data size. The above example has been validated on 8 x A100 80GB.
+
+For `GPTBigCode` models, Hugging Face has enabled Flash v2 and one can simply replace the `'LlamaDecoderLayer'` with `'GPTBigCodeBlock'` for proper sharding of the model.
+
+## Validation
+
+We can use [`lm-evaluation-harness`](https://github.com/EleutherAI/lm-evaluation-harness) from EleutherAI for evaluating the generated model. For example, for the Llama-13B model, using the above command and the model at the end of Epoch 5, we evaluated MMLU score to be `53.9` compared to base model to be `52.8`.
+
+How to run the validation:
+```
```
nairbv, line 122: might want to add `bash`

nairbv (approved):
there are a bunch of minor comments/suggestions we may want to consider but overall LGTM

tuning/data/tokenizer_data_utils.py
```diff
@@ -0,0 +1,31 @@
...
+import torch
+
+def tokenizer_and_embedding_resize(
+    special_tokens_dict: Dict,
+    tokenizer: transformers.PreTrainedTokenizer,
+    model: transformers.PreTrainedModel,
+):
+    """Resize tokenizer and embedding.
+    
+    TODO: In the future, make sure we can have vocab size divisible by 64.
```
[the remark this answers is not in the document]
lchu-ibm: @raghukiran1224 ↑

tuning/config/peft.py
raghukiran1224: No reason, we can change defaults once we know what works best.

lchu-ibm: will make it 8 then as that seems to be a more popular default value.

tuning/sft_trainer.py
raghukiran1224: This was based on the examples of correctness from the `sft-trainer` in: https://huggingface.co/docs/trl/main/en/sft_trainer#using-tokenids-directly-for-responsetemplate. This is probably different for the `GPTNeoX` tokenizer.

PR #2, opened 2023-12-22
allows disable flash attn and torch dtype param

Ssukriti:
The PR fixes the code to disable flash attention and allows passing torch dtype as an argument

### DIFF
--- a/tuning/sft_trainer.py
+++ b/tuning/sft_trainer.py
@@ -8,7 +8,7 @@ import datasets
 from tuning.data import tokenizer_data_utils
 from tuning.config import configs
 from tuning.utils.config_utils import get_peft_config, update_config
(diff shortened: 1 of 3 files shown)


tuning/sft_trainer.py
```diff
@@ -36,12 +37,12 @@ def main(**kwargs):
     # make sure to unset FSDP args when running on single gpu
     if not run_distributed:
         training_args.fsdp = ""
-        training_args.fsdp_config = {}
+        training_args.fsdp_config = {'xla':False}
```
Ssukriti, line 40: this key is accessed later in code, hence it had to be set to False to disable

Ssukriti:
I am merging it with a FYI as it is a minor refactor. I will put all major refactors for review

PR #10, opened 2024-01-05
fix : the way args are passed

Ssukriti:
Now that I am using HF parser the arguments are returned in their own namespace so we need to access them as class objects.

Without this fix, it was always setting tuning config to None and doing Fine tuning

### DIFF
--- a/tuning/sft_trainer.py
+++ b/tuning/sft_trainer.py
@@ -167,9 +167,9 @@ def main(**kwargs):
                                                             peft_config.PromptTuningConfig))
     parser.add_argument('--peft_method', type=str.lower, choices=['pt', 'lora', None, 'none'], default="pt")
     model_args, data_args, training_args, lora_config, prompt_tuning_config, peft_method, _ = parser.parse_args_into_dataclasses(return_remaining_strings=True)
-    if peft_method =="lora":
+    if peft_method.peft_method =="lora":
         tune_config=lora_config
-    elif peft_method =="pt":
+    elif peft_method.peft_method =="pt":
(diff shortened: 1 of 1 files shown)


lchu-ibm (approved):
LGTM, with the new refactor way, your namespace.field fix looks good.  An alternative way would be simply having peft_method under train_config, which is the old way we had that follows community style like what [llama-receipe](https://github.com/facebookresearch/llama-recipes/blob/main/src/llama_recipes/configs/training.py#L29) does, so we don't have to create one more namespace to hold a single value. 

I am good with either way.

Ssukriti:
Thank you for the review. 
I removed `peft_method` from train_args , with the move to modular functions, because otherwise for a production repo we will have to add parameter validation to the train() function that if train_args.peft==pt, we have to ensure PEFTConfig is passed , otherwise return an error. It is also redundant in train() function. Currently if train() gets PEftconfig it will do pt, if it gets LoraConfig it will apply that and do lora and `peft_method` is not needed in train()

we can keep peft_method as a separate command line argument. The code in the main() will be moved to an example , and users may or may not use command line arguments. They can alternatively just call train() directly
```
model args = configs.ModelArguments(arg1=value1...) # The values can come from any HTML parsing or Json files etc.
fms_fine_tuning.train(model_args, ...) 
```
the code in main just shows how to use command line arguments to create the necessary data classes to pass to train and makes experiments easier for datascientists

Ssukriti:
with more modular functions it is also easier to add unit tests than having everything in a giant main, so we have to ensure that parameters each function accepts is needed by that function

PR #13, opened 2024-01-12
fix import of aim_loader

anhuong:
When I install `tuning` and try to run `sft_trainer.train()` from within another python script, I get error
```
  File "/app/launch_training.py", line 28, in <module>
    from tuning import sft_trainer
  File "/usr/local/lib/python3.11/site-packages/tuning/sft_trainer.py", line 13, in <module>
    from aim_loader import get_aimstack_callback
ModuleNotFoundError: No module named 'aim_loader'
```

This should fix this error


Ssukriti (approved):
thank you

PR #15, opened 2024-01-18
add EOS token to dataset

Ssukriti:
This PR formats the dataset to add EOS token which is needed for CausalLMs

I have verified that format of dataset generated looks as expected

###Label : complaint<|endoftext|><pad>.....

### DIFF
--- a/tuning/sft_trainer.py
+++ b/tuning/sft_trainer.py
@@ -124,7 +124,7 @@ def train(
     
     # load the data by parsing JSON
     json_dataset = datasets.load_dataset('json', data_files=data_args.data_path)
-    formatted_dataset = json_dataset['train'].map(lambda example : {f"{data_args.dataset_text_field}" : example[f"data_args.dataset_text_field"] + tokenizer.eos_token})
+    formatted_dataset = json_dataset['train'].map(lambda example : {f"{data_args.dataset_text_field}" : example[f"{data_args.dataset_text_field}"] + tokenizer.eos_token})
     logger.info(f"Dataset length is {len(formatted_dataset)}")
 
     aim_callback = get_aimstack_callback()
--- a/tuning/sft_trainer.py
+++ b/tuning/sft_trainer.py
@@ -88,10 +88,12 @@ def train(
             "unk_token": "<unk>",
             "pad_token": "<pad>",
         })
+        response_template_ids = tokenizer.encode(data_args.response_template, add_special_tokens=False)[2:]
     elif isinstance(tokenizer, GPTNeoXTokenizerFast):
         tokenizer.add_special_tokens({
             "pad_token": "<pad>",
         })
+        response_template_ids = tokenizer.encode(data_args.response_template, add_special_tokens=False)
     else:
         logger.error("Unsupported model")
         exit(-1)
@@ -140,14 +142,6 @@ def train(
         
         if data_args.dataset_text_field is None:
             logger.error("Error, dataset_text_field is None, needs to be set for training")
-            exit(-1)
-        
-        if isinstance(tokenizer, LlamaTokenizer) or isinstance(tokenizer, LlamaTokenizerFast):
-            response_template_ids = tokenizer.encode(data_args.response_template, add_special_tokens=False)[2:]
-        elif isinstance(tokenizer, GPTNeoXTokenizerFast):
-            response_template_ids = tokenizer.encode(data_args.response_template, add_special_tokens=False)
-        else:
-            logger.error("Unsupported model")
             exit(-1)
         
         data_collator = DataCollatorForCompletionOnlyLM(response_template_ids, tokenizer=tokenizer, ignore_index=configs.IGNORE_INDEX)
--- a/tuning/sft_trainer.py
+++ b/tuning/sft_trainer.py
@@ -68,6 +68,9 @@ def main(**kwargs):
         tokenizer.add_special_tokens({
             "pad_token": "<pad>",
         })
+    else:
+        logger.error("Unsupported model")
+        exit(-1)
     
     model_max_length = tokenizer.model_max_length
     logger.info(f"Model max length {model_max_length}")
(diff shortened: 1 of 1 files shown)


Ssukriti:
Tests I have run with this PR:

1. granite 13 B , using formatted twitter dataset as specified here https://github.com/tharapalanivel/fms-hf-tuning/tree/main/examples/prompt_tuning_twitter_complaints
2. llama 7b on same
Technique: pt

script to test inference: 
```
from transformers import AutoModelForCausalLM, AutoTokenizer
import json
import peft

peft_model_id = "/data/sukriti/llama_eos/checkpoint-200"
model_id = "/data/llama/7B"

base_model = AutoModelForCausalLM.from_pretrained(model_id)

# Attempting to use the "easy ways" of loading this peft mdoel fail
# Both of these methods raise with
# >>> `create_and_replace` does not support prompt learning and adaption prompt yet.
# peft_model = AutoModelForCausalLM.from_pretrained(peft_model_id)
# base_model.load_adapter(peft_model_id)

# Here are our attempts instead to recreate the peft model from the saved artifacts...

# (We're not really sure if this bit is required, since we end up just loading the adapter separately later)
with open(peft_model_id + "/adapter_config.json") as f:
    adapter_config = json.load(f)
peft_config = peft.get_peft_config(adapter_config)
peft_model = peft.PeftModelForCausalLM(base_model, peft_config)

# Load the actual adapter as 'twitter' and set it as the active one for inference
peft_model.load_adapter(peft_model_id, "twitter")
peft_model.set_adapter("twitter")

# Uses the tokenizer from the base model, which is what TGIS would be doing
# (Switching to the one from `peft_model_id` doesn't seem to affect anything anyway)
tokenizer = AutoTokenizer.from_pretrained(model_id)

inputs = tokenizer("It was a fine day on twitter when ", return_tensors="pt").input_ids

peft_outputs = peft_model.generate(input_ids=inputs, max_new_tokens=50)
base_outputs = base_model.generate(input_ids=inputs, max_new_tokens=50)

peft_result = tokenizer.batch_decode(peft_outputs, skip_special_tokens=True)
base_result = tokenizer.batch_decode(base_outputs)

print(base_result)
print(peft_result)

inputs = tokenizer("### Text: This product was great. \n### Label:", return_tensors="pt").input_ids

peft_outputs = peft_model.generate(input_ids=inputs, max_new_tokens=50)
peft_result = tokenizer.batch_decode(peft_outputs)
print(peft_result)
```

inference on llama model looks like this :` ['<s> ### Text: This product was great. \n### Label: no complaint\n...\n</s>']`
inference on granite models look like this : ...### Label: no complaint no complaint complaint.... (doesnt stop)

I am unable to get the output to stop for granite 13b

Ssukriti:
we have verified that the masking is similar to caikit NLP codebase, where only pad tokens are getting masked and EOS token is not . Hence the EOS token is being used to calculate loss

example: labels: -100 -100.....-100 560 5238 0 

but we still see not very good outputs with granite models and it hallucinates till max length achieved

tuning/sft_trainer.py
```diff
@@ -92,7 +92,11 @@ def train(
         tokenizer.add_special_tokens({
             "pad_token": "<pad>",
         })
+    else:
```
gkumbhat, line 94: I think the special tokens are getting added 2 times here. One here and another time at line [118](https://github.com/foundation-model-stack/fms-hf-tuning/pull/15/files#diff-7d8f6f8626340337d03a2f04fec2840d2ede32b0c7af2640eed9759af45bb3cfR118)

tuning/sft_trainer.py
```diff
@@ -92,7 +91,11 @@ def train(
...
+        logger.error("Unsupported model")
+        exit(-1)
```
lchu-ibm, line 96: do we want to exit here?

I think the logics were:
for some models, we want to specially taken care; for the rest, don't do it.

so we might want to consider remove this else block.

Ssukriti: i wanted to ask @raghukiran1224 about this since I took this from a previous commit in this branch. Bloom Causal LMs have their own tokenizer and if we wnat to continue supporting all causal LMs , I will remove it.

If we only want to officially support granite and llama I will keep it . I think we can keep it open to all Causal LMs Raghu ? (I can remove the else)

Ssukriti: I removed the else

tuning/sft_trainer.py
Ssukriti: yes, there is a TODO for cleanup, confirmed not an issue

Ssukriti:
running tests with this PR now gave good results with granite model. The issue was in formatting of dataset where we need to use simple template without '###'

data formats will be discussed in consequent issues. This PR is adding EOS token, quality is verified on llama and granite models. 
This PR also adds TODO s for all the code sections we need to revisit and cleanup

tuning/sft_trainer.py
```diff
@@ -78,20 +78,27 @@ def train(
...
+    # TODO: understand if we need to hardcode these here or just use defaults in model
     if isinstance(tokenizer, LlamaTokenizer) or isinstance(tokenizer, LlamaTokenizerFast):
         tokenizer.add_special_tokens({
             "bos_token": "<s>",
             "eos_token": "</s>",
             "unk_token": "<unk>",
             "pad_token": "<pad>",
         })
-    elif isinstance(tokenizer, GPTNeoXTokenizerFast):
+    if isinstance(tokenizer, GPTNeoXTokenizerFast) or isinstance(tokenizer, GPT2Tokenizer):
```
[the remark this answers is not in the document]
Ssukriti: yup will change it. also just for clarification I added gPT2 here because granite 20B models are using GPT2. But I think in cleanup we should get rid of thsi block entirely and just do if pad token unset , set it to default as done below regardless of type of tokenizer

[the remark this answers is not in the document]
Ssukriti: done

Ssukriti:
Pr is ready for review @gkumbhat , thanks

gkumbhat (approved):

PR #17, opened 2024-01-19
fix: set model max length to either passed in or tokenizer value

anhuong:
Should use the user provided model max length if it valid and below the model's max length. Will also print log message if the user provided one exceeds the model one and both values which is helpful for debugging. Note that `train_args.model_max_length` is set by default to 4096 if the user does not provide a value.

### DIFF
--- a/tuning/sft_trainer.py
+++ b/tuning/sft_trainer.py
@@ -93,8 +93,10 @@ def train(
             "pad_token": "<pad>",
         })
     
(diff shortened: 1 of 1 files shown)


Ssukriti:
thank you Anh, this looks good. Due to lack of unit tests in repo at the moment, I trust that you have tested to ensure the change works?

Ssukriti (approved):

PR #20, opened 2024-01-22
feat: decouple callbacks from the train function

kmehant:
Decoupling callbacks preparation from the train function would help use train function without having callbacks being enforced by default (for instance aim callback) when using fms-hf-tuning repo as a module


dushyantbehl:
@kmehant this can also be closed now....

Issue #23, opened 2024-01-26
model checkpoints are broken when doing full parameter tuning 


lchu-ibm:
Currently when doing full parameter tuning (`peft_config=None`), the model can be fine-tuned and ckpt can be saved, however, the ckpt cannot be re-load back.

Root cause has been discovered with offline discussions, and most specifically:
1. `PeftSavingCallback` is not necessary for saving a full-parameter-tuning model. This callback is designed to separately save an adapter-only model, and the rationale behind it was that Trainer by itself will save everything in the root folder but for PEFT we want a clean adapter only folder, thus we duplicately save some of them in a separate place.
2. there is some [corner case bug](https://github.com/huggingface/transformers/issues/28630) in the model saving with safe tensor if saving with `save_pretrained` directly as it will drop some shared tensors (e.g. lm_head) during saving. The better way to do it would be `save_pretrained(..., state_dict=state_dict)` by explicitly passing the full state dict, same as [how Trainer natively save it](https://github.com/huggingface/transformers/blob/v4.37.1/src/transformers/trainer.py#L2937).

So... due to 2, the ckpt is missing some tensors thus cannot load back, and due to 1, this callback-generated broken ckpt is either being used or overwrites original ckpt.

Solution would be revise the callback to skip doing anything on full parameter tuning.

fabianlim:
@lchu-ibm `PeftSavingCallback` is not needed anymore after this fix https://github.com/huggingface/transformers/pull/28297

I had made a PR #53  to remove it.

PR #24, opened 2024-01-26
Fix PeftSavingCallback

lchu-ibm:
This PR introduces fix on PeftSavingCallback for both saving PEFT and non-PEFT (i.e. full parameter tuning).

1. Fix saving peft when original implementation would save peft in root ckpt folder which violates the purpose of this callback.
2. Fix saving full-parameter-tuning when it is actually not necessary for saving separately. And original implementation would introduce broken ckpt as discussed in https://github.com/foundation-model-stack/fms-hf-tuning/issues/23

### DIFF
--- a/tuning/sft_trainer.py
+++ b/tuning/sft_trainer.py
@@ -12,6 +12,8 @@ from tuning.utils.data_type_utils import get_torch_dtype
 
 from tuning.aim_loader import get_aimstack_callback
 from transformers.utils import logging
+from transformers.modeling_utils import unwrap_model, PreTrainedModel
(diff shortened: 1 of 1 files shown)


Ssukriti:
@fabianlim does this PR become obsolete with #53 ?

fabianlim:
@Ssukriti to my knowledge yes, as these checkpoint manipulation techniques are not required anymore, now that HF trainer will save the checkpoints correct as described in the writeup of #53

Ssukriti:
closing unless asked otherwise. PR #53 will be merged soon

PR #27, opened 2024-01-30
Local inference

alex-jw-brooks:
WIP PR for local inference, this description is still being updated.

Notable changes are:
- Adds a new class, `TunedCausalLM`, which are tuned causal LM models that are loadable, runnable, etc
- Implements .load / .run on that class ^ for local inference (wrapping around .generate API)
- Refactors sft_trainer's train capability into the `.train` `@classmethod` on the `TunedCausalLM class`; `sft_trainer.py` now contains only the CLI for tuning
- Adds a new script with a CLI for running inference, which can operate on either a single provided text or a file containing multiple texts; assumes data has been preformatted
- Adds a utility class, `AdapterConfigPatcher`, which can patch an override into Peft's adapter config (NOTE: This _does_ write to disk, and is _not_ safe for parallel loads with different overrides). To keep things simple, this is written generically, be we only allow patching of the base model for now
- Updates the README with examples of running inference 

### Other Notes
*AdapterConfigPatcher*
The reason it seems helpful to have this capability is that tuning may be carried out with a local base model. Since the local path is then used in the adapter config, the model will likely break when run on a different device. This is a bit of an issue, since Peft provides a nice automodel API for loading a tuned model, but doesn't seem to provide a way of specifying the path from the base model (i.e., pulls from the config).

*AdapterConfigPatcher* is a context manager, then when entered will patch an override into the `adapter_config.json` that Peft uses, and restores the original values when exited. Currently this is used to allow specification of a base model override when loading a model to run inference on.


tuning/tuned_model.py
```diff
@@ -0,0 +1,226 @@
...
+        By default, the paths for the base model and tokenizer are contained within the adapter
+        config of the tuned model. Note that in this context, a path may refer to a model to be
+        downloaded from HF hub, or a local path on disk, the latter of which we must be careful
+        with when using a model that was written on a different device.
+
+        Returns:
+            TunedCausalLM
+                An instance of this class on which we can run inference.
+        """
+        overrides = {"base_model_name_or_path": base_model_name_or_path} if base_model_name_or_path is not None else {}
```
alex-jw-brooks, line 59: Currently the adapter config patcher is an internal detail, and we only use it to override the base model, since it does not seem like we can do this from the automodel API, but it's a helpful capability to have

Ssukriti:
@alex-jw-brooks couple of thoughts on this PR:

1. I think `tuned_model` is a confusing name for the file and class to use past tense, because it is doing action of tuning or training. It makes it sound more like model has to be pretuned. Generally people use Trainer or Tuner for such kind of files ,example: https://github.com/hiyouga/LLaMA-Factory/blob/main/src/llmtuner/train/tuner.py

2. I think for the simplicity that we were aiming for and to avoid hopping too many files to find actual train code, It would be beneficial to include the entire class in the main entry point script - tuning/sft_trainer.py . I dont think we need to remove all the code from there. We can name the class `CausalLMTuner` or something like that 

3. In the future we can move out the __main__ from tuning/sft_trainer.py and create another run_trainer.py just like you created run_inference.py . Then main code to tune and infer models will be in sft_trainer.py and we will have run_trainer and run_inference to take command line arguments

Curious to hear what you think

tuning/tuned_model.py
```diff
@@ -0,0 +1,226 @@
+"""Interface for loading and running trained causal LMs. In the future,
+these capabilities will be unified with the sft_trainer's tuning capabilities.
```
Ssukriti, line 2: I think I didnt understand what this comment meant - in the future part

tuning/sft_trainer.py
```diff
@@ -175,7 +17,7 @@ def main(**kwargs):
         tune_config=prompt_tuning_config
     else:
         tune_config=None
-    train(model_args, data_args, training_args, tune_config)
+    TunedCausalLM.train(model_args, data_args, training_args, tune_config)
```
Ssukriti, line 20: this is the part in particular thats confusing calling it  TunedCausalLM.train

Ssukriti:
or another thought would be to create a separate folder for inference and we can move the load and run functions there with the class name as it is TunedCausalLM and example script ?  the train doesnt have to be a classmethod because its just saving model to disk and just can be a function.

Currently the local inference is just for our testing . As library evolves more we can think of the right design ?

tuning/sft_trainer.py
alex-jw-brooks: Hey - I agree with you 😄 most of the confusion from here comes from trying to avoid naming collisions with stuff in Peft, transformers etc.

After discussing in person, I have split .train back out of this class and refactored the inference capabilities into a single script outside of the tuning package

alex-jw-brooks:
Hey Sukriti - that sounds good, if we're not planning to support inference as a library (for now), it makes more sense to isolate the train / inference code in as portable a way as possible. 

For now, I've split train / inference back apart and refactored all of the inference stuff into a single file under `scripts`, with the same usage as the CLI described above. It doesn't depend on the tuning package either, so it should be easy for people to leverage for quick experiments.

As far as consolidating the utils etc into a single file - the main reason I chose to do that is so that we don't have a separate scripts package from the tuning package for just one CLI. Especially while we don't have Docker builds etc, having a module scripts package for one file would probably just lead to people having python path issues, and it's not that much code to begin with

scripts/run_inference.py
```diff
@@ -0,0 +1,234 @@
...
+            except OSError as e:
+                print("Failed to initialize checkpoint model!")
+                raise e
+        device = "cuda" if torch.cuda.is_available() else None
+        print(f"Inferred device: {device}")
+        peft_model.to(device)
+        return cls(peft_model, tokenizer, device)
+
+
+    def run(self, text: str, *, max_new_tokens: int) -> str:
```
Ssukriti, line 157: we arent doing anything with the extra arguments * . maybe it is a typo here?

Ssukriti (approved):
Have a minor comment, rest of code and README looks good. I would let @anhuong test it out as part of LoRA to ensure inference works and after the comment is addressed, we can merge

scripts/run_inference.py
alex-jw-brooks: It's kind of an obscure feature, but `*` in the func def forces everything after it to be passed as keyword arguments, and they can't be passed positionally.

```
>>> def foo(bar, *, baz):
...     pass
...
>>>
>>> foo(3, 100) # can't pass baz as a positional arg
  File "<stdin>", line 1, in <module>
TypeError: foo() takes 1 positional argument but 2 were given
>>> foo(bar=3, baz=100) # as long as we pass kwarg here, it's fine
```

Since there are lots of options that can be used when generating, I thought this would be best to avoid confusion in the future

Ssukriti: oh so it was on purpose :) thank you Alex !

anhuong:
Tested tuning llama2-7b model with  LoRA and then running inference script from this branch and it runs nicely
```
(venv) (tuning) bash-4.4$ python scripts/run_inference.py --model /dccstor/anhtest/model_output/tuning/llama-7b-twitter-lora-alex/checkpoint-60/ --max_new_tokens 50 --text "### Text: Never shopping at @MACcosmetics again. Every time I go in there, their employees are super rude/condescending. I'll take my 2955375 to @Sephora\n\n### Label:"
Loading checkpoint shards: 100%|█████████████████████████████████████████████████████████████████| 2/2 [00:19<00:00,  9.97s/it]
Inferred device: cuda
100%|████████████████████████████████████████████████████████████████████████████████████████████| 1/1 [00:01<00:00,  1.05s/it]
Exported results to: inference_result.json
(venv) (tuning) bash-4.4$ cat inference_result.json 
[
    {
        "input": "### Text: Never shopping at @MACcosmetics again. Every time I go in there, their employees are super rude/condescending. I'll take my 2955375 to @Sephora\\n\\n### Label:",
        "output": "<s> ### Text: Never shopping at @MACcosmetics again. Every time I go in there, their employees are super rude/condescending. I'll take my 2955375 to @Sephora\\n\\n### Label: complaint</s>"
    }

```

PR #28, opened 2024-02-01
[WIP] add early stopping callback

dushyantbehl:
Add early stopping callback to tuning script so tuning can be stopped if loss is not decreasing for subsequent epochs.


dushyantbehl:
Progress tracked in https://github.com/foundation-model-stack/fms-hf-tuning/pull/35

Closing this PR.

PR #30, opened 2024-02-01
Raise if fsdp plugin is unset

alex-jw-brooks:
When we try to run multigpu training, we explicitly set:
```
trainer.accelerator.state.fsdp_plugin.auto_wrap_policy = fsdp_auto_wrap_policy(model)
```
before launching the trainer. However, the Accelerator state doesn't always define the `fsdp_plugin` attribute, so this can crash in a cryptic way if we launch with Torchrun for multigpu with the wrong settings.

The `fsdp_plugin` of the Accelerator state for multigpu configurations is set [here](https://github.com/huggingface/accelerate/blob/main/src/accelerate/state.py#L816) in the Accelerate source. If you don't set `ACCELERATE_USE_FSDP`, it won't be defined.

This PR explicitly checks to see if the accelerator state has an `fsdp_plugin` before trying to update the auto wrap policy; if it doesn't, it raises an attribute error with a hint that you probably need to set `ACCELERATE_USE_FSDP` to run with a multigpu configuration.

### DIFF
--- a/tuning/sft_trainer.py
+++ b/tuning/sft_trainer.py
@@ -167,6 +167,11 @@ def train(
     )
 
     if run_distributed and peft_config is not None:
+        if not hasattr(trainer.accelerator.state, "fsdp_plugin"):
+            raise AttributeError(
+                "Accelerator state has no attr fsdp_plugin! Hint: did you set ACCELERATE_USE_FSDP?"
+            )
+        
         trainer.accelerator.state.fsdp_plugin.auto_wrap_policy = fsdp_auto_wrap_policy(model)
     trainer.train()
 


fabianlim:
@alex-jw-brooks @Ssukriti I dont think his change is needed if #53  is merged. This is because we now explicitly check `trainer.is_fsdp_enabled`, which will be True only if FSDP is enabled.

alex-jw-brooks:
Hey @fabianlim, that sounds good - I'm not super familiar with the Accelerate codebase around the plugin, but I assume that will handle the current attribute error and that the `fsdp_plugin` will always be defined.

I know that a lot of default args are set to `None` in `Accelerate`, including the `fsdp_plugin` in the `AcceleratorState`, so it might be a good idea to take a closer look at some point to make sure that we can't end up in a situation where `fsdp_plugin` is a defined state attribute that is set to `None`, since that would break similarly if it were possible. For now, let's close this PR  in favor of the other one though!

PR #31, opened 2024-02-02
Add formatting tools

### LINKED ISSUE #8: integrate black formatter
**Description**
To keep the code in good health and retain styling conventions, it is a good practice to integrate tools such as black formatter to the repo  https://black.readthedocs.io/en/stable/

**Acceptance criteria**
1. integrate black formatter
2. make sure it runs with every build as a check and build fails if formatter fails
3. Add it as a pre commit hook, so it checks on every commit https://stackoverflow.com/questions/71180810/python-black-code-formatter-is-there-any-way-to-apply-automatic-black-formatti

hickeyma:
Added the following formatters:
- [black](https://github.com/psf/black)
~- [prettier](https://github.com/prettier/prettier)~

Added the following import sorter:
- [isort](https://pycqa.github.io/isort/)

Added [tox](https://tox.wiki/en/4.12.1/) as means to kick off formatting which can also be used for setting up test environments.

Partial fix #8. Finish by integrating as part of build.

**Note to reviewers:**

- The formatting is not automated yet as need to add CI/CD
- Any changes recommended by the formatters are not committed until we automate as part of the build
- The formatters can be run manually:
   - Installing `tox` and dependencies: `pip install -r setup_requirements.txt`
   - Running it: `tox`


.pre-commit-config.yaml
```diff
@@ -0,0 +1,15 @@
+repos:
+    - repo: https://github.com/pre-commit/mirrors-prettier
```
Ssukriti, line 2: do we think and isort and black should be sufficient for python packages? 

I think we dont need prettier ?

hickeyma: Probably unnecessary with black formatter already. I'll remove for now. We can always add it back in if need be.

hickeyma:
Thanks for review @Ssukriti. Updated and ready for review again.

Ssukriti (approved):
Thank you Martin!!

PR #32, opened 2024-02-07
Merge models

alex-jw-brooks:
This PR adds a post-processing utility for merging one or more lora models back into a base model. If multiple lora models are provided, they are combined with equal weights (references supporting that this is what this PR is actually doing are provided in the docstring). An optional flag can be set to also export the tokenizer.

Example usage showing calling the merge util func and reloading + running an inference call with a tiny llama model - I've tried the same to with a llama 7b model to make sure it doesn't explode successfully:

```python
from tuning.utils.merge_model_utils import create_merged_model

save_tokenizer = True
base_model = "TinyLlama/TinyLlama-1.1B-step-50K-105b"
# checkpoints from tuning the tinyllama model
checkpoint_models = ["out/checkpoint-9", "out/checkpoint-13"]
merged_model_id = "my_merged_model"

# Creating the merged model...
create_merged_model(
    checkpoint_models=checkpoint_models,
    export_path=merged_model_id,
    base_model=base_model,
    save_tokenizer=save_tokenizer
)


# Reloading the tokenizer + merged model
import transformers
tokenizer = transformers.AutoTokenizer.from_pretrained(merged_model_id)
model = transformers.AutoModelForCausalLM.from_pretrained(merged_model_id).to("cuda")
## Make sure we can run some text through without exploding
tok_res = tokenizer("This is a text", return_tensors="pt")
gen_result = model.generate(tok_res.input_ids.to("cuda"))
decoded_toks = tokenizer.batch_decode(gen_result, skip_special_tokens=False)[0]
print(f"Model text: {decoded_toks}")
```

### DIFF
--- a/tuning/utils/merge_model_utils.py
+++ b/tuning/utils/merge_model_utils.py
@@ -2,9 +2,14 @@ import argparse
 import json
 import os
 from peft import PeftModel
-from transformers import AutoModelForCausalLM
+from transformers import AutoModelForCausalLM, AutoTokenizer
 
-def create_merged_model(checkpoint_model: str, export_path: str=None, base_model: str=None):
+def create_merged_model(
+    checkpoint_model: str,
+    export_path: str=None,
+    base_model: str=None,
+    save_tokenizer: bool=True
+):
     """Given a base model & a checkpoint model containing adapters, which were tuned with lora,
     load both into memory & create a merged model. If an export path is specified, write it
     to disk.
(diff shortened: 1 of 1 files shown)


anhuong (approved):
This looks great, I tested it with merging a llama-7b model with a single checkpoint and with multiple checkpoints and it succeeded. I also tested with granite-13b model which also ran successfully.

tuning/utils/merge_model_utils.py
```diff
@@ -0,0 +1,91 @@
...
+
+    Returns:
+        transformers model
+            Merged model created from the checkpoint / base model.
+    """
+    if isinstance(checkpoint_models, str):
+        checkpoint_models = [checkpoint_models]
+
+    if base_model is None:
+        base_model = fetch_base_model_from_checkpoint(checkpoint_models)
```
anhuong, line 52: I was just running this, and `fetch_base_model_from_checkpoint()` expects a string but `checkpoint_models` is being set to a list here which is causing an error. Sorry for missing this before

PR #38, opened 2024-02-09
Draft : HACK to preprocess data and pass to SFT TRainer

Ssukriti:


gkumbhat:
QQ for the testing parameters:
1. Not sure which model you tried with, but larger models with `float32` will probably get closer to OOM, specially train time.
1. Do we really need to use `1500` virtual token for twitter dataset? That seems a lot. Would have implications on quality

PR #39, opened 2024-02-09
feat: support str in target_modules for LoraConfig

### LINKED ISSUE #37: Support str in target_modules for LoraConfig
### Request
[LoraConfig](https://huggingface.co/docs/peft/en/package_reference/lora) can accept a List or a str for the target_modules as seen in the description below. This would be useful in order to support passing `"all-linear"` as an option instead of the specific attention layers.

[fms-hf-tuning](https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/tuning/config/peft_config.py#L8) `LoraConfig` currently accepts only a List: `target_modules: List[str] = field(default_factory=lambda: ["q_proj", "v_proj"])`. This causes it so that if one tries to pass `all-linear` it is interpreted as a List

Example
```
$ python tuning/sft_training.py --target_modules "all-linear"

# interpreted as
LoraConfig(r=8, lora_alpha=16, target_modules=['all-linear'], lora_dropout=0.05)

# subsequently gets used in SFTTrainer as
LoraConfig(peft_type=<PeftType.LORA: 'LORA'>, auto_mapping=None, base_model_name_or_path=None, revision=None, task_type='CAUSAL_LM', inference_mode=False, r=8, target_modules='all-linear', lora_alpha=16, lora_dropout=0.05, fan_in_fan_out=False, bias='none', use_rslora=False, modules_to_save=None, init_lora_weights=True, layers_to_transform=None, layers_pattern=None, rank_pattern={}, alpha_pattern={}, megatron_config=None, megatron_core='megatron.core', loftq_config={}) 

# errors with
ValueError: Target modules {'all-linear'} not found in the base model. Please check the target modules and try again.
```

I tried testing setting `target_modules: Union[List[str], str] = field(default_factory=lambda: ["q_proj", "v_proj"])` however  "all-linear" was still interpreted as a List instead of a string. This is likely due to the command-line parsing.

Note that `all-linear` is supported in PEFT 1.8.0 and must be upgraded.

### Acceptance criteria 

1. we should be able to support passing target_modules = 'all_linear' to fms-hf-tuning from command line as well for Lora Tuning
2. we have tested change works with at least one model llama-7b and doesnt crash

VassilisVassiliadis:
Resolves #37

- [x] Update definition of LoraConfig.target_modules to match that of the target_modules in [lora](https://github.com/huggingface/peft/blob/c1a83fd692cfbd5297d252e4eae4fdc4c59efa8c/src/peft/tuners/lora/config.py#L51-L58)
- [x] support --target_modules=all-linear
- [x] updated requirements.txt with depenncy to peft>0.8.0 (this version introduces support for --target_modules=all-linear)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,7 +8,7 @@ tokenizers>=0.13.3
 tqdm
 trl
 ninja
-peft
+peft>=0.8.0
 datasets>=2.15.0
 flash-attn
 fire
--- a/tuning/config/peft_config.py
+++ b/tuning/config/peft_config.py
@@ -1,11 +1,17 @@
 from dataclasses import dataclass, field
-from typing import List
+from typing import List, Union, Optional
 
 @dataclass
 class LoraConfig:
     r: int = 8
     lora_alpha: int = 32
-    target_modules: List[str] = field(default_factory=lambda: ["q_proj", "v_proj"])
+    target_modules: Optional[Union[List[str], str]] = field(default_factory=lambda: ["q_proj", "v_proj"], metadata={
+        "help": "The names of the modules to apply LORA to. When using a string, LORA uses regex matching. When using "
+            "a list of strings, LORA selects modules which either completely match or end with one of the strings. "
+            "If the value is the string 'all-linear', then LORA selects all linear and Conv1D modules except for "
+            "the output layer. If the value is none, then LORA chooses modules depending on the model architecture. "
+            "In that case, in that case LORA raises an exception prompting the user to manually select the modules."
+    })
     bias = "none"
     lora_dropout: float = 0.05
 
--- a/tuning/config/peft_config.py
+++ b/tuning/config/peft_config.py
@@ -1,11 +1,11 @@
 from dataclasses import dataclass, field
-from typing import List, Union, Optional
+from typing import List
 
 @dataclass
 class LoraConfig:
     r: int = 8
     lora_alpha: int = 32
-    target_modules: Optional[Union[List[str], str]] = field(default_factory=lambda: ["q_proj", "v_proj"], metadata={
+    target_modules: List[str] = field(default_factory=lambda: ["q_proj", "v_proj"], metadata={
         "help": "The names of the modules to apply LORA to. When using a string, LORA uses regex matching. When using "
             "a list of strings, LORA selects modules which either completely match or end with one of the strings. "
             "If the value is the string 'all-linear', then LORA selects all linear and Conv1D modules except for "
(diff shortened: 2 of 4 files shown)


VassilisVassiliadis:
Due to the way that HFArgumentParser constructs an ArgumentParser (https://github.com/huggingface/transformers/blob/2749e479f30ab13235b0b9b4a6bbcf4c3b29a081/src/transformers/hf_argparser.py#L206-L208) the type of `target_modules` *must* be a `List[str]`. Otherwise `--target_modules foo bar` ends up being `["foo"]` instead of `["foo", "bar"]`.

Tried this PR by running:

```
model="hf-tiny-model-private/tiny-random-BloomForCausalLM"
python tuning/sft_trainer.py --output_dir tuned --model_name_or_path ${model} --use_flash_attn=false\
        --num_train_epochs 1 --evaluation_strategy no --response_template "\n### Response" \
        --dataset_text_field output --tokenizer_name_or_path ${model} \
        --torch_dtype float32 --data_path common_en_news_combined_512-preprocessed.jsonl \
        --peft_method lora  --target_modules=all-linear 
```

Then I purposely introduced a mistake by leaving out `--target_modules` so that the code picks up the default value of that parameter:

```commandline
model="hf-tiny-model-private/tiny-random-BloomForCausalLM"
python tuning/sft_trainer.py --output_dir tuned --model_name_or_path ${model} --use_flash_attn=false\
        --num_train_epochs 1 --evaluation_strategy no --response_template "\n### Response" \
        --dataset_text_field output --tokenizer_name_or_path ${model} \
        --torch_dtype float32 --data_path common_en_news_combined_512-preprocessed.jsonl \
        --peft_method lora
```

I got the following (expected) exception:

```
ValueError: Target modules {'v_proj', 'q_proj'} not found in the base model. Please check the target modules and try again.
```

Same thing happened when I manually entered these 2 layer names:

```commmandline
model="hf-tiny-model-private/tiny-random-BloomForCausalLM"
python tuning/sft_trainer.py --output_dir tuned --model_name_or_path ${model} --use_flash_attn=false\
        --num_train_epochs 1 --evaluation_strategy no --response_template "\n### Response" \
        --dataset_text_field output --tokenizer_name_or_path ${model} \
        --torch_dtype float32 --data_path common_en_news_combined_512-preprocessed.jsonl \
        --peft_method lora --target_modules do not exist
```

I got the following (expected) exception:

```
ValueError: Target modules {'not', 'exist', 'do'} not found in the base model. Please check the target modules and try again.
```

Ssukriti:
@VassilisVassiliadis this will get confusing if we state from command line the argument has to be list, but from function call it can be string. I ahve this alternate proposal :

1. lets stick to List[str] only in our Loraconfig datastructure definition . we can say it can be list of modules or ["all-linear"]
2. while passing it to HuggingFace, the huggingface config gets created here https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/tuning/utils/config_utils.py#L53 
so there you can check if tuning_config.target_modules==["all-linear"], then tuning_config.target_modules = "all-linear"

VassilisVassiliadis:
Done!

PR #44, opened 2024-02-14
Data formats

alex-jw-brooks:


alex-jw-brooks:
Closing - we decided to go with an alternate approach here, which is implemented by https://github.com/foundation-model-stack/fms-hf-tuning/pull/166

PR #45, opened 2024-02-15
WIP: Policy driven training control

seshapad:
Signed-off-by: Padmanabha Venkatagiri Seshadri [seshapad@in.ibm.com](mailto:seshapad@in.ibm.com)
Co-authored-by: Dushyant Behl [dushyantbehl@hotmail.com](mailto:dushyantbehl@hotmail.com)

Addresses the issue [537](https://github.ibm.com/ai-foundation/watson-fm-stack-tracker/issues/537).

# User Design details:
In this PR, we have implemented a trainer callback which accepts a `training control definition` file (in YAML format) which facilitates the definition of:
1.  Rules to control training loop
2. Trigger points that evaluate the above rules
3. Control action that needs to be performed if rule is evaluated to true.

The training definition file is as illustrated below:
```
controllers:
  - name: loss-controller
    controller-metrics:
      - name: wsl
        handler-class: WindowStepLoss
        arguments:
          window-size: 3
    triggers:
      - on_step_end
    control-operation:
      should_training_stop: True
    rules:
      - wsl_consistently_increasing == 1 and wsl_average_loss > 2.5
 ```
 Following is a description of the fields:
 - `controllers`: List of controllers. There can be more than one, each defining its own metrics and rules.
 - `controllers[i].name`: Name of the controller
 - `controllers[i].controller-metrics`: List of controller metric classes. These classes are defined in the `tuning/policydrivencontroller/controllermetrics/metrics.py` file. Each class inherits the following structure:
     ```
     class MetricHandler:
        def validate(self, training_args):
            pass
        
        def compute(self, training_state, training_args=None, metrics=None):
            pass 
     ```
     The `validate()` method validates the trainining arguments to check if the given metric could be evaluated with the settings of the training arguments. If not, an assert is thrown.
     The `compute()` method computes the metric and exposes certain variables and their corresponding values. The names of these variables could be used in the rules to evaluate conditions at which control actions need to be performed with respect to the training loop.
     These classes can be user-defined. To add a new metric class, simply implement the above structure and add the class to the above specified file.
     The `name` (user-defined), `handle-class` (name of the class) and corresponding arguments can be specified in this section.
- `triggers`: Events (e.g. `on_step_end`, `on_epoch_end` etc) at which the metrics and rules need to evaluated. These are derived from the HF (TrainerCallback)[https://huggingface.co/docs/transformers/v4.37.2/en/main_classes/callback#transformers.TrainerCallback].
- `control-operation`: Control operations that need to be performed when a rule is evaluated to True. These control  operations are derived from HF (TrainerControl)[https://huggingface.co/docs/transformers/v4.37.2/en/main_classes/callback#transformers.TrainerControl] class.
- `rules`: Rules that need to be evaluated. These rules use the variables exposed by the metric class. For example, `wsl_consistently_increasing` and `wsl_average_loss` are variables which are exposed by WindowStepLoss class. The values of these variables change as the training progresses, and correspondingly rules are evaluated on the state of the variables every time a trigger event happens (in the above example, whenever `on_step_end` gets triggered.

### CODE REVIEW

File: tuning/policydrivencontroller/controllermetrics/metrics.py [hunk 1]
```diff
@@ -0,0 +1,244 @@
...
+
+class MetricHandler:
+    """Base class for the controller-metrics"""
+    def validate(self, training_args) -> bool:
+        pass
+    
+    def compute(self, training_state, training_args=None, metrics=None) -> dict:
+        pass   
+
+class WindowStepLoss(MetricHandler):
```
ashokponkumar, line 18: Why not have the same function for both step and epoch, and take interval strategy as a parameter? We can call the class as something like `LossInWindow`

tuning/policydrivencontroller/controllermetrics/metrics.py
```diff
@@ -0,0 +1,244 @@
...
+        avg_loss = np.mean(w)
+        std_loss = np.std(w, dtype=np.float64)
+        first_and_last_loss = window[0] < window[len(window)-1]
+        exposed_data = {self.__name + '_' + 'consistently_increasing': int(consistently_increasing), \
+                self.__name + '_' + 'average_loss': avg_loss,\
+                self.__name + '_' + 'std_loss': std_loss,\
+                self.__name + '_' + 'first_and_last_loss': int(first_and_last_loss)}
+        return exposed_data
+
+class EpochLoss(MetricHandler):
```
ashokponkumar, line 74: Do we need a separate class for epoch?

tuning/policydrivencontroller/controllermetrics/metrics.py
```diff
@@ -0,0 +1,244 @@
...
+    def validate(self, training_args):
+        """Validate the training arguments (e.g logging_steps) are compatible with the computation of this metric
+
+        Args:
+            training_args: Training arguments
+
+        Returns:
+            bool
+        """
+        logger.info("VALIDATE ==> %s %s %s %s" % (str(training_args.load_best_model_at_end), \
```
ashokponkumar, line 191: Do we need info level logs here?

tuning/policydrivencontroller/pdt_callback.py
```diff
@@ -0,0 +1,236 @@
...
+from typing import Optional, Union
+import json
+import yaml
+import os
+import copy
+from .controllermetrics import metrics as contmetrics
+
+logger = logging.get_logger(__name__)
+
+class PolicyDrivenTrainerControl(TrainerCallback):
```
ashokponkumar, line 14: Better to explicitly say it is a callback
```suggestion
class TrainingControlCallBack(TrainerCallback):
```

tuning/policydrivencontroller/pdt_callback.py
```diff
@@ -0,0 +1,236 @@
...
+
+        Args:
+            train_control_args: File path for trainer control definition file
+            training_args: TrainingArguments object
+        """
+        self.__controllers = {}
+        if os.path.exists(train_control_args.training_control_definition_file):
+            with open(train_control_args.training_control_definition_file, "r") as f:
+                self.training_control_def = yaml.safe_load(f)
+                for controller in self.training_control_def['controllers']:
```
ashokponkumar, line 28: If the section is missing, it will result in null pointer errors, isn't it?

tuning/policydrivencontroller/controllermetrics/metrics.py
```diff
@@ -0,0 +1,244 @@
...
+                continue
+        if len(loss_array) > 0 and added == False:
+            l = np.array(loss_array)
+            epoch_avg_loss = np.mean(l)
+            epoch_std_loss = np.std(l, dtype=np.float64)
+            self.__cache.append({'epoch': epoch, 'avg_loss': epoch_avg_loss, 'std_loss': epoch_std_loss, 'end_loss': latest_log_loss})
+        return self.__externalize_data()
+        
+            
+class EvalMetricBasedControl(MetricHandler):
```
ashokponkumar, line 168: Do we need a separate controller for this? Why not loss be a parameter in the above class?

tuning/policydrivencontroller/pdt_callback.py
```diff
@@ -0,0 +1,236 @@
...
+        if 'should_training_stop' in cb['control-operation']:
+            control.should_training_stop = cb['control-operation']['should_training_stop']
+        elif 'should_epoch_stop' in cb['control-operation']:
+            control.should_epoch_stop = cb['control-operation']['should_epoch_stop']
+        elif 'should_save' in cb['control-operation']:
+            control.should_save = cb['control-operation']['should_save']
+        elif 'should_evaluate' in cb['control-operation']:
+            control.should_evaluate = cb['control-operation']['should_evaluate']
+        elif 'should_log' in cb['control-operation']:
+            control.should_log = cb['control-operation']['should_log']
```
ashokponkumar, line 68: can we use a for loop through the operations listed in `control-operation`, so that if it changes, it gets automatically supported?

ashokponkumar (changes requested):
Nice work. Few suggestions @seshapad !

tuning/config/configs.py
```diff
@@ -57,3 +57,14 @@ class TrainingArguments(transformers.TrainingArguments):
         default=False,
         metadata={"help": "Packing to be enabled in SFT Trainer, default is False"},
     )
+
+@dataclass
+class PTCArguments():
```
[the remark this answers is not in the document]
seshapad: Done.

tuning/config/configs.py
```diff
@@ -57,3 +57,14 @@ class TrainingArguments(transformers.TrainingArguments):
...
+    training_control_definition_file: str = field(
+        default="ctldef.yaml",
```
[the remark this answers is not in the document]
seshapad: Done.

tuning/config/configs.py
```diff
@@ -57,3 +57,14 @@ class TrainingArguments(transformers.TrainingArguments):
         default=False,
         metadata={"help": "Packing to be enabled in SFT Trainer, default is False"},
     )
+
+@dataclass
+class PTCArguments():
+    training_control_definition_file: str = field(
```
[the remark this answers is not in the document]
seshapad: Done.

tuning/config/configs.py
```diff
@@ -57,3 +57,14 @@ class TrainingArguments(transformers.TrainingArguments):
...
+        default="ctldef.yaml",
+        metadata={
+            "help": (
+                "Training control definition file in YAML format."
```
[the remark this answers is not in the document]
seshapad: Done.

tuning/policydrivencontroller/__init__.py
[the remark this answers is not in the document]
seshapad: Done.

tuning/policydrivencontroller/pdt_callback.py
```diff
@@ -0,0 +1,236 @@
...
+        """Given a controller-block, applies the control operation to the training loop.
+
+        Args:
+            cb: Controller block dictionary
+            control: TrainerControl object
+
+        Returns:
+            None.
+        """
+        if 'should_training_stop' in cb['control-operation']:
```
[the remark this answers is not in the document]
seshapad: Done.

tuning/policydrivencontroller/pdt_callback.py
```diff
@@ -0,0 +1,236 @@
...
+                if cm_res == None:
+                    continue
+                metric_result.update(cm_res)
+            for rule in controller['rules']:
+                try:
+                    mr_copy = copy.deepcopy(metric_result)
+                    rule_outcome = eval(rule, metric_result)
+                    if rule_outcome:
+                        logger.warn('[%s] metrics so far: %s' % (name, str(mr_copy)))
+                        logger.warn('[%s] rule[%s] TRIGGERED' % (name, str(rule)))
```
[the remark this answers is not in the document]
seshapad: Its to warn the user that a rule has been triggered.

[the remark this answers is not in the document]
seshapad: Done.

examples/training-control-configs/Readme.md
```diff
@@ -0,0 +1,11 @@
...
+To use one of these files with the trainer, execute the `sft_trainer.py` with the following option: 
+```
+--training_control_config_file "examples/training-control-configs/<file-name>"
+```
+
+# Note on training control configuration examples
+- `ctldef_step_v0.3.yaml`: Defines a training controller which computes loss at every step and loss consistently increases for three steps, then the training is stopped.
+- `ctldef_epoch_v0.3.yaml`: Defines a epoch level training controller which computes loss at every epoch. The rule applied here is to compare a current epoch loss with previous epoch loss and it turns out to be more, then training is stopped.
+- `ctldef_epoch_threshold_v0.3.yaml`: Defines a training controller similar to previous case, but also adds a threshold constraint.
+- `ctldef_evaluate_v0.3.yaml`: Defines a training controller which behaves similar to the `EarlyStoppingCallback` from hugging face which can be found [here](https://github.com/huggingface/transformers/blob/v4.37.2/src/transformers/trainer_callback.py#L543).
```
ashokponkumar, line 11: Do we want to change the names of these files?

examples/training-control-configs/ctldef_epoch_threshold_v0.3.yaml
```diff
@@ -0,0 +1,13 @@
...
+      - name: elc
+        handler-class: EpochLoss
+        arguments:
+          window_size: 2
+    triggers:
+      - on_epoch_end
+    control-operations:
+      should_training_stop: True
+    rules:
+      - (elc["final_loss"][-1] > elc["final_loss"][-2]) or (elc["final_loss"][-2] - elc["final_loss"][-1] < 0.5)
```
ashokponkumar, line 13: Didn't we want to remove the array?

examples/training-control-configs/ctldef_evaluate_v0.3.yaml
```diff
@@ -0,0 +1,11 @@
+controllers:
+  - name: eval-based-stopback-controller
+    controller-metrics:
+      - name: embc
+        handler-class: EvalMetricBasedControl
+    triggers:
+      - on_evaluate
+    control-operations:
+      should_training_stop: True
+    rules:
```
ashokponkumar, line 10: Didn't we want to remove the array?

examples/training-control-configs/ctldef_step_v0.3.yaml
```diff
@@ -0,0 +1,13 @@
...
+      - name: wsl
+        handler-class: WindowStepLoss
+        arguments:
+          window_size: 3
+    triggers:
+      - on_step_end
+    control-operations:
+      should_training_stop: True
+    rules:
+      - wsl["consistently_increasing"] == 1
```
ashokponkumar, line 13: Do we still have rules?

tuning/sft_trainer.py
```diff
@@ -216,6 +218,12 @@ def train(
     file_logger_callback = FileLoggingCallback(logger)
     peft_saving_callback = PeftSavingCallback()
     callbacks = [aim_callback, peft_saving_callback, file_logger_callback]
+    try:
+        tc_callback = TrainerController(train_control_args, train_args)
+        callbacks.append(tc_callback)
+    except Exception as e:
+        logger.error(f'TrainerController callback was not enabled due to this exception: {e}')
+        exit(0)
```
ashokponkumar, line 226: Why not just use logger.fatal?

tuning/trainingcontroller/controllermetrics/metrics.py
```diff
@@ -0,0 +1,218 @@
...
+from transformers import TrainerState, IntervalStrategy
+from transformers.utils import logging
+from collections import deque
+import math
+import copy
+
+logger = logging.get_logger(__name__)
+
+class MetricHandler:
+    """Base class for the controller-metrics"""
```
ashokponkumar, line 11: Don't we want a init to store the name of the handler?

ashokponkumar (changes requested):
Few more comments..

tuning/sft_trainer.py
seshapad: Removed the exit() call and changed the logger to warn. We dont need to exit the trainer in this case. If the callback doesnt initialize successfully, then training could still continue without issues.

examples/training-control-configs/ctldef_step_v0.3.yaml
seshapad: Removed.

examples/training-control-configs/ctldef_evaluate_v0.3.yaml
seshapad: Removed.

examples/training-control-configs/ctldef_epoch_threshold_v0.3.yaml
seshapad: Removed.

tuning/trainingcontroller/controllermetrics/metrics.py
seshapad: This is not required anymore.

examples/trainer-controller-configs/Readme.md
```diff
@@ -0,0 +1,11 @@
...
+To use one of these files with the trainer, execute the `sft_trainer.py` with the following option: 
+```
+--training_control_config_file "examples/training-control-configs/<file-name>"
+```
+
+# Note on training control configuration examples
+- `trainercontroller_config_step_v0.3.yaml`: Defines a training controller which computes loss at every step and loss consistently increases for three steps, then the training is stopped.
+- `trainercontroller_config_epoch_v0.3.yaml`: Defines a epoch level training controller which computes loss at every epoch. The rule applied here is to compare a current epoch loss with previous epoch loss and it turns out to be more, then training is stopped.
+- `trainercontroller_config_epoch_threshold_v0.3.yaml`: Defines a training controller similar to previous case, but also adds a threshold constraint.
+- `trainercontroller_config_evaluate_v0.3.yaml`: Defines a training controller which behaves similar to the `EarlyStoppingCallback` from hugging face which can be found [here](https://github.com/huggingface/transformers/blob/v4.37.2/src/transformers/trainer_callback.py#L543).
```
ashokponkumar, line 11: The file names don't seem to have 0.3, isn't it?

examples/trainer-controller-configs/Readme.md
```diff
@@ -0,0 +1,11 @@
+# How-To
+To use one of these files with the trainer, execute the `sft_trainer.py` with the following option: 
+```
+--training_control_config_file "examples/training-control-configs/<file-name>"
```
ashokponkumar, line 4:
```suggestion
--trainer_controller_config_file "examples/trainer-controller-configs/<file-name>"
```

examples/trainer-controller-configs/Readme.md
```diff
@@ -0,0 +1,11 @@
...
+```
+
+# Note on training control configuration examples
```
ashokponkumar, line 7:
```suggestion
# Note on trainer controller configuration examples
```

examples/trainer-controller-configs/Readme.md
```diff
@@ -0,0 +1,11 @@
...
+- `trainercontroller_config_step_v0.3.yaml`: Defines a training controller which computes loss at every step and loss consistently increases for three steps, then the training is stopped.
```
ashokponkumar, line 8:
```suggestion
- `trainercontroller_config_step_v0.3.yaml`: Defines a trainer controller which computes loss at every step and if the loss consistently increases for three steps, then the training is stopped.
```

examples/trainer-controller-configs/Readme.md
```diff
@@ -0,0 +1,11 @@
...
+- `trainercontroller_config_epoch_v0.3.yaml`: Defines a epoch level training controller which computes loss at every epoch. The rule applied here is to compare a current epoch loss with previous epoch loss and it turns out to be more, then training is stopped.
+- `trainercontroller_config_epoch_threshold_v0.3.yaml`: Defines a training controller similar to previous case, but also adds a threshold constraint.
```
ashokponkumar, line 10:
```suggestion
- `trainercontroller_config_epoch_threshold_v0.3.yaml`: Defines a trainer controller similar to previous case, but also adds a threshold constraint.
```

examples/trainer-controller-configs/Readme.md
ashokponkumar, line 11:
```suggestion
- `trainercontroller_config_evaluate_v0.3.yaml`: Defines a trainer controller which behaves similar to the `EarlyStoppingCallback` from hugging face which can be found [here](https://github.com/huggingface/transformers/blob/v4.37.2/src/transformers/trainer_callback.py#L543).
```

tuning/trainercontroller/controllermetrics/metrics.py
```diff
@@ -0,0 +1,218 @@
...
+
+class MetricHandler:
+    """Base class for the controller-metrics"""
+    def validate(self, training_args) -> bool:
+        pass
+    
+    def compute(self, training_state, training_args=None, metrics=None) -> dict:
+        pass   
+
+class WindowStepLoss(MetricHandler):
```
ashokponkumar, line 18: Do we need Window in the name? We don't have it in Epoch loss.

tuning/trainercontroller/controllermetrics/metrics.py
```diff
@@ -0,0 +1,218 @@
...
+from transformers import TrainerState, IntervalStrategy
+from transformers.utils import logging
+from collections import deque
+import math
+import copy
+
+logger = logging.get_logger(__name__)
+
+class MetricHandler:
+    """Base class for the controller-metrics"""
```
ashokponkumar, line 11: Shouldn't we process the name and default arguments in the base class constructor?

tuning/trainercontroller/controllermetrics/metrics.py
```diff
@@ -0,0 +1,218 @@
...
+    def validate(self, training_args) -> bool:
+        pass
+    
+    def compute(self, training_state, training_args=None, metrics=None) -> dict:
+        pass
```
ashokponkumar, line 16: Should we be passing training args to each and every function? shouldn't it be processed and kept in init?

tuning/trainercontroller/controllermetrics/metrics.py
```diff
@@ -0,0 +1,218 @@
...
+            if window[i] > window[i+1]:
+                consistently_increasing = False
+                break
+        w = np.array(window)
+        avg_loss = np.mean(w)
+        std_loss = np.std(w, dtype=np.float64)
+        exposed_data = {'consistently_increasing': int(consistently_increasing), \
+                    'average_loss': avg_loss,\
+                    'std_loss': std_loss,\
+                    'window': window}
```
ashokponkumar, line 65: Should we call it something like, loss_in_window?

tuning/trainercontroller/controllermetrics/metrics.py
```diff
@@ -0,0 +1,218 @@
...
+        pass
+    
+    def compute(self, training_state, training_args=None, metrics=None) -> dict:
+        pass   
+
+class WindowStepLoss(MetricHandler):
+    """Implements the controller metric which evaluates loss-per-step over a user-defined window"""
+    
+    def __init__(self, window_size):
+        # Initialize the handler arguments
```
ashokponkumar, line 22: I thought we discussed to have a cache in here too similar to epoch loss class, isn't it?

tuning/trainercontroller/controllermetrics/metrics.py
```diff
@@ -0,0 +1,218 @@
...
+            training_state: TrainerState object
+            training_args: TrainingArguments object
+            metrics: [optional] metrics data
+
+        Returns:
+            dict
+        """
+        previous_epoch = -1
+        loss_array = None
+        logs_latest_first = list(reversed(training_state.log_history))
```
ashokponkumar, line 119: Instead of processing the entire history, reversing and more, can we instead just store the index to which we have previously processed as a object variable and process only the remaining?

tuning/trainercontroller/controllermetrics/metrics.py
```diff
@@ -0,0 +1,218 @@
...
+                    final_loss = loss
+                epoch = math.ceil(log['epoch'])
+                if previous_epoch == -1:
+                    previous_epoch = epoch
+                    loss_array = []
+                if epoch != previous_epoch and loss_array != None:
+                    if len(loss_array) > 0:
+                        l = np.array(loss_array)
+                        epoch_avg_loss = np.mean(l)
+                        epoch_std_loss = np.std(l, dtype=np.float64)
```
ashokponkumar, line 136: Do we really need numpy for just mean and std? why not just use python statistic package?

tuning/trainercontroller/controllermetrics/metrics.py
```diff
@@ -0,0 +1,218 @@
...
+                loss_array.append(loss)
+            except:
+                # Ignoring log lines not containing relevant fields
+                continue
+        if len(loss_array) > 0 and added == False:
+            l = np.array(loss_array)
+            epoch_avg_loss = np.mean(l)
+            epoch_std_loss = np.std(l, dtype=np.float64)
+            self.__cache.append({'epoch': epoch, 'avg_loss': epoch_avg_loss, 'std_loss': epoch_std_loss, 'final_loss': final_loss})
+        return self.__externalize_data()
```
ashokponkumar, line 149: If the cache size is greater than window size, we return longer than the window size?

ashokponkumar (changes requested):

tuning/trainercontroller/controllermetrics/metrics.py
seshapad: numpy is considered more efficient for these operations isnt it?

examples/trainer-controller-configs/Readme.md
seshapad: done

tuning/policydrivencontroller/pdt_callback.py
seshapad: done

tuning/policydrivencontroller/controllermetrics/metrics.py
seshapad: Addressed as discussed.

tuning/policydrivencontroller/pdt_callback.py
seshapad: Addressed

examples/trainer-controller-configs/Readme.md
seshapad: done

examples/trainer-controller-configs/Readme.md
seshapad: done

tuning/trainercontroller/controllermetrics/metrics.py
seshapad: done

tuning/trainercontroller/controllermetrics/metrics.py
seshapad: there are no default arguments now.

tuning/trainercontroller/controllermetrics/metrics.py
seshapad: training args are not used anywhere except the validate().

tuning/trainingcontroller/controllermetrics/metrics.py
seshapad: Added back the name and made related changes to base classes as discussed.

tuning/trainercontroller/controllermetrics/metrics.py
seshapad: added name.

tuning/trainercontroller/controllermetrics/metrics.py
seshapad: window seems better... `loss_in_window` can be inferred by looking at class name and argument name.

tuning/trainercontroller/controllermetrics/metrics.py
seshapad: addressed

tuning/trainercontroller/controllermetrics/metrics.py
seshapad: addressed.

tuning/trainercontroller/controllermetrics/metrics.py
seshapad: addressed.

tuning/trainercontroller/controllermetrics/metrics.py
seshapad: Only the EarlyStopping metric uses numpy as it comes from the original implementation.

tuning/sft_trainer.py
```diff
@@ -223,6 +225,11 @@ def train(
     file_logger_callback = FileLoggingCallback(logger)
     peft_saving_callback = PeftSavingCallback()
     callbacks = [aim_callback, peft_saving_callback, file_logger_callback]
+    try:
+        tc_callback = TrainerControllerCallback(trainer_controller_args, train_args)
+        callbacks.append(tc_callback)
+    except Exception as e:
+        logger.warn(f'TrainerController callback was not enabled due to this exception: {repr(e)}')
```
ashokponkumar, line 232: Make it logger.error

tuning/trainercontroller/callback.py
```diff
@@ -0,0 +1,250 @@
...
+
+        Args:
+            cb: Controller block dictionary
+            control: TrainerControl object
+
+        Returns:
+            None.
+        """
+        for k, v in cb['control-operations'].items():
+            setattr(control, k, v)
```
ashokponkumar, line 85: We don't need a function for it I guess.

tuning/trainercontroller/callback.py
```diff
@@ -0,0 +1,250 @@
...
+            for i in range(len(controller['controller-metrics'])):
+                cm_obj = controller_metrics_objs[i]
+                cm_res = cm_obj.compute(state, args, metrics)
+                if cm_res == None:
+                    continue
+                metric_result.update(cm_res)
+            rule = controller['rule']
+            try:
+                if eval(rule, metric_result):
+                    logger.warn('<%s> rule[%s] triggered' % (name, str(rule)))
```
ashokponkumar, line 119: Lets make it an info message.

ashokponkumar (approved):
Very nice! just 2 minor nits.

tuning/trainercontroller/callback.py
seshapad: done.

tuning/trainercontroller/callback.py
seshapad: done.

alex-jw-brooks (changes requested):
Hi @seshapad, this is super cool, thanks for putting so much thought into it! I've added some comments, but as far as some general thoughts go:

- It would be best to move away from generic exception catching leading to `continue/pass` silently - this happens in a few places, but in such situations it would probably be better to have throw a well-defined error as soon as possible, otherwise I think there’s risk of early stopping etc not working properly, and not logging due to continue/pass in generic exceptions. This is probably the biggest change needed IMO, the framework itself looks great

- It would be great to get some unit tests here, e.g., by converting some of the examples into end to end tests on a tiny model and validate that the behavior is actually being triggered correctly

- I’ll leave this up to you, but splitting the PR up into a more atomic smaller PRs might make it both easier to test and review quickly. For example, one PR could contain the base framework / validation / callback etc, while other PR(s) could contain the early stopping implementation, the cache metric handler, etc. But this is up to you, if you'd prefer to keep it all together, that is fine also

Sorry for the delayed review. Please reach out if you need anything or have any questions! Great work!

examples/trainer-controller-configs/Readme.md
```diff
@@ -0,0 +1,11 @@
+# How-To
+To use one of these files with the trainer, execute the `sft_trainer.py` with the following option: 
+```
+--trainer_controller_config_file "examples/trainer-controller-configs/<file-name>"
+```
+
+# Note on trainer controller configuration examples
+- `trainercontroller_config_step.yaml`: Defines a trainer controller which computes loss at every step and loss consistently increases for three steps, then the training is stopped.
+- `trainercontroller_config_epoch.yaml`: Defines a epoch level trainer controller which computes loss at every epoch. The rule applied here is to compare a current epoch loss with previous epoch loss and it turns out to be more, then training is stopped.
```
Akash-Nayak, line 9:
```suggestion
- `trainercontroller_config_epoch.yaml`: Defines an epoch-level trainer controller, which computes loss at every epoch. The rule applied here is to compare the current epoch loss with the previous epoch loss, and if the current epoch loss turns out to be more, then the training is stopped.
```

examples/trainer-controller-configs/Readme.md
```diff
@@ -0,0 +1,11 @@
+# How-To
+To use one of these files with the trainer, execute the `sft_trainer.py` with the following option: 
+```
+--trainer_controller_config_file "examples/trainer-controller-configs/<file-name>"
+```
+
+# Note on trainer controller configuration examples
+- `trainercontroller_config_step.yaml`: Defines a trainer controller which computes loss at every step and loss consistently increases for three steps, then the training is stopped.
```
Akash-Nayak, line 8:
```suggestion
- `trainercontroller_config_step.yaml`: Defines a trainer controller, which computes loss at every step, and if the loss consistently increases for three steps, then the training is stopped.
```

tuning/trainercontroller/controllermetrics/steploss.py
```diff
@@ -0,0 +1,58 @@
...
+                consistently_increasing = False
+                break
+        avg_loss = stats.mean(cache['loss'])
+        std_loss = stats.stdev(cache['loss'])
+        return {self.get_name(): {'consistently_increasing': int(consistently_increasing), \
+            'average_loss': avg_loss,\
+            'std_loss': std_loss,\
+            'window': cache['loss']}}
+
+    def compute(self, training_state, training_args=None, metrics=None) -> dict:
```
kmehant, line 41: Do we want to introduce types to function arguments as well? (not critical)

tuning/trainercontroller/controllermetrics/steploss.py
```diff
@@ -0,0 +1,58 @@
...
+        """Computes the controller-metric (step-loss over window) and exposes the values of the variables used by the rules.
+
+        Args:
+            training_state: TrainerState object
+            training_args: TrainingArguments object
```
kmehant, line 46: I see training_args is as well optional here, may be you might want to regenerate the docstring. (not critical)

kmehant (commented):
not critical provided  small suggestions.

seshapad:
Comments from review:
- Fold the result marshalling for controller-metrics to the base class
- Handling the scenarios when eval() does not have the variables required to evaluate the rule - more discussion required
- eval() needs to be replaced by safer version

tuning/trainercontroller/controllermetrics/steploss.py
seshapad: done

tuning/trainercontroller/controllermetrics/steploss.py
seshapad: done

tuning/trainercontroller/callback.py
```diff
@@ -0,0 +1,237 @@
...
+    def __validate_config(self, config):
+        assert 'controllers' in config and len(config['controllers']) > 0, "List of controllers missing in config"
+        for c in config['controllers']:
+            assert 'name' in c, f"Controller should have a name"
+            assert ('triggers' in c) and (len(c['triggers']) > 0), f"Triggers not specified for controller {c['name']}"
+            assert ('rule' in c) and (len(c['rule']) > 0), f"Rule not specified for controller {c['name']}"
+            try:
+                compile(c['rule'], '<stdin>', 'eval')
+            except Exception as e:
+                raise SyntaxError(f"Rule for controller {c['name']} has this error {e}")
```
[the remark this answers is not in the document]
seshapad: compile() also throws a TypeError if c['rule'] is None. But, I have checked this condition before passing to compile(). 
Also addressed your comment on the error msg.

tuning/trainercontroller/callback.py
```diff
@@ -0,0 +1,237 @@
...
+            assert ('controller-metrics' in c) and (len(c['controller-metrics']) > 0), f"List of controller metrics missing for controller {c['name']}"
+            for cm in c['controller-metrics']:
+                assert ('name' in cm), f"Controller metric should have a name"
+                assert ('handler-class' in cm), f"Handler class not specified for controller metric {cm['name']}"
+            assert ('control-operations' in c) and (len(c['control-operations']) > 0), f"Control operations not specified for controller {c['name']}"
+            for k, v in c['control-operations'].items():
+                try:
+                    getattr(TrainerControl(), k)
```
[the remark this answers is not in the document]
seshapad: Addressed.

tuning/trainercontroller/controllermetrics/metricshandlerwithcache.py
```diff
@@ -0,0 +1,27 @@
+from . import MetricHandler
+
+class MetricHandlerWithCache(MetricHandler):
```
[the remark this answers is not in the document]
seshapad: Addressed.

tuning/trainercontroller/callback.py
```diff
@@ -0,0 +1,237 @@
...
+                        try:
+                            # Get the controller-metric class type
+                            class_type = getattr(contmetrics, cm['handler-class'])
+                            # Initialize the controller-metric instance
+                            if 'arguments' in cm:
+                                obj = class_type(cm['name'], **cm['arguments'])
+                            else:
+                                obj = class_type(cm['name'])
+                        except Exception as e:
+                            logger.fatal(e)
```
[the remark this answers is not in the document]
seshapad: Addressed.

tuning/trainercontroller/controllermetrics/steploss.py
```diff
@@ -0,0 +1,60 @@
...
+        """
+        size_of_log_history = len(training_state.log_history)
+        for i in range(size_of_log_history - 1, -1, -1):
+            log = training_state.log_history[i]
+            try:
+                self.add_to_cache(loss=log['loss'])
+            except Exception as e:
+                pass         
+            return self.compute_metrics_on_cache()
+        return None
```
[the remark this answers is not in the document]
seshapad: Addressed.

tuning/policydrivencontroller/__init__.py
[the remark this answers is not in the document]
seshapad: done

tuning/policydrivencontroller/controllermetrics/metrics.py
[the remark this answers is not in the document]
seshapad: done

tuning/policydrivencontroller/controllermetrics/metrics.py
```diff
@@ -0,0 +1,244 @@
...
+    def validate(self, training_args) -> bool:
+        pass
+    
+    def compute(self, training_state, training_args=None, metrics=None) -> dict:
+        pass   
+
+class WindowStepLoss(MetricHandler):
+    """Implements the controller metric which evaluates loss-per-step over a user-defined window"""
+    
+    def __init__(self, name, args):
```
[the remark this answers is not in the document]
seshapad: done

tuning/policydrivencontroller/controllermetrics/metrics.py
```diff
@@ -0,0 +1,244 @@
...
+        consistently_increasing = True
+        for i in range(len(window)-1):
+            if window[i] > window[i+1]:
+                consistently_increasing = False
+                break
+        w = np.array(window)
+        avg_loss = np.mean(w)
+        std_loss = np.std(w, dtype=np.float64)
+        first_and_last_loss = window[0] < window[len(window)-1]
+        exposed_data = {self.__name + '_' + 'consistently_increasing': int(consistently_increasing), \
```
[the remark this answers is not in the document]
seshapad: done

tuning/policydrivencontroller/controllermetrics/metrics.py
```diff
@@ -0,0 +1,244 @@
...
+    def validate(self, training_args):
+        """Validate the training arguments (e.g logging_steps) are compatible with the computation of this metric
+
+        Args:
+            training_args: Training arguments
+
+        Returns:
+            bool
+        """
+        return training_args.logging_strategy == 'steps' and \
```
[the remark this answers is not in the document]
seshapad: done

tuning/policydrivencontroller/controllermetrics/metrics.py
```diff
@@ -0,0 +1,244 @@
...
+        loss_values = [l['loss'] for l in training_state.log_history if 'loss' in l]
+        n = len(loss_values)
+        if n == 0:
+            logger.info('(N=0) Number of LOSS values: %d, Given window size: %d' % (n, self.__args['window-size']))
+            return None
+        if n <= self.__args['window-size']:
+            logger.info('(N<W) Number of LOSS values: %d, Given window size: %d' % (n, self.__args['window-size']))
+            return None
+        window = loss_values[n-self.__args['window-size']:n]
+        consistently_increasing = True
```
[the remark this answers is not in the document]
seshapad: done

tuning/policydrivencontroller/controllermetrics/metrics.py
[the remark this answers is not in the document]
seshapad: done

tuning/policydrivencontroller/controllermetrics/metrics.py
```diff
@@ -0,0 +1,244 @@
...
+from transformers.utils import logging
+from collections import deque
+import math
+import copy
+
+logger = logging.get_logger(__name__)
+
+class MetricHandler:
+    """Base class for the controller-metrics"""
+    def validate(self, training_args) -> bool:
```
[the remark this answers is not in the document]
seshapad: Done

tuning/policydrivencontroller/controllermetrics/metrics.py
```diff
@@ -0,0 +1,244 @@
...
+        pass
+    
+    def compute(self, training_state, training_args=None, metrics=None) -> dict:
```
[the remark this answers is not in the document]
seshapad: Done

tuning/policydrivencontroller/controllermetrics/metrics.py
```diff
@@ -0,0 +1,244 @@
...
+            metrics: [optional] metrics data
+
+        Returns:
+            dict
+        """
+        # Compute the metric using the training state
+        loss_values = [l['loss'] for l in training_state.log_history if 'loss' in l]
+        n = len(loss_values)
+        if n == 0:
+            logger.info('(N=0) Number of LOSS values: %d, Given window size: %d' % (n, self.__args['window-size']))
```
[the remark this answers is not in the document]
seshapad: Done.

examples/trainer-controller-configs/Readme.md
```diff
@@ -0,0 +1,11 @@
+# How-To
```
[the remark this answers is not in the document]
seshapad: Addressed

tuning/trainercontroller/callback.py
```diff
@@ -0,0 +1,237 @@
...
+            assert ('rule' in c) and (len(c['rule']) > 0), f"Rule not specified for controller {c['name']}"
+            try:
+                compile(c['rule'], '<stdin>', 'eval')
+            except Exception as e:
+                raise SyntaxError(f"Rule for controller {c['name']} has this error {e}")
+            assert ('controller-metrics' in c) and (len(c['controller-metrics']) > 0), f"List of controller metrics missing for controller {c['name']}"
+            for cm in c['controller-metrics']:
+                assert ('name' in cm), f"Controller metric should have a name"
+                assert ('handler-class' in cm), f"Handler class not specified for controller metric {cm['name']}"
+            assert ('control-operations' in c) and (len(c['control-operations']) > 0), f"Control operations not specified for controller {c['name']}"
```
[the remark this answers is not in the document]
seshapad: Removed altogether.

requirements.txt
```diff
@@ -12,4 +12,5 @@ ninja
 peft>=0.8.0
 datasets>=2.15.0
 fire
+typing
```
kmehant, line 15: typing is part of the standard library from Python>=3.5  (https://docs.python.org/3/library/typing.html#typing-support-for-type-hints)

tuning/trainercontroller/callback.py
```diff
@@ -0,0 +1,298 @@
...
+
+CONTROLLER_NAME_KEY = "name"
+CONTROLLER_TRIGGERS_KEY = "triggers"
+CONTROLLER_RULE_KEY = "rule"
+CONTROLLER_OPERATIONS_KEY = "operations"
+
+class TrainerControllerCallback(TrainerCallback):
+    """Implements the policy driven trainer loop control based on policy definition file and metrics"""
+    
+    def __init__(self, trainer_controller_config: dict):
```
kmehant, line 52:
```suggestion
    def __init__(self, trainer_controller_config: Union[dict, str]):
```

seshapad: Addressed.

requirements.txt
seshapad: Are you suggesting `typing` be removed from requirements?

kmehant: @seshapad 
My thinking is that the other dependencies like `transformers>=4.34.1` are tested stable for Python 3.8+. Given that it would be recommended to use Python 3.8+ which comes packed with typing as part of the standard lib. So I was just thinking should we be removing this from requirements?

tuning/trainercontroller/operations/operation.py
```diff
@@ -0,0 +1,45 @@
...
+        """Validates the action by checking if it valid action or not.
+
+        Args:
+            action: str. String depicting the action.
+
+        Returns:
+            bool. Indicates True if valid. If not, returns False.
+        """
+        if action not in self.valid_actions:
+            return False
```
kmehant, line 29:
```suggestion
        return action in self.valid_actions
```

kmehant: @seshapad Let me know if I have missed something, the function would never be returning True with the current code isn't it?

tuning/trainercontroller/operations/operation.py
```diff
@@ -0,0 +1,45 @@
...
+
+    def act(self, action: str, **kwargs):
+        """Validates the action and invokes it.
+
+        Args:
+            action: str. String depicting the action.
+            kwargs: List of arguments (key, value)-pairs.
+        """
+        if self.validate(action):
+            raise ValueError(f"Invalid operation {action}")
```
kmehant, line 39: @seshapad if valid we want to raise an Exception is it? Am I missing something?

tuning/trainercontroller/operations/operation.py
```diff
@@ -0,0 +1,45 @@
...
+    """Base class for operations"""
+
+    def __init__(self):
+        """Initializes the HuggingFace controls. In this init, we follow the convention that
+        every action should preceed with prefix `should_`. If so, it is treated as a valid
+        action.
+        """
+        self.valid_actions = {}       
+        for action_name, action_method in inspect.getmembers(self, predicate=inspect.ismethod):
+            if re.search(r'^should_', action_name) is not None:
```
kmehant, line 16: method name just being `should_` would also pass through isn't it? Do we want to allow that?
```suggestion
            if re.search(r'^should_.+', action_name) is not None:
```

seshapad: Addressed.

tuning/trainercontroller/operations/operation.py
seshapad: Addressed

tuning/trainercontroller/operations/operation.py
seshapad: Addressed

tuning/trainercontroller/callback.py
```diff
@@ -0,0 +1,353 @@
...
+        kwargs["control"] = control
+
+        # Check if there any metrics listed in the configuration
+        if CONTROLLER_METRICS_KEY not in self.trainer_controller_config:
+            logger.warn("Trainer controller config has no metrics.")
+
+        # Metric handler validation and registration is performed here.
+        for metric_name, metric_config in self.trainer_controller_config[CONTROLLER_METRICS_KEY].items():
+            # Get the metric class name from the config section.
+            metric_handler_name = next(iter(metric_config.keys()))
```
alex-jw-brooks, line 217: Trying to make sure I understand this bit - the iterable created out of that metric's config is of length one, correct? I.e., this is functionally the same as `list(metric_config.keys())[0]`?

tuning/trainercontroller/callback.py
```diff
@@ -0,0 +1,353 @@
...
+        self._compute_metrics(event_name, **kwargs)
+        self._take_control_actions(event_name, **kwargs)
+
+    def _validate_rule(self, rule):
+        """Validates the rule to check if there are any import attempts 
+
+        Returns:
+            bool
+        """
+        return re.search(r'__', rule) is None
```
alex-jw-brooks, line 193: I think I don't understand this correctly - can you explain how this regex is checking import attempts (e.g., with an example of something that is caught by this that is potentially malicious)?

tuning/trainercontroller/callback.py
```diff
@@ -0,0 +1,353 @@
...
+            # Get the handler class using the metric class name.
+            if metric_handler_name not in self.metric_handlers:
+                raise KeyError(f"Undefined metric handler {metric_handler_name}")
+            metric_handler = self.metric_handlers[metric_handler_name]
+            # Get the metric handler class arguments specified in the config.
+            metric_args = metric_config[metric_handler_name]
+            if metric_args is None:
+                metric_args = {}
+            # Metric handler instance is created here.
+            obj = metric_handler(name=metric_name, **metric_args, **kwargs)
```
alex-jw-brooks, line 227: What is the reason for passing kwargs through here? If something is written weirdly, there could be potential key collisions between that and the metric args being kwarg expanded - is there an advantage to passing it to a metric handler in this way?

tuning/trainercontroller/callback.py
```diff
@@ -0,0 +1,353 @@
...
+                    control = Control(name=controller[CONTROLLER_NAME_KEY], rule = compile(controller[CONTROLLER_RULE_KEY], '', 'eval'), operation_actions = [])
+                    for control_operation_name in controller[CONTROLLER_OPERATIONS_KEY]:
+                        if control_operation_name not in self.operation_actions:
+                            raise KeyError(f"Invalid operation {control_operation_name} for control {controller[CONTROLLER_NAME_KEY]}")
+                        control.operation_actions.append(self.operation_actions[control_operation_name])
+                    if event_name not in self.control_actions_on_event:
+                        self.control_actions_on_event[event_name] = []
+                    self.control_actions_on_event[event_name].append(control)
+                        
+        self._actions_on_event(event_name='on_init_end', **kwargs)
```
alex-jw-brooks, line 272: Same question here - in what situation is it useful invoke an action on init end if the focus of this is metric based and dependent on things that are being produced as training progresses?

alex-jw-brooks (changes requested):
Hey @seshapad @ashokponkumar @dushyantbehl - thanks for your patience and sorry for the late turnaround on the review, this is a pretty complex PR and I wanted to make sure I understood it properly to figure out what needs to happen for future work.

I added some comments, but in general I think the big things to answer are as follows:
- An ADR for this should be added; you can find the template [here](https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/architecture_records/template.md). The description will help cover most of it I think, but it would be awesome to add some explicit points of clarification for things that might be hard to understand for new devs that are looking at this, especially if they aren't familiar with callback overall. Some examples of things that come to mind:
    - The difference between an event and an operation action, i.e., the `on_X` methods vs the `should_X`
    - The expected return value of things `compute`d by implementations of `metric_handler` - I understand this is dependent on the rule being executed, since that makes it tied to the config, it can be hard to understand best practices


- How to consider metrics that are dependent on history, i.e., how we would actually use this framework to implement the normal early stopping callback's behavior, not just terminate early based on a hardcoded loss. Currently I don't think this is possible because the loss metric handler implementation is only producing the last loss and not maintaining any kind of history - is that correct?

- How we should use this framework for evaluation metrics, i.e., stopping early based on eval metrics. I suspect this might be easy and would basically be the same as the loss metric handler + on_eval event, with the primary difference being the key pulled out of the traversed logs, but it would be good to confir

tuning/trainercontroller/callback.py
```diff
@@ -0,0 +1,241 @@
...
+                raise SyntaxError(f"Rule [{c['rule']}] for controller {c['name']} has this error {e}")
+            assert ('controller-metrics' in c) and (len(c['controller-metrics']) > 0), f"List of controller metrics missing for controller {c['name']}"
+            for cm in c['controller-metrics']:
+                assert ('name' in cm), f"Controller metric should have a name"
+                assert ('handler-class' in cm), f"Handler class not specified for controller metric {cm['name']}"
+            assert ('control-operations' in c) and (len(c['control-operations']) > 0), f"Control operations not specified for controller {c['name']}"
+            for k, v in c['control-operations'].items():
+                for prefix in TrainerControllerCallback.__control_op_prefixes:
+                    if not k.startswith(prefix):
+                        raise AttributeError(f"Control operation {k} is not invalid for controller {c['name']}")
```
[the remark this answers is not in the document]
seshapad: @alex-jw-brooks This code does not exist in the PR. We had removed a while ago.

tuning/sft_trainer.py
```diff
@@ -92,7 +92,8 @@ def train(
     model_args: configs.ModelArguments,
     data_args: configs.DataArguments,
     train_args: configs.TrainingArguments,
-    peft_config: Optional[  # pylint: disable=redefined-outer-name
+    trainer_controller_args: configs.TrainerControllerArguments,
```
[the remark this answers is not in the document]
seshapad: @alex-jw-brooks I am assuming you are referring to `TrainerControllerArguments`. This already has default value of `None`. Here is the reference: 

https://github.com/foundation-model-stack/fms-hf-tuning/blob/b70dbbdce08909cd51fa86451ef651022097c430/tuning/config/configs.py#L97

tuning/trainercontroller/operations/hfcontrols.py
```diff
@@ -0,0 +1,39 @@
...
+    def __init__(self, **kwargs):
+        """Initializes the HuggingFace controls. In this init, the fields with `should_` of the
+        transformers.TrainerControl data class are extracted, and for each of those fields, the
+        control_action() method's pointer is set, and injected as a class member function.
+
+        Args:
+            kwargs: List of arguments (key, value)-pairs
+        """
+        for control_field in fields(TrainerControl):
+            if re.search(r'^should_', control_field.name) is not None:
```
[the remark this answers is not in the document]
seshapad: @alex-jw-brooks The reason we went with a regex for the following reasons:
1. It is to avoid name definitions like `should_` (which seems to be a valid function name in python)
2. To leave the door open for newer operations to be added later (in which case, we simply expand the scope of regex accordingly).
3. Also, as a side note, this is a init time operation, and the regex will not bog down the trainer loop.

tuning/trainercontroller/operations/hfcontrols.py
```diff
@@ -0,0 +1,39 @@
...
+import inspect
+from dataclasses import fields
+
+# Third Part
+from transformers import TrainerControl
+
+# Local
+from tuning.trainercontroller.operations import Operation
+
+class HFControls(Operation):
```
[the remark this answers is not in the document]
seshapad: @alex-jw-brooks The intent of having a layer of abstraction for `custom operations` is as follows:
1. In future, this will help expand the framework beyond huggingface TrainerControl to other usecases (e.g. notifications to end user on training anomalies).
2. It allows for a registration mechanism which helps developers add custom operations without changing the `trainercontroller framework` code.

seshapad:
> Hey @seshapad @ashokponkumar @dushyantbehl - thanks for your patience and sorry for the late turnaround on the review, this is a pretty complex PR and I wanted to make sure I understood it properly to figure out what needs to happen for future work.
> 
> I added some comments, but in general I think the big things to answer are as follows:
> 
>     * An ADR for this should be added; you can find the template [here](https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/architecture_records/template.md). The description will help cover most of it I think, but it would be awesome to add some explicit points of clarification for things that might be hard to understand for new devs that are looking at this, especially if they aren't familiar with callback overall. Some examples of things that come to mind:
>       
>       * The difference between an event and an operation action, i.e., the `on_X` methods vs the `should_X`
>       * The expected return value of things `compute`d by implementations of `metric_handler` - I understand this is dependent on the rule being executed, since that makes it tied to the config, it can be hard to understand best practices
> 
>     * How to consider metrics that are dependent on history, i.e., how we would actually use this framework to implement the normal early stopping callback's behavior, not just terminate early based on a hardcoded loss. Currently I don't think this is possible because the loss metric handler implementation is only producing the last loss and not maintaining any kind of history - can this be changed, since early stopping is the main motivation for this framework in general?
> 
>     * How we should use this framework for evaluation metrics, i.e., stopping early based on eval metrics. I suspect this might be easy and would basically be the same as the loss metric handler + on_eval event, with the primary difference being the key pulled out of the traversed logs, but it would be good to confir

@alex-jw-brooks Thank you very much for your insightful comments and taking the time to review the PR in depth. We have attempted to address your comments:
1. We have added an ADR document as part of this PR. It is available [here](https://github.com/foundation-model-stack/fms-hf-tuning/blob/7691921e97f6b827063860597ef80b5e45bb9bf6/architecture_records/001-trainer-controller-framework.md).

1. As the current PR is already large and required time to review and merge, we planned to include additional metrics in the subsequent PRs once this PR gets merged.

1. Yes. The eval metrics based stopping could be implemented using the current framework. We plan to add it in subsequent PRs.

tuning/trainercontroller/callback.py
seshapad: @alex-jw-brooks The reasons for the above signature in the `metric_handler` are as follows:
1. If you notice, the `metric_handler` is actually a class type that is stored following registration (every metric class type and corresponding arguments are stored and referred to using a metric name). Following code [link](https://github.com/foundation-model-stack/fms-hf-tuning/blob/7691921e97f6b827063860597ef80b5e45bb9bf6/tuning/trainercontroller/callback.py#L220) illustrates this. The  arguments could come from the trainercontroller config file. They could be any set of key-value pairs. These are passed using ` **metric_args`.
1. In addition to `**metric_args`, we also have the `**kwargs` coming from which include trainer arguments, state and controls which could potentially be used by a metric during initialization. 

Hence, we need both `**metric_args` and `**kwargs`  to remain separate as they come from different sources.

However, as you pointed out, there could be potential collision of keys between these two sources. To remedy this without compromising on the flexibility of the above approach, we will document a warning in the code as well as in the ADR, or if required add a README to inform the users to be careful in choosing the argument names.

tuning/trainercontroller/callback.py
```diff
@@ -0,0 +1,353 @@
...
+                    self.trainer_controller_config[OPERATIONS_KEY] = {}
+                if operation_name not in self.trainer_controller_config[OPERATIONS_KEY]:
+                    self.trainer_controller_config[OPERATIONS_KEY][operation_name] = \
+                        default_operations_config[OPERATIONS_KEY][operation_name]
+
+        # Load list of valid events
+        self.valid_events = set()
+        for callback_method_name, _ in inspect.getmembers(self, predicate=inspect.ismethod):
+            if re.search(r'^on_', callback_method_name) is not None:
+                self.valid_events.add(callback_method_name)
```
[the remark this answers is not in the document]
seshapad: A debug log has been added.

tuning/trainercontroller/callback.py
seshapad: @alex-jw-brooks From what we understand on the use of `eval()`, sanitizing the input is one of the ways to mitigate the risk of using `eval()`.

To that end, we have added this additional layer of check to prevent the use of variables like `__code__`, `__subclasses__` etc which could potentially leak information.

tuning/trainercontroller/callback.py
seshapad: Yes. That is correct.

tuning/trainercontroller/callback.py
seshapad: @alex-jw-brooks We could perform actions like validate configs and  if validation fails `stop the training` before it starts or Notify user on training started or is about start. It has some uses and hence we have left it to cover any such future use cases.

tuning/trainercontroller/callback.py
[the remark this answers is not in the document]
alex-jw-brooks: Yup thanks - this is also an old comment

seshapad:
@alex-jw-brooks We fixed most of the lint comments, but we were left with the following three:

1. The below warning seems to be coming from the `sft_trainer.py`
  ```
  tuning/sft_trainer.py:98:4: W0621: Redefining name 'peft_config' from outer scope (line 40) (redefined-outer-name)
  ```

1. This one might be because we are using more than 7 attributes in our class. But they are required.
  ```
  tuning/trainercontroller/callback.py:52:0: R0902: Too many instance attributes (8/7) (too-many-instance-attributes)
  ```

1. This one is because of the use of `eval()`:
  ```
  tuning/trainercontroller/callback.py:173:24: W0123: Use of eval (eval-used)
  ```

Please suggest how we could address these.
cc: @ashokponkumar

tuning/trainercontroller/operations/hfcontrols.py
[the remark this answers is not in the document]
alex-jw-brooks: Cool, just wanted to check since `^should_` also allows `should_` - I see it's been changed now, sounds good to me, thanks!

tuning/trainercontroller/callback.py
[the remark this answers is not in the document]
alex-jw-brooks: Thanks!

tuning/trainercontroller/callback.py
alex-jw-brooks: Thanks for the clarification!

tuning/trainercontroller/callback.py
alex-jw-brooks: Ahhh I understand now, I forgot about the `__import__` function hook and thought that this meant the `import` statement. Thanks, that helps, makes sense

tuning/trainercontroller/callback.py
alex-jw-brooks: Makes sense, thanks for the clarification!

seshapad:
> Hey @seshapad - thanks for addressing the requested changes and adding the ADR! For the pylint errors:
> 
>     1. Do you get this on `main`? I am a bit surprised to see it here, since this PR doesn't really touch the peft config stuff, but it probably comes from the local name and importing `peft_config` directly from `tuning.config`. If it is on `main`, I'll fix this one in a separate PR - I'll also try running the format check stuff here
> 
> 
> For 2 & 3, we can disable them using a comment [like this](https://docs.pylint.org/faq.html#do-i-have-to-remember-all-these-numbers), since it seems like these are both needed.
> 
> I think everything else looks good! Great work 🥳

@alex-jw-brooks @Ssukriti We have addressed the issues raised during linting, formatting and testing. The `make fmt`, `make lint` and `make test` have been run to verify it. The changes have also been pushed to the PR.
If you are satisfied with the PR, We request you to take the process forward in triggering the workflow checks, approval and merge.

alex-jw-brooks (approved):
LGTM! Thank you very much! 🚀

PR #46, opened 2024-02-15
docs: lora and getting modules list

anhuong:


README.md
```diff
@@ -113,6 +113,102 @@ tuning/sft_trainer.py \
...
+```py
+LoraConfig(
+    r=8,
+    lora_alpha=16,
+    target_modules=['q_proj', 'v_proj'],
+    lora_dropout=0.05
+)
+```
+
+Notice the `target_modules` that are set are the default values. `target_modules` are the names of the modules to apply the adapter to. If this is specified, only the modules with the specified names will be replaced. When passing a list of strings, either an exact match will be performed or it is checked if the name of the module ends with any of the passed strings. If this is specified as `all-linear`, then all linear/Conv1D modules are chosen, excluding the output layer. If this is not specified, modules will be chosen according to the model architecture. If the architecture is not known, an error will be raised — in this case, you should specify the target modules manually. See [HuggingFace docs](https://huggingface.co/docs/peft/en/package_reference/lora#peft.LoraConfig) for more details.
```
Ssukriti, line 157: do we support target_modules as 'all-linear' or ['all-linear'] ? I dont remember now, but can you check our definition to see how we accept the target modules from command line?

README.md
```diff
@@ -113,6 +113,102 @@ tuning/sft_trainer.py \
...
+        (post_attention_layernorm): LlamaRMSNorm()
+      )
+    )
+    (norm): LlamaRMSNorm()
+  )
+  (lm_head): Linear(in_features=4096, out_features=32000, bias=False)
+)>
+```
+
+You can specify attention or linear layers. With the CLI, you can specify layers with `--target_modules "q_proj" "v_proj" "k_proj" "o_proj"` or `--target_modules "all-linear"`.
```
Ssukriti, line 210: hmm according to our config we accept List[str] ? https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/tuning/config/peft_config.py#L7

README.md
```diff
@@ -113,6 +113,102 @@ tuning/sft_trainer.py \
 
 For `GPTBigCode` models, Hugging Face has enabled Flash v2 and one can simply replace the `'LlamaDecoderLayer'` with `'GPTBigCodeBlock'` in `tuning/config/fsdp_config.json` for proper sharding of the model.
 
+### Lora Tuning Example
```
Ssukriti, line 116:
```suggestion
### LoRA Tuning Example
```

README.md
Ssukriti: I think we only accept List[str]

README.md
anhuong: It depends on what we want to document, the usage via Python or the usage via CLI. In Shell, this is how one would pass a list via command line. So what is interpreted here (which I tested) is:

--target_modules "all-linear" 
becomes 
LoraConfig(target_modules=['all-linear'])
becomes 
LoraConfig(target_modules='all-linear')
to pass into SFTTrainer in the form accepted.

whereas

--target_modules "q_proj" "v_proj" "k_proj" "o_proj"
becomes
LoraConfig(target_modules=['q_proj', 'v_proj', 'k_proj', 'o_proj'])
becomes
LoraConfig(target_modules={'k_proj', 'v_proj', 'q_proj', 'o_proj'})

But you are correct that only a list is accepted so if using via Python this should be passed as `['all-linear']` (relevant to your previous comment)

README.md
anhuong: see below comment for details

README.md
Ssukriti: oh understood, probably what you documented is fine then. If we see confusion we will clarify it further

Ssukriti (approved):
thank you Anh!

PR #49, opened 2024-02-19
feat: custom callbacks for train() and return TrainOutput plus model_load_time

### LINKED ISSUE #33: Add support for collecting metrics programmatically
**Issue**

I would like to use fms-hf-tuning to collect system level metrics while finetuning models, these include model load time, as well as device related metrics (like these that AIM collects).

One way would be to rely on just the measurements that AIM collects by pointing fms-hf-tuning to an AIM server then contacting AIM to retrieve the data. However, this is a bit restricting in that we cannot collect custom data, collect system metrics at a period other than the one that AIM is using - 30 seconds, or collect data at all if we don't spin up an AIM server.

A more convenient solution would be to allow providing an optional parameter to `train()` for which can contain a list of callbacks

https://github.com/foundation-model-stack/fms-hf-tuning/blob/fc070604e788702fb18af935ed4e99f24278412e/tuning/sft_trainer.py#L29-L34

In the same spirit I'd like to get access to the `TrainingOutput` object that `sft_trainer.train()` returns here (input_tokens_per_second, train_runtime, etc) :

https://github.com/foundation-model-stack/fms-hf-tuning/blob/fc070604e788702fb18af935ed4e99f24278412e/tuning/sft_trainer.py#L168

(just by returning the output of trainer.train() as the output of train()).

**Done when**

- [ ] support collecting custom metrics via custom callbacks
- [ ] return the output of `trainer.train()` to the caller of `train()`

VassilisVassiliadis:
This PR resolves #33 

Changes:

- support for custom callbacks
- record time it took to load model
- train() method now returns output of SFTTrainer.train() as well as `model_load_time`

### DIFF
--- a/tuning/sft_trainer.py
+++ b/tuning/sft_trainer.py
@@ -1,8 +1,9 @@
 # Standard
 from datetime import datetime
-from typing import Optional, Union
+from typing import Optional, Union, List, NamedTuple
 import json
 import os
+import time
 
 # Third Party
 from peft.utils.other import fsdp_auto_wrap_policy
@@ -38,6 +39,12 @@ class PeftSavingCallback(TrainerCallback):
 
         if "pytorch_model.bin" in os.listdir(checkpoint_path):
             os.remove(os.path.join(checkpoint_path, "pytorch_model.bin"))
+
+
+class EnhancedTrainOutput(NamedTuple):
+    train_output: transformers.trainer.TrainOutput
+
+    model_load_time: float
 
 
 class FileLoggingCallback(TrainerCallback):
@@ -84,7 +91,8 @@ def train(
     peft_config: Optional[
         Union[peft_config.LoraConfig, peft_config.PromptTuningConfig]
     ] = None,
-):
+    callbacks: Optional[List[TrainerCallback]] = None,
+) -> EnhancedTrainOutput:
     """Call the SFTTrainer
 
     Args:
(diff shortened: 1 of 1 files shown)


dushyantbehl:
@VassilisVassiliadis can this be closed if you agree with the design and direction of PR #50 ?

tuning/sft_trainer.py
```diff
@@ -260,7 +277,13 @@ def train(
...
             model
         )
-    trainer.train()
+
+    train_output: "transformers.trainer.TrainOutput" = trainer.train()
+
+    return EnhancedTrainOutput(
+        train_output=train_output,
+        model_load_time=model_load_time,
+    )
```
VassilisVassiliadis, line 286: @dushyantbehl I need the `train()` method to return the TrainOutput + model_load_time.

dushyantbehl: do you have any specific need for that one metric coming out of the `train()` function?

Could you not fetch it from the tracker?

VassilisVassiliadis: We'd like to programmatically execute a large array of fms-hf-tuning experiments to collect data (things like performance of model, system metrics, etc). Some of these runs may take place on machines which do not have network connectivity. Other runs we may not want to register to AIM at all as they contain experimental code/models/datasets and we wouldn't want to pollute the AIM database with data that we're not sure we'd like to keep around.

As a result, we need to collect the measured metrics (trainoutput + model_load_time) straight  from the return value of the `train()` method.

Issue #34, opened 2024-02-07
Add support for registering custom AIM metadata


VassilisVassiliadis:
**Issue**

As a user that is running a large finetuning campaign of runs I would like to annotate each individual run with metadata. This way I'll be able to query my AIM server for my finetuning experiments in the future.

**Done when**
- [ ] add new parameter `aim_metadata` to the train() method for optional metadata that fms-hf-tuning propagates to AIM
- [ ] commandline argument which can point to a `json` file with aim metadata

dushyantbehl:
@VassilisVassiliadis this also can be closed now... 🥂

PR #50, opened 2024-02-19
feat: change tracker API to initialize tracker early and track additional metrics.

dushyantbehl:
1. This PR adds a tracker API with Aimstack as the default tracker. This is simple plug and play architecture which can support multiple trackers.
1. The tracker config is now taken as command line arguments (making it easier for any automation to pass tracker arguments).
1. With the new API I have added support to track any additional metrics of interest.
1. As an example I have added one single line to track model_load_time in AIM possibly fixing https://github.com/foundation-model-stack/fms-hf-tuning/issues/33 
1. Also bumps Aim version to 3.18.1 which is more stable and newer release.
See. https://github.com/foundation-model-stack/fms-hf-tuning/pull/42

Example of how new api can be invoked

```
 torchrun --nnodes=1 --nproc_per_node=2 --master_port=1234 tuning/sft_trainer.py --tokenizer_name_or_path ${MODEL_PATH} --model_name_or_path ${MODEL_PATH} --data_path ${DATA_PATH} --use_peft --bf16 True --output_dir ${OUTPUT_PATH} --num_train_epochs 1 --per_device_train_batch_size 1 --per_device_eval_batch_size 1 --gradient_accumulation_steps 8 --evaluation_strategy "no" --save_strategy "steps" --save_steps 2000 --save_total_limit 1 --learning_rate 2e-5 --weight_decay 0. --warmup_ratio 0.03 --lr_scheduler_type "cosine" --logging_steps 1 --fsdp "full_shard auto_wrap" --fsdp_config tuning/config/fsdp_config.json --bf16 True --response_template "\n### Response:" --dataset_text_field "output" --tracker aim --aim_repo /data/aim --experiment sft-llama7b-test 
```


dushyantbehl:
cc @VassilisVassiliadis comments are welcome

tuning/sft_trainer.py
```diff
@@ -115,13 +115,28 @@ def train(
...
+
+    model_load_time = time.time()
     model = AutoModelForCausalLM.from_pretrained(
         model_args.model_name_or_path,
         cache_dir=train_args.cache_dir,
         torch_dtype=get_torch_dtype(model_args.torch_dtype),
         use_flash_attention_2=model_args.use_flash_attn,
     )
+    model_load_time = time.time() - model_load_time
+    tracker.track(metric=model_load_time, name='model_load_time')
```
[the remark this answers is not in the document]
dushyantbehl: @VassilisVassiliadis can you please explain what you had in mind for the metadata. I can certainly add a parameter like `extra_metadata` which can be passed to the tracker but I am trying to understand what can be covered other than the grouping via experiment name.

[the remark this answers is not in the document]
VassilisVassiliadis: The code that I linked annotates an experiment with key values which show up when you view the run on aim. 

You can imagine a dictionary that looks like this:

```json
{"gpu_model": "A100-80GB", "dataset_id": "my-amazing-dataset"}
```

This is nifty because you can query AIM to receive a list of all experiments that used a certain `gpu_model` or `dataset_id`, etc

If you prefer, I can open a PR for #34 after this one goes through so as not to bundle together too many changes.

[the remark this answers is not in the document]
dushyantbehl: That is small enough Vassilis let me merge this in with this PR. Thanks for the clarification.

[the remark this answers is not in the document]
dushyantbehl: do you suspect this extra metadata to be pased via main as arguments? as a file? or just to the train function.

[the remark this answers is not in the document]
VassilisVassiliadis: Personally I'd be satisfied with just the `aim_metadata` parameter in `train()` as I use fms-hf-tuning via its python API - i.e. directly invoke the train() method from inside my python code.

That said, users who run sft_trainer.py and have access to an AIM Server may want use something like `--aim_metadata` (or equivalent) to associate their run with metadata.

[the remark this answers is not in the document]
dushyantbehl: @VassilisVassiliadis check now. Added a new API in tracker which can solve your problem. 

I have added a command line to sft trainer for extra metadata which accepts a loose json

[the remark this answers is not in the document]
VassilisVassiliadis: The one thing I'd change inside `train()` is guard accessing `tracker` via an `if tracker is not None:` statement just because the default value of tracker is None: https://github.com/foundation-model-stack/fms-hf-tuning/blob/5086918887b02c8ebd064aac9e19dc70625fe474/tuning/sft_trainer.py#L95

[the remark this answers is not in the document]
dushyantbehl: Changed the default so we won't have to say `if tracker is not None` everywhere we track metrics.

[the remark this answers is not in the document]
VassilisVassiliadis: I think it'll be better to  better to keep the default value as `None` and have a check at the top of the function which instantiates the `tracker` object if its value is zero. This way the code doesn't create 2 tracker objects, one when python defines the `train()` method and another when `main()` calls `train()`.

tuning/sft_trainer.py
```diff
@@ -83,14 +84,15 @@ def _track_loss(self, loss_key, log_file, logs, state):
...
-
 def train(
     model_args: configs.ModelArguments,
     data_args: configs.DataArguments,
     train_args: configs.TrainingArguments,
     peft_config: Optional[
         Union[peft_config.LoraConfig, peft_config.PromptTuningConfig]
     ] = None,
+    callbacks: Optional[List[TrainerCallback]] = None,
+    tracker: Optional[Tracker] = Tracker() # default tracker is dummy tracker
```
VassilisVassiliadis, line 95:
```suggestion
    tracker: Optional[Tracker] = None
```

tuning/sft_trainer.py
```diff
@@ -105,8 +107,6 @@ def train(
     """
     run_distributed = int(os.environ.get("WORLD_SIZE", "1")) > 1
```
VassilisVassiliadis, line 108:
```suggestion
    if tracker is None:
       tracker = Tracker()
    
    run_distributed = int(os.environ.get("WORLD_SIZE", "1")) > 1
```

VassilisVassiliadis (changes requested):
I was thinking of something along these lines (need to double check the indentation as I used the web interface to suggest the changes)

dushyantbehl:
@VassilisVassiliadis the design and some functions had changed due to inconsistency with FSDP runs. I feel most of your ask is incorporated now. Can you check again if you have any other questions. 

This argument `--exp_metadata '{"gpu_model": "A100-80GB", "dataset_id": "my-amazing-dataset"}'` passed to the run will result in the parameters recorded like this. 

image

VassilisVassiliadis:
Thanks! this looks good to me

ashokponkumar:
@alex-jw-brooks PTAL

tuning/sft_trainer.py
```diff
@@ -83,14 +84,16 @@ def _track_loss(self, loss_key, log_file, logs, state):
...
-
 def train(
     model_args: configs.ModelArguments,
     data_args: configs.DataArguments,
     train_args: configs.TrainingArguments,
     peft_config: Optional[
         Union[peft_config.LoraConfig, peft_config.PromptTuningConfig]
     ] = None,
+    callbacks: Optional[List[TrainerCallback]] = None,
+    tracker: Optional[Tracker] = None,
```
fabianlim, line 95: Since the `train` function originall accepts only configurations by design, I feel we need a strong reason to allow it to also include objects. The more natural way would be to accept one "tracker config" input

tuning/sft_trainer.py
```diff
@@ -286,23 +296,67 @@ def main(**kwargs):
...
+    else:
+        tracker_config=None
+
+    # Initialize callbacks
+    file_logger_callback = FileLoggingCallback(logger)
+    peft_saving_callback = PeftSavingCallback()
+    callbacks = [peft_saving_callback, file_logger_callback]
+
+    # Initialize the tracker
+    tracker = get_tracker(tracker_name, tracker_config)
```
fabianlim, line 335: why not just have a `tracker_install_callbacks` function somewhere to reduce 3 lines into 1

tuning/sft_trainer.py
ashokponkumar: Using a yaml as config was the first thought that came too. But given HF has chosen to use arguments for most configurations we went with it as a design choice. But if in fms-hf-tuning we choose to use config files for configs, we can do that too.

tuning/sft_trainer.py
ashokponkumar: Agreed. Here too it's a design choice we have to make. We have tried to keep it consistent with how config is managed in fms-hf-tuning here too. This has tried to follow the pattern used in the peft config block above. But if we choose to have a different approach we can follow that.

tuning/sft_trainer.py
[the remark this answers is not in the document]
dushyantbehl: Done

tuning/sft_trainer.py
dushyantbehl: This can be done @fabianlim its just a minor nit but I would like to collect these changes and do them once post review.

tuning/sft_trainer.py
dushyantbehl: Its simple choice to either perform things explicitly or run the code in some other funciton reverting to no callback under the hood. Does not affect the functionality. If you strongly feeling a meta function can help we can make that change.

tuning/config/configs.py
```diff
@@ -57,3 +57,7 @@ class TrainingArguments(transformers.TrainingArguments):
         default=False,
         metadata={"help": "Packing to be enabled in SFT Trainer, default is False"},
     )
+    tracker: str.lower = field(
+        default=None,
+        metadata={"help": "Experiment tracker to use. Requires additional configs, see tuning.configs/tracker_configs.py"}
```
nairbv, line 62: would be helpful to list the available options. from the code, looks like it would be "aim"

tuning/sft_trainer.py
```diff
@@ -286,23 +296,67 @@ def main(**kwargs):
         choices=["pt", "lora", None, "none"],
         default="pt",
     )
+    parser.add_argument(
+        "--exp_metadata",
+        type=str,
+        default=None,
```
nairbv, line 302: this probably needs help text. looks like it expects a json string?

tuning/sft_trainer.py
```diff
@@ -286,23 +296,67 @@ def main(**kwargs):
...
+        additional,
         _,
     ) = parser.parse_args_into_dataclasses(return_remaining_strings=True)
-    if peft_method.peft_method == "lora":
-        tune_config = lora_config
-    elif peft_method.peft_method == "pt":
-        tune_config = prompt_tuning_config
+
+    peft_method = additional.peft_method
+    if peft_method =="lora":
```
nairbv, line 316: you'll want to run black on your code, there should be a space after the `==`

requirements.txt
```diff
@@ -2,7 +2,7 @@ numpy
 accelerate>=0.20.3
 transformers>=4.34.1
 torch
-aim==3.17.5
+aim==3.18.1
```
nairbv, line 5: since there's a desire to implement multiple trackers, did we want to make the dependency (and imports) optional, just used when available and configured?

tuning/config/tracker_configs.py
```diff
@@ -0,0 +1,15 @@
+from dataclasses import dataclass
+
+@dataclass
+class AimConfig:
+    # Name of the experiment
+    experiment: str = None
+    # 'repo' can point to a locally accessible directory (e.g., '~/.aim') or a remote repository hosted on a server.
+    # When 'remote_server_ip' or 'remote_server_port' is set, it designates a remote aim repo.
+    # Otherwise, 'repo' specifies the directory, with a default of None representing '.aim'.
```
nairbv, line 9: if None implies `.aim` why not just make it default `.aim`?

tuning/config/tracker_configs.py
```diff
@@ -0,0 +1,15 @@
+from dataclasses import dataclass
+
+@dataclass
+class AimConfig:
+    # Name of the experiment
+    experiment: str = None
+    # 'repo' can point to a locally accessible directory (e.g., '~/.aim') or a remote repository hosted on a server.
+    # When 'remote_server_ip' or 'remote_server_port' is set, it designates a remote aim repo.
```
nairbv, line 8: when it's remote, does it mean a remote directory path?

tuning/config/tracker_configs.py
```diff
@@ -0,0 +1,15 @@
+from dataclasses import dataclass
+
+@dataclass
+class AimConfig:
+    # Name of the experiment
+    experiment: str = None
+    # 'repo' can point to a locally accessible directory (e.g., '~/.aim') or a remote repository hosted on a server.
```
nairbv, line 7: might be clearer to use the actual config name in the description (`aim_repo`)... though the `aim_` in the config names might be a bit redundant too.

tuning/sft_trainer.py
dushyantbehl: done

tuning/sft_trainer.py
dushyantbehl: thanks for the pointer. done

tuning/config/configs.py
dushyantbehl: done

requirements.txt
dushyantbehl: can be done..but then do we still list these in the requirements.txt?

or do we throw an error and ask user to install the required tracker before importing it in the code.

tuning/config/tracker_configs.py
dushyantbehl: well taken..this was a typo..fixed.

tuning/config/tracker_configs.py
dushyantbehl: No. Remote here means `http://server:port` I can link to the aim documentation here but felt anyone using this would already know how aimstack works.

tuning/config/tracker_configs.py
dushyantbehl: done

scripts/run_inference.py
```diff
@@ -8,6 +8,7 @@
 
 If these things change in the future, we should consider breaking it up.
 """
+
```
dushyantbehl, line 11: Generated by running `black`

dushyantbehl:
I have split this PR into two and opened them separately as https://github.com/foundation-model-stack/fms-hf-tuning/pull/89 and https://github.com/foundation-model-stack/fms-hf-tuning/pull/90 

Not closing this right away due to the comments.

dushyantbehl:
#89 and #90 is split version of this PR. We will track them for now.

PR #52, opened 2024-02-21
Add linting capability

hickeyma:
Added the following linter:
- [pylint](https://pylint.readthedocs.io/en/stable/)

**Note to reviewers:**

- Added [tox](https://tox.wiki/en/4.12.1/) target which can kick off linting.
- Separated packages that require CUDA into `cuda_requirements.txt` as they are not required for linting and also are not required for non CUDA tuning.
- The linting is not automated yet as need to add CI/CD
- Any changes recommended by the linter are not committed until we automate as part of the build
- The linting can be run manually:
   - Installing `tox` and dependencies: `pip install -r setup_requirements.txt`
   - Running it: `tox -e lint`


.pylintrc
```diff
@@ -0,0 +1,646 @@
...
+
+# Specify a score threshold under which the program will exit with error.
+fail-under=10
+
+# Interpret the stdin as a python script, whose filename needs to be passed as
+# the module_or_package argument.
+#from-stdin=
+
+# Files or directories to be skipped. They should be base names, not paths.
+ignore=CVS,protobufs
```
anhuong, line 49: This repo doesn't have CVS or protobfs

.pylintrc
```diff
@@ -0,0 +1,646 @@
...
+const-naming-style=UPPER_CASE
+
+# Regular expression matching correct constant names. Overrides const-naming-
+# style. If left empty, constant names will be checked with the set naming
+# style.
+#const-rgx=
+
+# Minimum line length for functions/classes that require docstrings, shorter
+# ones are exempt.
+docstring-min-length=-1
```
anhuong, line 171: This does seem like a good practice to instill but will fail immediately since we don't have docstrings for all classes and functions

anhuong (commented):
A few questions Martin
- Could you add how to run the linter and formatter in the README? Especially if this is something that is expected to be run by all contributors else PRs will fail (in near term future when automation added)
- If we separate torch_requirements, this should also be updated in the README since it will require users to run `pip install -r torch_requirements.txt` and `pip install -r requirements.txt`. Which will also affect the sft-trainer image that is being built. This I can update, but this will affect others users and unsure of any other automation being used for this installation, installing and running this repo in clusters as well. Could be nice to separate out the use case of GPU vs CPU

.pylintrc
hickeyma: Its a standard file from OOTO so no harm in leaving it in.

.pylintrc
hickeyma: Because it is set to `-1` it will not check for docstrings.

hickeyma:
@anhuong Thanks for the review and feedback.

> Could you add how to run the linter and formatter in the README? Especially if this is something that is expected to be run by all contributors else PRs will fail (in near term future when automation added)

I intend to do that in the contributors doc as part of the enablement of linting in CI/CD. Next PR essentially.

 > If we separate torch_requirements, this should also be updated in the README since it will require users to run pip install -r torch_requirements.txt and pip install -r requirements.txt. Which wil affect any current automation that is installing this repo.

Added to the README now.

hickeyma:
@anhuong Updated and ready for review again.

README.md
```diff
@@ -13,6 +13,8 @@ pip install -U datasets
 pip install -e .
 ```
 
+> Note: If you wish to use [FlashAttention](https://github.com/Dao-AILab/flash-attention), then you need to install these requirements: `pip install -r flashattn_requirements`. [FlashAttention](https://github.com/Dao-AILab/flash-attention) requires the [CUDA Toolit](https://developer.nvidia.com/cuda-toolkit) to be pre-installed.
```
fabianlim, line 16: I thought that `pip install flash-attn` does not require CUDA toolkit, because I have installed it in docker containers with no CUDA toolkit before. 

My experience is only that `pip install flash-atten --no-bulid-isolation` will run the compile process (and therefore require toolkit). The above command simply installs pre-compiled binaries so no toolkit should be required.

anhuong:
I tried uninstalling packages, wheel, and flash-attn and running tuning to make sure it can run without these deps. How did you come up with these packages as optional? I think flash-attn is okay to separate, but it looks like packages is needed. When trying to run PT, I get error:
```
  File "/dccstor/anhtest/fms-hf-tuning/tuning/sft_trainer.py", line 8, in <module>
    from peft.utils.other import fsdp_auto_wrap_policy
  File "/dccstor/anhtest/fms-hf-tuning/venv/lib/python3.9/site-packages/peft/__init__.py", line 22, in <module>
    from .auto import (
  File "/dccstor/anhtest/fms-hf-tuning/venv/lib/python3.9/site-packages/peft/auto.py", line 23, in <module>
    from transformers import (
  File "/dccstor/anhtest/fms-hf-tuning/venv/lib/python3.9/site-packages/transformers/__init__.py", line 26, in <module>
    from . import dependency_versions_check
  File "/dccstor/anhtest/fms-hf-tuning/venv/lib/python3.9/site-packages/transformers/dependency_versions_check.py", line 16, in <module>
    from .utils.versions import require_version, require_version_core
  File "/dccstor/anhtest/fms-hf-tuning/venv/lib/python3.9/site-packages/transformers/utils/__init__.py", line 19, in <module>
    from packaging import version
ModuleNotFoundError: No module named 'packaging'
```

anhuong:
Can verify that with set of optional deps uninstalled and with packaging installed, Lora tuning also runs successfully.

hickeyma:
As discussed @anhuong, I have moved `packaging` to the core requirements file.

anhuong (approved):
LGTM, thanks Martin!

tedhtchang:
We should enable fmt and lint checks for each PR https://github.com/foundation-model-stack/fms-hf-tuning/issues/59

PR #53, opened 2024-02-23
Let Huggingface Properly Initialize Arguments, and Fix FSDP-LORA Checkpoint-Saves and Resumption

fabianlim:
@raghukiran1224 @Ssukriti @anhuong Im suggesting two fixes here

1. We observe failures with newer version of transformers because of a newly added `xla_fsdp_v2` flag here. The current strategy is to manually patch missing flags because `transformers.TrainingArguments.__post_init__` is not called. But this means one has to constantly update the manual patches if there are changes to HF code, which is not ideal.  Huggingface trainer will properly initialize everything based on  `TrainingArguments` (including [gradient checkpointing](https://github.com/huggingface/transformers/blob/2cc8cf6ce7ae0416561acbb639df4bbc5f409b6f/src/transformers/trainer.py#L1749-L1755). 
2. The `PeftSavingCallback` is not the ideal patch now that https://github.com/huggingface/transformers/pull/28297 has been merged, where now FSDP will properly save and resume adapter checkpoints. This will supercede the `PeftSavingCallback` strategy, as `PeftSavingCallback` will not even properly handle resumptions and different state dict saving strategies, but the already merged fix will. the `PEFTCallback` is removed in `trl` [here](https://github.com/huggingface/trl/pull/1110)

### DIFF
--- a/tuning/config/configs.py
+++ b/tuning/config/configs.py
@@ -72,6 +72,3 @@ class TrainingArguments(transformers.TrainingArguments):
         default=False,
         metadata={"help": "Packing to be enabled in SFT Trainer, default is False"},
     )
-
-    def __post_init__(self):
-        super().__post_init__()
--- a/tuning/config/configs.py
+++ b/tuning/config/configs.py
@@ -72,3 +72,6 @@ class TrainingArguments(transformers.TrainingArguments):
         default=False,
         metadata={"help": "Packing to be enabled in SFT Trainer, default is False"},
     )
+
+    def __post_init__(self):
+        super().__post_init__()
--- a/tuning/sft_trainer.py
+++ b/tuning/sft_trainer.py
@@ -44,17 +44,6 @@ from tuning.utils.config_utils import get_hf_peft_config
 from tuning.utils.data_type_utils import get_torch_dtype
 
 
-class PeftSavingCallback(TrainerCallback):
-    def on_save(self, args, state, control, **kwargs):
-        checkpoint_path = os.path.join(
-            args.output_dir, f"checkpoint-{state.global_step}"
-        )
-        kwargs["model"].save_pretrained(checkpoint_path)
-
-        if "pytorch_model.bin" in os.listdir(checkpoint_path):
-            os.remove(os.path.join(checkpoint_path, "pytorch_model.bin"))
-
-
 class FileLoggingCallback(TrainerCallback):
     """Exports metrics, e.g., training loss to a file in the checkpoint directory."""
 
@@ -118,7 +107,6 @@ def train(
         None for fine tuning
             The peft configuration to pass to trainer
     """
-    run_distributed = int(os.environ.get("WORLD_SIZE", "1")) > 1
 
     logger = logging.get_logger("sft_trainer")
 
@@ -222,8 +210,7 @@ def train(
 
     aim_callback = get_aimstack_callback()
     file_logger_callback = FileLoggingCallback(logger)
-    peft_saving_callback = PeftSavingCallback()
-    callbacks = [aim_callback, peft_saving_callback, file_logger_callback]
+    callbacks = [aim_callback, file_logger_callback]
 
     if train_args.packing:
         logger.info("Packing is set to True")
@@ -264,7 +251,7 @@ def train(
         peft_config=peft_config,
     )
 
-    if run_distributed and peft_config is not None:
+    if trainer.is_fsdp_enabled and peft_config is not None:
         trainer.accelerator.state.fsdp_plugin.auto_wrap_policy = fsdp_auto_wrap_policy(
             model
         )
--- a/tuning/sft_trainer.py
+++ b/tuning/sft_trainer.py
@@ -122,21 +122,6 @@ def train(
 
     logger = logging.get_logger("sft_trainer")
 
-    # Validate parameters
-    if (not isinstance(train_args.num_train_epochs, float)) or (
-        train_args.num_train_epochs <= 0
-    ):
-        raise ValueError("num_train_epochs has to be an integer/float >= 1")
-    if (not isinstance(train_args.gradient_accumulation_steps, int)) or (
-        train_args.gradient_accumulation_steps <= 0
-    ):
-        raise ValueError("gradient_accumulation_steps has to be an integer >= 1")
-
-    # make sure to unset FSDP args when running on single gpu
-    if not run_distributed:
-        train_args.fsdp = ""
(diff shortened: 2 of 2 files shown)


fabianlim:
I noticed that any changes here is also releated to this file https://github.ibm.com/ai-foundation/sft-trainer-image/blob/main/launch_training.py.

tuning/sft_trainer.py
```diff
@@ -103,25 +92,9 @@ def train(
...
-    ):
-        raise ValueError("num_train_epochs has to be an integer/float >= 1")
-    if (not isinstance(train_args.gradient_accumulation_steps, int)) or (
-        train_args.gradient_accumulation_steps <= 0
-    ):
-        raise ValueError("gradient_accumulation_steps has to be an integer >= 1")
-
-    # make sure to unset FSDP args when running on single gpu
-    if not run_distributed:
-        train_args.fsdp = ""
```
Ssukriti, line 122: does single GPU still work after this change? without `train_args.fsdp_config = {"xla": False}` I remember single GPU training was failing

tuning/sft_trainer.py
```diff
@@ -263,7 +233,7 @@ def train(
         peft_config=peft_config,
     )
 
-    if run_distributed and peft_config is not None:
+    if trainer.is_fsdp_enabled and peft_config is not None:
```
Ssukriti, line 236: does this flag `trainer.is_fsdp_enabled` basically do the same thing where its enabled only in a multi GPU env that has the desired HW and disabled for single gPU env?

Ssukriti (changes requested):
1. left some questions and comments
2. DCO check is failing
3. Now that we have linted and formatted our files in main branch, the PR will need to be rebased on latest main . you need to update your fork to the latest main and then this branch with `git merge main`
sorry for the trouble, but any major changes upstream with linting and formatting is done now, so its a one-time pain :) 

Thank you for the contribution

tuning/sft_trainer.py
fabianlim: Yes I have verified that i runs in both single and multi gpu settings. 

if I can point your attention to [`TrainingArguments.__post_init__`](https://github.com/huggingface/transformers/blob/45c065109074d60c587d3e562f16531d02a422f6/src/transformers/training_args.py#L1691-L1707), you will see that it already has logic to assemble all these `self.fsdp_config` flags appropriately. 

As [it has been coming up](https://github.com/foundation-model-stack/fms-hf-tuning/pull/79#discussion_r1513617678) that we should keep `transformers < 4.38.0`, the reason is because at `4.38` they introduced a new `xla_fsdp_v2 ` flag, and it did not suffice to only set the `xla` flag. Hence, we should leave this to the `__post_init__`  

My guess is why it did not run before was that before this changes, the `__post_init__` of `TrainingArguments` is not called, and perhaps the flag `fsdp_config["xla"]` is

tuning/sft_trainer.py
```diff
@@ -103,25 +92,9 @@ def train(
         None for fine tuning
             The peft configuration to pass to trainer
     """
-    run_distributed = int(os.environ.get("WORLD_SIZE", "1")) > 1
 
     logger = logging.get_logger("sft_trainer")
 
-    # Validate parameters
```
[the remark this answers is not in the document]
fabianlim: hmm. i ran the `trainer` with `num_train_epochs=0`, with a `linear` schedule and it didnt throw error, and returned me the final output

```
{'train_runtime': 0.2356, 'train_samples_per_second': 0.0, 'train_steps_per_second': 0.0, 'train_tokens_per_second': 0.0, 'train_loss': 0.0, 'epoch': 0}
0it [00:00, ?it/s]
```

My thinking is such a basic check, if is necessary will already been done by the HF `Trainer`

tuning/sft_trainer.py
fabianlim: this is the API correct way to check if FSDP is enabled. it is [initialized like this](https://github.com/huggingface/transformers/blob/45c065109074d60c587d3e562f16531d02a422f6/src/transformers/trainer.py#L4101) from the `trainer.accelerator.state`, which is the official way HF handles FSDP configs.

tuning/sft_trainer.py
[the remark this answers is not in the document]
fabianlim: Update: i put back the checks.

fabianlim:
@Ssukriti I have rebased the changes and linted.

tuning/sft_trainer.py
Ssukriti: thank you!

Ssukriti (approved):
Thanks a lot @fabianlim for thoroughly testing your PR!

tuning/sft_trainer.py
Ssukriti: actually post_init is always called as it is a derived class. I think the error I was referring to was because in original first commit we were oevrriding it with the wrong value . we were doing `
        train_args.fsdp_config = {}` or something like that, so I changed it to `"xla": False` . But you are right that if we dont do anything the _post_init should take care of it. So makes sense to remove it, thanks

Ssukriti:
cant merge because of failing pylint, which seems to be a valid warning. Since this is your first commit Fabian, I have to manually start the workflow checks - pylint etc, thats why you didnt catch it earlier. Once this PR is merged and your fork is recognized as a contributor, going forward  the pylint checks with run automatically on your PRs and you wont have to wait for an admin to start checks :) 

to run pylint locally on your machine you can do `tox -e lint` as documented here 
https://github.com/foundation-model-stack/fms-hf-tuning/pull/84/files

Our contributing guides should be merged soon. 

I have approved the PR and will merge as soon as pylint checks pass

tuning/config/configs.py
```diff
@@ -72,3 +72,6 @@ class TrainingArguments(transformers.TrainingArguments):
         default=False,
         metadata={"help": "Packing to be enabled in SFT Trainer, default is False"},
     )
+
+    def __post_init__(self):
+        super().__post_init__()
```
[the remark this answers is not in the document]
fabianlim: ah ic. yes sorry it was my misunderstanding. I assumed the reason for putting in the `fsdp_config` hardcode was that `__post_init__` is not called automatically. But I have verified that it indeed is.

fabianlim:
@Ssukriti I have added one commit on top of your merge above #84, that removes the extranous `__post_init__`. I have also ran the pylint (which checks ok). Can you help to trigger the approving workflows?

Ssukriti (approved):
thank you!!!

PR #54, opened 2024-02-23
deps: limit dependency ranges

anhuong:
Limit dependency ranges to automatic minor upgrades, must manually check for major upgrades to see what breaking changes are included. Matches current versions that are being used


requirements.txt
```diff
@@ -1,15 +1,15 @@
...
-tqdm
-trl
-ninja
-peft>=0.8.0
-datasets>=2.15.0
-fire
+numpy>=1.26.4,<2.0
+accelerate>=0.20.3,<1.0
+transformers>=4.34.1,<5.0
+torch>=2.2.0,<3.0
```
tedhtchang, line 4: I ran into an error while using transformers 4.38.2.I think we need to pin specific versions. For example `transformers==4.37.2`
```
  File "/Users/tedchang/pub_git/fms-hf-tuning/tuning/sft_trainer.py", line 312, in <module>
    fire.Fire(main)
  File "/Users/tedchang/.pyenv/versions/fms-hf-tuning/lib/python3.11/site-packages/fire/core.py", line 141, in Fire
    component_trace = _Fire(component, args, parsed_flag_args, context, name)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/tedchang/.pyenv/versions/fms-hf-tuning/lib/python3.11/site-packages/fire/core.py", line 475, in _Fire
    component, remaining_args = _CallAndUpdateTrace(
                                ^^^^^^^^^^^^^^^^^^^^
  File "/Users/tedchang/.pyenv/versions/fms-hf-tuning/lib/python3.11/site-packages/fire/core.py", line 691, in _CallAndUpdateTrace
    component = fn(*varargs, **kwargs)
                ^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/tedchang/pub_git/fms-hf-tuning/tuning/sft_trainer.py", line 308, in main
    train(model_args, data_args, training_args, tune_config)
  File "/Users/tedchang/pub_git/fms-hf-tuning/tuning/sft_trainer.py", line 256, in train
    trainer = SFTTrainer(
              ^^^^^^^^^^^
  File "/Users/tedchang/.pyenv/versions/fms-hf-tuning/lib/python3.11/site-packages/trl/trainer/sft_trainer.py", line 295, in __init__
    super().__init__(
  File "/Users/tedchang/.pyenv/versions/fms-hf-tuning/lib/python3.11/site-packages/transformers/trainer.py", line 648, in __init__
    self.is_fsdp_xla_v2_enabled = args.fsdp_config["xla_fsdp_v2"]
                                  ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^
KeyError: 'xla_fsdp_v2'
```

anhuong: Indeed, minor changes will still cause breaking changes and we need to discuss how to handle dependency upgrades and testing. This is meant to have some kind of upper bounds so as to not include further bigger changes but does not solve the dependency breaks completely

anhuong:
relates/depends on #68

requirements.txt
gkumbhat: These types of bugs can arrive anytime and in any version. I think the goal here should be:
1. Open enough range for pip dep version resolution
2. Min version of dep that is required to run the code
3. Max version that we think should not contain any breaking change
4. platform and other environment specifications 


@anhuong for above case, for transformers range, we can provide an exclusion as well, like: 

`transformers>=4.34.1,<5.0,!=4.38.2`

the other option would be to restrict the version of transformers to a specific version like in [this PR](https://github.com/foundation-model-stack/fms-hf-tuning/pull/68/files#diff-4d7c51b1efe9043e44439a949dfd92e5827321b34082903477fd04876edb7552R4)

requirements.txt
```diff
@@ -1,15 +1,15 @@
...
+aim>=3.17.5,<4.0
+sentencepiece>=0.1.99,<2.0
+tokenizers>=0.13.3,<1.0
+tqdm>=4.66.2,<5.0
+trl>=0.7.10,<1.0
```
gkumbhat, line 9: for libraries that are not yet `1.0`, I think it is not guaranteed that there won't be any API breaking change between subversions, so it is probably better to be more conservative about their versions in range.

gkumbhat: applies less for `accelerate` though, since that library just never went `1.0` but is very popular, so unlikely to make such changes now without big splash.

pyproject.toml
```diff
@@ -26,24 +26,24 @@ classifiers=[
...
+"simpleeval>=0.9.13,<1.0",
 ]
 
 [project.optional-dependencies]
-dev = ["wheel", "packaging", "ninja", "scikit-learn>=1.0, <2.0"]
-flash-attn = ["flash-attn"]
-aim = ["aim==3.19.0"]
+dev = ["wheel>=0.42.0,<1.0", "packaging>=23.2,<24", "ninja>=1.11.1.1,<2.0", "scikit-learn>=1.0, <2.0"]
+flash-attn = ["flash-attn>=2.5.3,<3.0"]
+aim = ["aim>=3.19.0,<4.0"]
```
anhuong, line 46: Note that from [PR #93 comment](https://github.com/foundation-model-stack/fms-hf-tuning/pull/93#issuecomment-2100720341)
> [aim] was initially capped to specific version as newer major version of Aim contained some stability issues. Now I feel we can keep it directly to 3.X.X

pyproject.toml
```diff
@@ -26,24 +26,24 @@ classifiers=[
...
-"sentencepiece",
-"tokenizers>=0.13.3",
-"tqdm",
-"trl",
-"peft>=0.8.0",
-"datasets>=2.15.0",
-"fire",
-"simpleeval",
+"numpy>=1.26.4,<2.0",
+"accelerate>=0.20.3,<1.0",
```
tedhtchang, line 30: perhaps set more conservative 0.4 (current 0.31) since it's lower than 1.0

pyproject.toml
```diff
@@ -26,24 +26,24 @@ classifiers=[
...
+"transformers>=4.34.1,<5.0,!=4.38.2",
+"torch>=2.2.0,<3.0",
+"sentencepiece>=0.1.99,<1.0",
```
tedhtchang, line 33: perhaps set more conservative limit say 0.3  (current 0.2). This package updates very slow.

pyproject.toml
```diff
@@ -26,24 +26,24 @@ classifiers=[
...
+"tokenizers>=0.13.3,<1.0",
+"tqdm>=4.66.2,<5.0",
+"trl>=0.7.10,<1.0",
```
tedhtchang, line 36: perhaps 0.10.0 current is 0.9.4

pyproject.toml
```diff
@@ -26,24 +26,24 @@ classifiers=[
...
+"peft>=0.8.0,<1.0",
```
tedhtchang, line 37: perhaps set limit to 0.13 current 0.11.1

tedhtchang (commented):
@anhuong Could you rebase.
For the important packages, I think we could set more conservative limit for those lower than 1.0.
In new PR we can figure out how we freeze packages versions in case we need to create older commit.

pyproject.toml
```diff
@@ -26,24 +26,24 @@ classifiers=[
...
-"tqdm",
-"trl",
-"peft>=0.8.0",
-"datasets>=2.15.0",
-"fire",
-"simpleeval",
+"numpy>=1.26.4,<2.0",
+"accelerate>=0.20.3,<1.0",
+"transformers>=4.34.1,<5.0,!=4.38.2",
+"torch>=2.2.0,<3.0",
```
tedhtchang, line 32: May set to 2.4.0 current 2.3.1. I see API changes in minor release.

anhuong: since this one is above 1.0 and i think we would want to pull in new changes, unless something like dependabot is updating this i am more inclined to keep 3.0 as the upper limit

pyproject.toml
```diff
@@ -26,24 +26,24 @@ classifiers=[
...
 ]
 dependencies = [
-"numpy",
-"accelerate>=0.20.3",
-"transformers",
-"torch",
-"sentencepiece",
-"tokenizers>=0.13.3",
-"tqdm",
+"numpy>=1.26.4,<2.0",
```
anhuong, line 29: Note that numpy is now at 2.0 as of a few days ago. I have found that numpy 2.0 works still with tuning and the stack of deps. 

Although I did run into an issue running unit tests locally if torch==2.2.2 and numpy==2.0.0 then I got error `RuntimeError: Numpy is not available` from the unit tests. See [slack thread](https://ibm-research.slack.com/archives/C06MZPQ4ENL/p1718827390959019?thread_ts=1718820929.017379&cid=C06MZPQ4ENL). But as you can see in unittests it succeeds
- [ ran today](https://github.com/foundation-model-stack/fms-hf-tuning/actions/runs/9583556592/job/26425112934) passed with numpy 2.0.0, transformers 4.41.2, torch 2.3.1
- [older transformers](https://github.com/foundation-model-stack/fms-hf-tuning/actions/runs/9587647439/job/26438105396?pr=204) numpy 2.0.0, transformers 4.40.2, torch 2.3.1

anhuong: In addition, the pytorch community did test numpy 2.0.0rc1 from this [issue](https://github.com/pytorch/pytorch/issues/107302) and deemed it suitable and they have updated their deps to use numpy 2.0. But the upgrade should be evaluated further as it does contain breaking changes, see [release notes](https://numpy.org/devdocs/release/2.0.0-notes.html).

VassilisVassiliadis: The latest version of aim has a dependency to [`numpy<2,>=1.12.0`](https://github.com/aimhubio/aim/blob/fddfe1ed48908d5a5db60ca27ec675c039fcdf41/setup.py#L60). So I think it's best to keep `numpy>=1.26.4,<2.0` for now.

anhuong: good to know, thanks Vassilis!

tedhtchang (dismissed):
/LGTM

tedhtchang (approved):
/LGTM

PR #61, opened 2024-02-27
Add sample unit tests

### LINKED ISSUE #5: Add unit tests
**Description**

The repo is in need for unit tests so we can start enforcing tests to be contributed with every code contribution. In this issue we can add a simple unit test framework to the repository and some simple tests to get started . We also need to enable tests to run as part of every PR build. 

**Acceptance criteria**

1. Add unit tests framework to the repo - preferred pytorch
2. Tests organization in the repository - new folder called tests where we organize tests going forward
3. Add some simple unit tests as a good example (example given in comments)
4. Enable tests to run as part of every build

tedhtchang:
Changes:
1. Adds unit tests for utilts/data_type_utils module
2. Enables unit test framework in the workflow for every PR

Closes #5

Verify in a virtual env:
```bash
git clone https://github.com/tedhtchang/fms-hf-tuning -b add_sample_unit_test
cd fms-hf-tuning
pip install -r setup_requirement.txt
tox run -e py
```
Should see something like:
```
============================================================================================================================= test session starts ==============================================================================================================================
platform darwin -- Python 3.11.7, pytest-8.0.2, pluggy-1.4.0
cachedir: .tox/py/.pytest_cache
rootdir: /Users/tedchang/pub_git/fms-hf-tuning
collected 3 items

tests/test_data_type_utils.py ...                                                                                                                                                                                                                                        [100%]

============================================================================================================================== 3 passed in 0.97s ===============================================================================================================================
```

### DIFF
--- a/tests/utils/test_data_type_utils.py
+++ b/tests/utils/test_data_type_utils.py
@@ -29,5 +29,6 @@ def test_str_to_torch_dtype_exit():
 
 def test_get_torch_dtype():
     for t in dtype_dict.keys():
+        # When passed a string, it gets converted to torch.dtype
         assert data_type_utils.get_torch_dtype(t) == dtype_dict.get(t)
         assert data_type_utils.get_torch_dtype(dtype_dict.get(t)) == dtype_dict.get(t)
--- a/tests/utils/test_data_type_utils.py
+++ b/tests/utils/test_data_type_utils.py
@@ -1,0 +1,33 @@
+# SPDX-License-Identifier: Apache-2.0
+# https://spdx.dev/learn/handling-license-info/
+
+# Third Party
+import pytest
+import torch
+
+# Local
+from tuning.utils import data_type_utils
+
+dtype_dict = {
+    "bool": torch.bool,
+    "double": torch.double,
+    "float32": torch.float32,
+    "int64": torch.int64,
+    "long": torch.long,
+}
+
+
+def test_str_to_torch_dtype():
+    for t in dtype_dict.keys():
+        assert data_type_utils.str_to_torch_dtype(t) == dtype_dict.get(t)
+
+
+def test_str_to_torch_dtype_exit():
+    with pytest.raises(SystemExit):
+        data_type_utils.str_to_torch_dtype("foo")
+
+
+def test_get_torch_dtype():
+    for t in dtype_dict.keys():
+        assert data_type_utils.get_torch_dtype(t) == dtype_dict.get(t)
+        assert data_type_utils.get_torch_dtype(dtype_dict.get(t)) == dtype_dict.get(t)
--- a/tests/utils/test_data_type_utils.py
+++ b/tests/utils/test_data_type_utils.py
@@ -31,4 +31,5 @@ def test_get_torch_dtype():
     for t in dtype_dict.keys():
         # When passed a string, it gets converted to torch.dtype
         assert data_type_utils.get_torch_dtype(t) == dtype_dict.get(t)
+        # When passed a torch.dtype, we get the same torch.dtype returned
         assert data_type_utils.get_torch_dtype(dtype_dict.get(t)) == dtype_dict.get(t)
--- a/tests/utils/test_data_type_utils.py
+++ b/tests/utils/test_data_type_utils.py
@@ -23,7 +23,7 @@ def test_str_to_torch_dtype():
 
 
 def test_str_to_torch_dtype_exit():
-    with pytest.raises(SystemExit):
+    with pytest.raises(ValueError):
         data_type_utils.str_to_torch_dtype("foo")
 
 
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,13 @@
 [tox]
-envlist = lint, fmt
+envlist = py, lint, fmt
+
+[testenv]
+description = run unit tests
+deps =
+    pytest>=7
+    -r requirements.txt
+commands =
+    pytest {posargs:tests}
 
 [testenv:fmt]
 description = format with pre-commit
--- a/tuning/utils/data_type_utils.py
+++ b/tuning/utils/data_type_utils.py
@@ -22,7 +22,7 @@ def str_to_torch_dtype(dtype_str: str) -> torch.dtype:
     dt = getattr(torch, dtype_str, None)
     if not isinstance(dt, torch.dtype):
         logger.error(" ValueError: Unrecognized data type of a torch.Tensor")
-        exit(-1)
+        raise ValueError("Unrecognized data type of a torch.Tensor")
     return dt
 
 
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -1,0 +1,22 @@
+name: Test
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
+    branches: [ "main" ]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Python 3.9
+        uses: actions/setup-python@v4
+        with:
+          python-version: 3.9
+      - name: Install dependencies
+        run: |
+          python -m pip install -r setup_requirements.txt
+      - name: Run unit tests
+        run: tox -e py


Ssukriti:
the black formatter checks are failing

tedhtchang:
> the black formatter checks are failing

Thanks
Installed pre-commit hook locally to make sure formatting is run automatically for future PR.

jbusche:
I tried it on my laptop and on a Red Hat server... both looked good:

Mac Laptop:
```
====================================================================== test session starts ======================================================================
platform darwin -- Python 3.11.7, pytest-8.0.2, pluggy-1.4.0
cachedir: .tox/py/.pytest_cache
rootdir: /Users/jamesbusche/projects/RESEARCH/TED/fms-hf-tuning
collected 3 items                                                                                                                                               

tests/test_data_type_utils.py ...                                                                                                                         [100%]
======================================================================= 3 passed in 4.38s
```
Red Hat Server:
```
====================================================================== test session starts ======================================================================
platform linux -- Python 3.11.5, pytest-8.0.2, pluggy-1.4.0
cachedir: .tox/py/.pytest_cache
rootdir: /root/JIM/fms-hf-tuning
collected 3 items                                                                                                                                               

tests/test_data_type_utils.py ...                                                                                                                         [100%]

======================================================================= 3 passed in 1.42s
```

Ssukriti:
@joe.olson@ibm.com

tests/utils/test_data_type_utils.py
```diff
@@ -0,0 +1,64 @@
...
+    "quint4x2": torch.quint4x2,
+    "quint8": torch.quint8,
+    "short": torch.short,
+    "uint8": torch.uint8,
+}
+
+
+def test_str_to_torch_dtype():
+    for t in dtype_dict.keys():
+        assert data_type_utils.str_to_torch_dtype(t) == dtype_dict.get(t)
```
fabianlim, line 53: im not sure if this is a good function to test. THis is a function that deals with configs, but for such things, its better to defer to huggingface, because changes come in very fast

As an example, this function will fail if `auto` is passed in, because it is not an "actual" torch dtype

however, it is a valid flag, as seen in this code https://github.com/huggingface/transformers/blob/main/src/transformers/modeling_utils.py#L3317-L3340

tedhtchang: >>> from tuning.utils import data_type_utils
>>>data_type_utils.get_torch_dtype("auto")
 ValueError: Unrecognized data type of a torch.Tensor

fabianlim: right but what is the motivation of checking the formatting here, when it is already checked more throughly in the huggingface trainer? If there is an error, the huggingface trainer with throw the Exception which will be bubbled up the script. So what difference in behavior is this check trying to achieve, if in both cases, the idea is to return an error code.

fabianlim: ok i see, I think i misunderstood https://github.com/huggingface/transformers/blob/0ad770c3733f9478a8d9d0bc18cc6143877b47a2/src/transformers/modeling_utils.py#L2707-L2727, 

the only `str` value it will take is `auto`, the rest relies on it to be a torch dtype. So i think its a valid function. 

If think omitting `auto` is a design decision. either way is fine.,

fabianlim: Sorry an update to my previous comment
> the only str value it will take is auto, the rest relies on it to be a torch dtype. So i think its a valid function.

It turns out that `trl` has a `ModelConfig` dataclass which handles `torch_dtype`. I think its better to rely on that

https://github.com/huggingface/trl/blob/3bd02380c7e39cae2ec5013f27907b30bc4ce6bf/trl/trainer/model_config.py#L21

tedhtchang: @fabianlim Are you suggesting 
1. These functions need to to handle `auto`? If so create an issue.
2. Or we need to add more test case to test these functions? If so, what additional input to the functions do I need to test and what is the expected output of the functions do I need to assert.

Ssukriti: @tedhtchang as far as these unit go, they look great , thank you! we do not need to test each and every torch dtype , if conversion works for some , it will work for others. So i will add a separate comment on reducing the number of tests - remember every parameter value you pass is like another test - CI/CD tests also need to run fast. So we need CODE coverage and pass parameters that would go through a separate codepath while testing - example if "floats" were handled differently from "ints" then it makes sense to have 2 tests. Here all string to torch dtype is happening with 1 codepath - so we need not have 20+ tests for one codepath

tests/utils/test_data_type_utils.py
```diff
@@ -0,0 +1,64 @@
...
+# https://spdx.dev/learn/handling-license-info/
+
+# Third Party
+import pytest
+import torch
+
+# Local
+from tuning.utils import data_type_utils
+
+dtype_dict = {
```
Ssukriti, line 11: **please reduce this list to 5 unique types.**
 Purpose of unit test is to test OUR code path, all the different strings go through same codepath and are internally handled by gettattr() function in python to get the torch.dtype. We are not testing other Python provided utilities to see if that works. we are just testing our wrappers - do they pass code correctly and return correct types

tox.ini
```diff
@@ -1,5 +1,14 @@
 [tox]
-envlist = lint, fmt
+envlist = py, lint, fmt
+
+[testenv]
+description = run unit tests
+deps =
+    pytest>=7
+    torch
```
Ssukriti, line 8: why are torch and transformers here ? these deps are already installed as part of project installation. test environment should install anything else missing like pytest

Ssukriti: torch and transformers should come from here https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/requirements.txt

tedhtchang: It seems the tox always creates an isolated virtual env to run the test. I could remove them but ran into error such as
`E   ModuleNotFoundError: No module named 'torch'`

Ssukriti: for projects that do not have setup.py the recommendation is to install from requirements https://web.archive.org/web/20150112223937/https://blog.schwuk.com/2014/03/19/using-tox-django-projects/ 

but pls chekc with @hickeyma on best practice around this, we do not want to manage same list of dependencies multiple times by copy pasting them into files

Ssukriti: sorry I missed your comment above, ya the snippet you posted with `-r requirements.txt` looks better @tedhtchang ! Thanks, we just dont want to copy out deps from requirements

tedhtchang: @Ssukriti I think the requirements.txt will also be gone after pyproject.toml get merged.

tests/utils/test_data_type_utils.py
```diff
@@ -0,0 +1,33 @@
...
+}
+
+
+def test_str_to_torch_dtype():
+    for t in dtype_dict.keys():
+        assert data_type_utils.str_to_torch_dtype(t) == dtype_dict.get(t)
+
+
+def test_str_to_torch_dtype_exit():
+    with pytest.raises(SystemExit):
```
Ssukriti, line 26: this should test for vlaueError and not exit , if we pass a string with wrong dtype , it gives ValueError

tests/utils/test_data_type_utils.py
```diff
@@ -0,0 +1,33 @@
...
+        data_type_utils.str_to_torch_dtype("foo")
+
+
+def test_get_torch_dtype():
+    for t in dtype_dict.keys():
+        assert data_type_utils.get_torch_dtype(t) == dtype_dict.get(t)
```
Ssukriti, line 32:
```suggestion
        # When passed a string, it gets converted to torch.dtype
        assert data_type_utils.get_torch_dtype(t) == dtype_dict.get(t)
```

tox.ini
tedhtchang: We can ask @jbusche to keep the requirements.txt for now so we can get this merged. After he merged his PR, we can think about how to remove the requirements.txt. I am thinking something like:
```
skip_install = true
whitelist_externals = poetry
commands =
    poetry install
    poetry run pytest
```

tests/utils/test_data_type_utils.py
```diff
@@ -0,0 +1,33 @@
...
+        assert data_type_utils.get_torch_dtype(dtype_dict.get(t)) == dtype_dict.get(t)
```
Ssukriti, line 33:
```suggestion
        # When passed a torch.dtype, we get the same torch.dtype returned
        assert data_type_utils.get_torch_dtype(dtype_dict.get(t)) == dtype_dict.get(t)
```

Ssukriti (changes requested):
some small minor changes, comments and test for ValueError

when these changes are addressed, PR will be merged

tests/utils/test_data_type_utils.py
tedhtchang: @Ssukriti The function should `raise ValueError("Unrecognized data type of a torch.Tensor")` instead of [exit(-1)](https://github.com/tedhtchang/fms-hf-tuning/blob/863cc53b8508bdf4f53aed98c788d4f8585c75f1/tuning/utils/data_type_utils.py#L25)

Ssukriti: Yikes I did not realize the code was just logging valueerror `logger.error(" ValueError: Unrecognized data type of a torch.Tensor")` , but exiting without throwing error. Thank you, I will have someone change it

tuning/utils/data_type_utils.py
```diff
@@ -22,7 +22,7 @@ def str_to_torch_dtype(dtype_str: str) -> torch.dtype:
     dt = getattr(torch, dtype_str, None)
     if not isinstance(dt, torch.dtype):
         logger.error(" ValueError: Unrecognized data type of a torch.Tensor")
-        exit(-1)
+        raise ValueError("Unrecognized data type of a torch.Tensor")
```
Ssukriti, line 25: thank you!

Ssukriti (approved):
Excellent work, thank you!

tedhtchang:
> Excellent work, thank you!

Thanks for the review.

PR #63, opened 2024-02-27
Enable pylint in the github workflow

### LINKED ISSUE #59: Enable lint for PR
Enable lint and checks for every PR

tedhtchang:
Changes
1. Add pylint to workflow
2. Fix existing pylint errors

Closes #59
Test locally:
git clone https://github.com/tedhtchang/fms-hf-tuning -b Enable-pylint
```bash
cd fms-hf-tuning
pip install -r setup_requirement.txt
tox run -e lint
```

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -443,7 +443,9 @@ disable=raw-checker-failed,
         attribute-defined-outside-init,
         abstract-method,
         pointless-statement,
-        wrong-import-order
+        wrong-import-order,
+        duplicate-code,
+        unbalanced-tuple-unpacking
 
 # Enable the message, report, category or checker with the given id(s). You can
 # either give multiple identifier separated by comma (,) or put this option
--- a/tuning/sft_trainer.py
+++ b/tuning/sft_trainer.py
@@ -3,6 +3,7 @@ from datetime import datetime
 from typing import Optional, Union
 import json
 import os
+import sys
 
 # Third Party
 from peft.utils.other import fsdp_auto_wrap_policy
@@ -80,7 +81,7 @@ class FileLoggingCallback(TrainerCallback):
             return
 
         # append the current log to the jsonl file
-        with open(log_file, "a") as f:
+        with open(log_file, "a", encoding="utf-8") as f:
             f.write(f"{json.dumps(log_obj, sort_keys=True)}\n")
 
 
@@ -88,7 +89,7 @@ def train(
     model_args: configs.ModelArguments,
     data_args: configs.DataArguments,
     train_args: configs.TrainingArguments,
-    peft_config: Optional[
+    peft_configs: Optional[
         Union[peft_config.LoraConfig, peft_config.PromptTuningConfig]
     ] = None,
 ):
@@ -98,7 +99,7 @@ def train(
         model_args: tuning.config.configs.ModelArguments
         data_args: tuning.config.configs.DataArguments
         train_args: tuning.config.configs.TrainingArguments
-        peft_config: peft_config.LoraConfig for Lora tuning | \
+        peft_configs: peft_config.LoraConfig for Lora tuning | \
         peft_config.PromptTuningConfig for prompt tuning | \
         None for fine tuning
             The peft configuration to pass to trainer
@@ -130,7 +131,7 @@ def train(
         use_flash_attention_2=model_args.use_flash_attn,
     )
 
-    peft_config = get_hf_peft_config(task_type, peft_config)
+    peft_configs = get_hf_peft_config(task_type, peft_configs)
 
     model.gradient_checkpointing_enable()
 
@@ -140,9 +141,7 @@ def train(
     )
 
     # TODO: understand if we need to hardcode these here or just use defaults in model
-    if isinstance(tokenizer, LlamaTokenizer) or isinstance(
-        tokenizer, LlamaTokenizerFast
-    ):
+    if isinstance(tokenizer, (LlamaTokenizer, LlamaTokenizerFast)):
         tokenizer.add_special_tokens(
             {
                 "bos_token": "<s>",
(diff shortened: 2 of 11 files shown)


tedhtchang:
@hickeyma What's your recommendation for the following warnings/convention? They seem reasonable to me. Do we have to correct every one?  
```
....
************* Module tuning.sft_trainer
tuning/sft_trainer.py:84:13: W1514: Using open without explicitly specifying an encoding (unspecified-encoding)
tuning/sft_trainer.py:171:4: W1203: Use lazy % or .format() formatting in logging functions (logging-fstring-interpolation)
tuning/sft_trainer.py:173:8: W1203: Use lazy % or .format() formatting in logging functions (logging-fstring-interpolation)
tuning/sft_trainer.py:208:21: C3001: Lambda expression assigned to a variable. Define a function using the "def" keyword instead. (unnecessary-lambda-assignment)
tuning/sft_trainer.py:215:4: W1203: Use lazy % or .format() formatting in logging functions (logging-fstring-interpolation)
tuning/sft_trainer.py:220:8: W1203: Use lazy % or .format() formatting in logging functions (logging-fstring-interpolation)
************* Module tuning.utils.merge_model_utils
tuning/utils/merge_model_utils.py:84:9: W1514: Using open without explicitly specifying an encoding (unspecified-encoding)
************* Module scripts.run_inference
scripts/run_inference.py:80:13: W1514: Using open without explicitly specifying an encoding (unspecified-encoding)
scripts/run_inference.py:84:13: W1514: Using open without explicitly specifying an encoding (unspecified-encoding)
scripts/run_inference.py:247:13: W1514: Using open without explicitly specifying an encoding (unspecified-encoding)
scripts/run_inference.py:260:9: W1514: Using open without explicitly specifying an encoding (unspecified-encoding)

....
Messages
--------
+------------------------------+------------+
|message id                    |occurrences |
+==============================+============+
|unspecified-encoding          |6           |
+------------------------------+------------+
|logging-fstring-interpolation |4           |
+------------------------------+------------+
|unnecessary-lambda-assignment |1           |
+------------------------------+------------+
-----------------------------------
Your code has been rated at 9.69/10
lint: exit 20 (15.69 seconds) /home/runner/work/fms-hf-tuning/fms-hf-tuning> pylint tuning 'scripts/*.py' pid=2242
.pkg: _exit> python /opt/hostedtoolcache/Python/3.9.18/x64/lib/python3.9/site-packages/pyproject_api/_backend.py True setuptools.build_meta __legacy__
  lint: FAIL code 20 (139.39=setup[123.70]+cmd[15.69] seconds)
  evaluation failed :( (139.45 seconds)
Error: Process completed with exit code 20.
```

jbusche:
Nice @tedhtchang, just tested and I'm seeing the same lint results as you.  They're warnings and convention errors, so I would imagine they're not critical at the moment but I suspect 3 issues could be created for each type and they could get addressed later as time permits...
[W1514](https://pylint.readthedocs.io/en/stable/user_guide/messages/warning/unspecified-encoding.html)
[W1203](https://pylint.readthedocs.io/en/stable/user_guide/messages/warning/logging-fstring-interpolation.html)
[C3001](https://pylint.readthedocs.io/en/stable/user_guide/messages/convention/unnecessary-lambda-assignment.html)

Ssukriti:
@tedhtchang one change I would suggest here is if we can update CI/CD checks to fail only if pylint score < 9.0 ? 

As you can see the check is failing on your PR . If checks fail, we will block the PR . Expecting 10 /10 may be unreasonable at the moment

jbusche:
That worked well @tedhtchang 
```
Your code has been rated at 9.69/10

.pkg: _exit> python /root/.env/lib64/python3.11/site-packages/pyproject_api/_backend.py True setuptools.build_meta __legacy__
  lint: OK (837.33=setup[810.68]+cmd[26.64] seconds)
  congratulations :) (837.42 seconds)
```

hickeyma:
> one change I would suggest here is if we can update CI/CD checks to fail only if pylint score < 9.0 ?

This is not good practise. I recommend checking out the pylint code first (search: `pylint W1203` for example) and if the change recommended doesn't suit for a good technical reason then disable using the comment: `# pylint: disable=`

hickeyma:
> They seem reasonable to me. Do we have to correct every one?

@tedhtchang Maybe you can't resolve everyone of the warnings/errors but need to decide for a good reason and disable as shown in previous comment above.

tedhtchang:
The `black` formatter doesn't like those`# pylint: disable=logging-fstring-interpolation` or `# pylint: disable=unspecified-encoding` comments because they make lines longer than 100 columns :(

I ended up fixed almost every warning except :
```    
        format_dataset = lambda example: {  # pylint: disable=unnecessary-lambda-assignment
        f"{data_args.dataset_text_field}": example[f"{data_args.dataset_text_field}"]
        + tokenizer.eos_token
    }
``` 
and 
```
- https://huggingface.co/docs/peft/main/en/package_reference/lora#peft.LoraModel.add_weighted_adapter # pylint: disable=line-too-long
```

@Ssukriti We should to manually run functional test that can exercise these modules:
sft_trainer.py, run_inference.py, merge_model_utils.py

tedhtchang:
I got an `ERROR: Could not consume arg: --model_name_or_path
Usage: sft_trainer.py` from  this command. May I borrow some eyes to debug.
```
export CUDA_VISIBLE_DEVICES=0
export WORLD_SIZE=0
export MODEL_PATH="bigscience/bloomz-560m"
export DATA_PATH="twitter_complaints.json"
export OUTPUT_PATH="/tmp/fms-hf-tuning/output"
python -m tuning/sft_trainer  \
--model_name_or_path $MODEL_PATH  \
--data_path $DATA_PATH  \
--output_dir $OUTPUT_PATH  \
--num_train_epochs 1  \
--per_device_train_batch_size 4  \
--per_device_eval_batch_size 4  \
--gradient_accumulation_steps 4  \
--evaluation_strategy "no"  \
--save_strategy "epoch"  \
--learning_rate 1e-5  \
--weight_decay 0.0  \
--lr_scheduler_type "cosine"  \
--logging_steps 1  \
--packing false  \
--include_tokens_per_second true \
--response_template "\n### Response:"  \
--tokenizer_name_or_path "bigscience/tokenizer" \
--use_flash_attn false \
--torch_dtype "float32" \
--peft_method "pt"
```

tuning/sft_trainer.py
```diff
@@ -80,15 +81,15 @@ def _track_loss(self, loss_key, log_file, logs, state):
...
+        with open(log_file, "a", encoding="utf-8") as f:
             f.write(f"{json.dumps(log_obj, sort_keys=True)}\n")
 
 
 def train(
     model_args: configs.ModelArguments,
     data_args: configs.DataArguments,
     train_args: configs.TrainingArguments,
-    peft_config: Optional[
+    peft_configs: Optional[
```
anhuong, line 92: Why did you update the name to `peft_configs`? Was this a requirement from pylinter? Only a single config should be passed in for training, either the LoraConfig or the PromptTuningConfig so it makes sense to leave it as singular

tedhtchang: @anhuong I was getting this lint error: `tuning/sft_trainer.py:91:4: W0621: Redefining name 'peft_config' from outer scope (line 26) (redefined-outer-name)`
I could revert this and mark it `# pylint: disable=redefined-outer-name`. What do you think?

anhuong: Good to know about the error, okay to leave it then

Ssukriti:
@tedhtchang @hickeyma we have seen those conflicts before pylint and black even before and we have not been able to mandate a 10 pylint score. I do not want to slow down contributions getting stuck in these kind of conflicts and have to figure out how to disable pylint.
There are lot of instances online that mention problems with pylint https://pythonspeed.com/articles/pylint/ (see how to use pylint section). 
1. since we want to use pylint only to catch production errors, we can either enable only few checks as suggested in the article
2. Another way to do it is, since Ted went over the changes and fixed everything he could - we cap CI/CD check to fail only if score below `whatever score it is at now`, to catch only new errors going forward. If score needs to be further reduced in future, we will

we do not want that CI/CD to always fail, otherwise no use having that CI/CD check

Ssukriti:
for functional testing if you ran command in your comment https://github.com/foundation-model-stack/fms-hf-tuning/pull/63#issuecomment-1973769329 , that should be sufficient.
we will do further testing on our side, @anhuong will run one test to make sure tuning still works as part of review

tedhtchang:
> @tedhtchang the above command worked fine for me. After checking out your PR branch and installing it:
> ...
> Not sure why it didn't work for you. Also worked with llama-7b model

@anhuong I think it failed to run with transformers==4.38.2 package installed. I had to downgrade to 4.37.2

tedhtchang:
@hickeyma 's Recommendation is we should fix, disable(in-line), disable globally in that order.

anhuong (commented):
@tedhtchang Do you plan to add details on how to run and address errors from the linter in a later PR after Contributions.md has been added?

.pylintrc
```diff
@@ -443,7 +443,9 @@ disable=raw-checker-failed,
         attribute-defined-outside-init,
         abstract-method,
         pointless-statement,
-        wrong-import-order
+        wrong-import-order,
+        duplicate-code,
```
anhuong, line 447: Was there duplicate code in the launch_training that wasn't worth fixing? Seems like a good check to have instead of globally removing

tedhtchang: @anhuong What do you think?  I can revert this change and fix launch_training.py in another PR properly.
It looked a little messy because I had to put many` #pylint: disable=duplicate-code` in those 2 files. I could fix but then I have to figure out how to test it which could further delay this PR. 
```
build/launch_training.py:1:0: R0801: Similar lines in 2 files
==build.launch_training:[65:74]
==tuning.sft_trainer:[277:286]
    parser = transformers.HfArgumentParser(
        dataclass_types=(
            configs.ModelArguments,
            configs.DataArguments,
            configs.TrainingArguments,
            peft_config.LoraConfig,
            peft_config.PromptTuningConfig,
        )
    ) (duplicate-code)
build/launch_training.py:1:0: R0801: Similar lines in 2 files
==build.launch_training:[80:86]
==tuning.sft_trainer:[292:298]
        (
            model_args,
            data_args,
            training_args,
            lora_config,
            prompt_tuning_config, (duplicate-code)
```

anhuong: That is true I duplicated the code because passing the arguments to sft_trainer was causing issues and we wanted the same parsing of arguments. That's fine then, thanks for showing me what the output looks like

anhuong (approved):
Thanks Ted for adding the linting, explaining the changes, and for creating issue #78 for documenting the linting process.

PR #69, opened 2024-03-04
Initial commit for trainer image

tharapalanivel:
### Description of the change

Open source sft-trainer Dockerfile and launch_training script 

### Was the PR tested

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -86,9 +86,12 @@ RUN python -m pip install packaging && \
     python -m pip install --upgrade pip && \
     python -m pip install torch && \
     python -m pip install wheel
+
+# TODO Move to installing wheel once we have proper releases setup instead of cloning the repo
 RUN git clone https://github.com/foundation-model-stack/fms-hf-tuning.git && \
     cd fms-hf-tuning && \
     python -m pip install -r requirements.txt && \
+    python -m pip install -r flashattn_requirements.txt && \
     python -m pip install -U datasets  && \
     python -m pip install /tmp/fms-hf-tuning
 


trainer_image/Dockerfile
```diff
@@ -0,0 +1,115 @@
...
+WORKDIR /tmp
+RUN python -m pip install packaging && \
+    python -m pip install --upgrade pip && \
+    python -m pip install torch && \
+    python -m pip install wheel
+RUN git clone https://github.com/foundation-model-stack/fms-hf-tuning.git && \
+    cd fms-hf-tuning && \
+    python -m pip install -r requirements.txt && \
+    python -m pip install -U datasets  && \
+    python -m pip install /tmp/fms-hf-tuning
```
anhuong, line 93: Instead of cloning the repo its in, it should COPY in the files needed and install directly. It shouldn't need the `tuning` dir since that is pip installed although having this is useful for testing. But including the `scripts/run_inference` would be useful

anhuong:
Also the format linter failed, please try running `tox -e fmt` to fix errors found in launch_training script.

trainer_image/Dockerfile
anhuong: That's fine with me, will you add a TODO note in the Dockerfile

build/Dockerfile
```diff
@@ -0,0 +1,115 @@
...
+RUN mkdir /app
+
+WORKDIR /tmp
+RUN python -m pip install packaging && \
+    python -m pip install --upgrade pip && \
+    python -m pip install torch && \
+    python -m pip install wheel
+RUN git clone https://github.com/foundation-model-stack/fms-hf-tuning.git && \
+    cd fms-hf-tuning && \
+    python -m pip install -r requirements.txt && \
```
anhuong, line 91: Please add 1. TODO to change this to pip install and 2. for now, the correct way of installing would also include a line after this with `pip install -r flashattn_requirements.txt`

anhuong (changes requested):

anhuong (approved):
Thanks Thara!

PR #79, opened 2024-03-05
Adding unit tests for each tuning type

### LINKED ISSUE #74: Add unit tests for tuning/sft_trainer.py

Add unit tests for the tuning/sft_trainer.py module.

If possible, each function should have test cases in the file tests/test_sft_trainer.py
and pass test with the command `pytest tests/test_sft_trainer.py`

## Additional context

Add any other context about the feature request here.

tharapalanivel:
### Description of the change

Adding unit tests for `pt` and `lora` tuning method using dummy model
### Related issue number

Closes #74 

### Was the PR tested

### DIFF
--- a/tests/test_sft_trainer.py
+++ b/tests/test_sft_trainer.py
@@ -16,18 +16,18 @@
 """
 
 # Standard
+import json
 import os
 import tempfile
-import json
 
 # First Party
+from scripts.run_inference import TunedCausalLM
 from tests.data import TWITTER_COMPLAINTS_DATA
 from tests.fixtures import CAUSAL_LM_MODEL
 from tests.helpers import causal_lm_train_kwargs
 
 # Local
 from tuning import sft_trainer
-from scripts.run_inference import TunedCausalLM
 
 HAPPY_PATH_KWARGS = {
     "model_name_or_path": CAUSAL_LM_MODEL,
@@ -74,7 +74,9 @@ def test_run_causallm_pt():
         )
 
         # Run inference on the text using the tuned model
-        loaded_model.run("Simply put, the theory of relativity states that ", max_new_tokens=500)
+        loaded_model.run(
+            "Simply put, the theory of relativity states that ", max_new_tokens=500
+        )
 
 
 def test_run_causallm_lora():
@@ -95,7 +97,9 @@ def test_run_causallm_lora():
         )
 
         # Run inference on the text using the tuned model
-        loaded_model.run("Simply put, the theory of relativity states that ", max_new_tokens=500)
+        loaded_model.run(
+            "Simply put, the theory of relativity states that ", max_new_tokens=500
+        )
 
 
 def _validate_training(tempdir, peft_type):


requirements.txt
```diff
@@ -1,7 +1,7 @@
 numpy
 accelerate>=0.20.3
 packaging
-transformers>=4.34.1
+transformers>=4.34.1,<4.38.0
```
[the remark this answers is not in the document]
fabianlim: the reason you are seeing this error is because `TrainingArguments` is not initialized properly. In the code, we inherit from the dataclass, but we did not call `.__post_init__.` This PR #53 fixes this issue.

fabianlim:
Not sure if its a good idea to commit binary files that are 17MB large into a git repo.

tests/test_sft_trainer.py
```diff
@@ -0,0 +1,86 @@
...
+    "per_device_eval_batch_size": 4,
+    "gradient_accumulation_steps": 4,
+    "learning_rate": 0.00001,
+    "weight_decay": 0,
+    "warmup_ratio": 0.03,
+    "lr_scheduler_type": "cosine",
+    "logging_steps": 1,
+    "include_tokens_per_second": True,
+    "packing": False,
+    "response_template": "\\n###Response:",
```
anhuong, line 44: This should be `### Label` instead of Response and note the space. This matches the twitter dataset input

tests/test_sft_trainer.py
```diff
@@ -0,0 +1,86 @@
...
+        model_args, data_args, training_args, tune_config = causal_lm_train_kwargs(
+            HAPPY_PATH_KWARGS
+        )
+        sft_trainer.train(model_args, data_args, training_args, tune_config)
+        _validate_training(tempdir)
+
+
+def _validate_training(tempdir):
+    assert any(x.startswith("checkpoint-") for x in os.listdir(tempdir))
+    loss_file_path = "{}/train_loss.jsonl".format(tempdir)
```
anhuong, line 84: perhaps update to `train_loss_file_path` to distinguish between eval loss and train loss

tests/helpers.py
```diff
@@ -0,0 +1,56 @@
...
+# limitations under the License.
+
+# Third Party
+import transformers
+
+# Local
+from tuning.config import configs, peft_config
+
+
+def causal_lm_train_kwargs(train_kwargs):
```
anhuong, line 22: This has come up a few times now, that this code is duplicated in launch_training script and now is duplicated in tests. Feels like it may be worthwhile to refactor the argument parsing into its own method that can be tested separated in unit tests and called in both places.

tests/helpers.py
```diff
@@ -0,0 +1,56 @@
...
+        prompt_tuning_config,
+    ) = parser.parse_dict(train_kwargs, allow_extra_keys=True)
+
+    # TODO: target_modules doesn't get set probably due to the way dataclass handles
+    # mutable defaults, needs investigation on better way to handle this
+    setattr(
+        lora_config,
+        "target_modules",
+        lora_config.__dataclass_fields__.get("target_modules").default_factory(),
+    )
```
anhuong, line 47: Can you elaborate on this error?

PR #82, opened 2024-03-07
Allow SFT_TRAINER_CONFIG_JSON_ENV_VAR to be encoded json string

kellyaa:
Allow SFT_TRAINER_CONFIG_JSON_ENV_VAR to be encoded json string, not just pickled python

### Description of the change

Users of SFT Trainer may be using other programming languages than python to programmatically deploy the job, therefore SFT_TRAINER_CONFIG_JSON_ENV_VAR should not just accept a pickled python object, but also accept just a plain encoded json string.


kellyaa:
@tedhtchang Done! Thank you

Ssukriti (approved):
Thanks for the review @tedhtchang ! merging

PR #83, opened 2024-03-07
Unit tests

### LINKED ISSUE #74: Add unit tests for tuning/sft_trainer.py

tharapalanivel:
### Description of the change

Adding unit tests for `pt` and `lora` tuning method using dummy model, edge cases, invalid requests, etc.

Cont. of PR https://github.com/foundation-model-stack/fms-hf-tuning/pull/79
### Related issue number

Closes #74 

### Was the PR tested


Ssukriti:
This PR looks good to me. Before we add edge cases, can we also add a unit test for fine tuning? there will not be any peft type associated with it

requirements.txt
```diff
@@ -1,7 +1,7 @@
 numpy
 accelerate>=0.20.3
 packaging
-transformers>=4.34.1
+transformers>=4.34.1,<4.38.0
```
Ssukriti, line 4: #53 is merged so you shouldnt need this cap, thanks

tedhtchang: @Ssukriti Should we keep the cap for the transformers package avoid un intended errors like `xla_fsdp_v2`. We could create github workflow to run tests and then update the cap regularly.

Ssukriti: we dont need to keep static version , but yes in optional dependencies PR , @gkumbhat is looking into how to cap and we may cap to next major release. Now that CI/CD with automatically pull new release versions , if we see failing builds, we will update accordingly
the errors we were seeing with `xla_fsdp_v2` was actually due to code we wrote , which was good to catch and fix . It was not a API change from transformers, but we were setting env variables incorrectly

tests/test_sft_trainer.py
```diff
@@ -0,0 +1,291 @@
...
+
+        # Run inference on the text
+        output_inference = loaded_model.run(
+            "### Text: @NortonSupport Thanks much.\n\n### Label:", max_new_tokens=50
+        )
+        assert len(output_inference) > 0
+        assert "### Text: @NortonSupport Thanks much.\n\n### Label:" in output_inference
+
+
+def _validate_training(tempdir):
```
Ssukriti, line 269: hmm why not pass a flag here for validation dataset as bool. if set to True, also verify validation_loss.jsonl exists

tests/test_sft_trainer.py
```diff
@@ -0,0 +1,291 @@
...
+        model_args, data_args, training_args, tune_config = causal_lm_train_kwargs(
+            validation_peft
+        )
+
+        assert data_args.validation_data_path == TWITTER_COMPLAINTS_DATA
+
+        sft_trainer.train(model_args, data_args, training_args, tune_config)
+        _validate_training(tempdir)
+
+        eval_loss_file_path = os.path.join(tempdir, "eval_loss.jsonl")
```
Ssukriti, line 159: we can move it inside validate training as well as mentioned below

tests/test_sft_trainer.py
```diff
@@ -0,0 +1,291 @@
...
+
+        # Run inference on the text
+        output_inference = loaded_model.run(
+            "Simply put, the theory of relativity states that ", max_new_tokens=50
+        )
+        assert len(output_inference) > 0
+        assert "Simply put, the theory of relativity states that" in output_inference
+
+
+def test_run_train_lora_target_modules():
```
Ssukriti, line 192: whats the difference between this and above test? can we combine to 1?

tests/test_sft_trainer.py
```diff
@@ -0,0 +1,291 @@
...
+        # validate lora tuning configs
+        _validate_training(tempdir)
+        checkpoint_path = os.path.join(tempdir, "checkpoint-5")
+        adapter_config = _get_adapter_config(checkpoint_path)
+        _validate_adapter_config(adapter_config, "LORA", BASE_LORA_KWARGS)
+        for module in lora_target_modules["target_modules"]:
+            assert module in adapter_config.get("target_modules")
+
+
+def test_run_train_lora_target_modules_all_linear():
```
Ssukriti, line 213: excellent test, thank you!

tests/test_sft_trainer.py
```diff
@@ -0,0 +1,291 @@
...
+    "tokenizer_name_or_path": MODEL_NAME,
+    "save_strategy": "epoch",
+    "output_dir": "tmp",
+}
+
+BASE_LORA_KWARGS = copy.deepcopy(BASE_PEFT_KWARGS)
+BASE_LORA_KWARGS["peft_method"] = "lora"
+
+BASE_FT_KWARGS = copy.deepcopy(BASE_PEFT_KWARGS)
+BASE_FT_KWARGS["peft_method"] = ""
```
Ssukriti, line 68: we should also unset `prompt_tuning_init_text` and `prompt_tuning_init_text` ideally

Ssukriti (changes requested):
I think some more tests can be i:
1. if num_epocs and num_gradient_acc steps is 0 , valueerror is returned
2. prompt tuning test can test with `"prompt_tuning_init": "RANDOM",` and `TEXT`

requirements.txt
gkumbhat: In general if there is a specific version that doesn't work, or has a bug ,then we can also ask pip to ignore that particular version. 

https://github.com/foundation-model-stack/fms-hf-tuning/pull/54#discussion_r1516271877

tests/data/__init__.py
```diff
@@ -0,0 +1,22 @@
+# Copyright The IBM Tuning Team
```
gkumbhat, line 1: Curious about the copyright notice..Where is this coming from?

tests/test_sft_trainer.py
```diff
@@ -0,0 +1,325 @@
...
+    """Check if error is raised when invalid params are used to peft tune causallm models"""
+    with tempfile.TemporaryDirectory() as tempdir:
+        invalid_params = copy.deepcopy(BASE_PEFT_KWARGS)
+        invalid_params["output_dir"] = tempdir
+        invalid_params[param_name] = param_val
+        model_args, data_args, training_args, tune_config = causal_lm_train_kwargs(
+            invalid_params
+        )
+
+        with pytest.raises(ValueError, match=exc_msg):
```
Ssukriti, line 191:
```suggestion
        with pytest.raises(ValueError):
```

Ssukriti, line 191: I generally avoid matching exact error message and just check for valueError with a comment explaining why, but will let this go and I dont think we will update the message much

Ssukriti (approved):
Thanks so much!!

PR #89, opened 2024-03-13
Generic tracker API and implementation of Aimstack tracker

dushyantbehl:
### Description of the change

1. This PR adds a tracker API with Aimstack as the default tracker. This is simple plug and play architecture which can support multiple trackers.
1. The tracker config is now taken as command line arguments (making it easier for any automation to pass tracker arguments).
1. With the new API I have added support to track any additional metrics of interest.
1. As an example I have added one single line to track model_load_time in AIM possibly fixing https://github.com/foundation-model-stack/fms-hf-tuning/issues/33 

### Related issue number

1. https://github.com/foundation-model-stack/fms-hf-tuning/issues/34
1. https://github.com/foundation-model-stack/fms-hf-tuning/issues/33

### How to verify the PR

Example of how new api can be invoked

```
 torchrun --nnodes=1 --nproc_per_node=2 --master_port=1234 tuning/sft_trainer.py --tokenizer_name_or_path ${MODEL_PATH} --model_name_or_path ${MODEL_PATH} --data_path ${DATA_PATH} --use_peft --bf16 True --output_dir ${OUTPUT_PATH} --num_train_epochs 1 --per_device_train_batch_size 1 --per_device_eval_batch_size 1 --gradient_accumulation_steps 8 --evaluation_strategy "no" --save_strategy "steps" --save_steps 2000 --save_total_limit 1 --learning_rate 2e-5 --weight_decay 0. --warmup_ratio 0.03 --lr_scheduler_type "cosine" --logging_steps 1 --fsdp "full_shard auto_wrap" --fsdp_config tuning/config/fsdp_config.json --bf16 True --response_template "\n### Response:" --dataset_text_field "output" --tracker aim --aim_repo /data/aim --experiment sft-llama7b-test 
```

### Was the PR tested


tuning/sft_trainer.py
```diff
@@ -254,6 +260,30 @@ def train(
         )
         callbacks.append(tc_callback)
 
+    if train_args.packing:
```
Ssukriti, line 263: why is this code duplicated ? its already in line 215 above?

tuning/sft_trainer.py
```diff
@@ -128,12 +134,16 @@ def train(
...
+    additional_metrics = {}
+
+    model_load_time = time.time()
     model = AutoModelForCausalLM.from_pretrained(
         model_args.model_name_or_path,
         cache_dir=train_args.cache_dir,
         torch_dtype=get_torch_dtype(model_args.torch_dtype),
         attn_implementation="flash_attention_2" if model_args.use_flash_attn else None,
     )
+    additional_metrics["model_load_time"] = time.time() - model_load_time
```
Ssukriti, line 146: wondering if this should also include tokenizer load time as its associated with model? should we move this block here `# TODO: Move these to a config as well
    tokenizer = AutoTokenizer.from_pretrained(
        model_args.model_name_or_path, cache_dir=train_args.cache_dir, use_fast=True
    )` 
    what do you think @dushyantbehl ?

tuning/sft_trainer.py
```diff
@@ -100,6 +98,9 @@ def train(
         Union[peft_config.LoraConfig, peft_config.PromptTuningConfig]
     ] = None,
     trainer_controller_args: configs.TrainerControllerArguments = None,
+    callbacks: Optional[List[TrainerCallback]] = None,
```
Ssukriti, line 101:
```suggestion
    callbacks: Optional[List[TrainerCallback]] = [FileLoggingCallback()],
```

curious if we want to do this instead. While default is set in main(), please be aware not all adopting scripts will call from main() or call using command line. Main function is an example on how to call train.

Example : in our docker image that we ship to SW we call train() directly and dont use command line. https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/build/launch_training.py#L70

In that case if callback not set, do we want to set it to default FileLoggingCallback ? also is that logger arg in __init__ of FileLoggingCallback needed?

.gitignore
```diff
@@ -29,6 +29,12 @@ venv/
 # Tox envs
 .tox
 
+# Build
+build/
```
kmehant, line 33: This is being already contributed by https://github.com/foundation-model-stack/fms-hf-tuning/pull/140 

Also we dont want to ignore complete `build` folder here, the files here should be git tracked.

tuning/sft_trainer.py
dushyantbehl: It was a merging issue. Fixed.

tuning/sft_trainer.py
dushyantbehl: @Ssukriti model load time here just refers to the overhead of filesystem or COS bucket and was just added as an example on how to track extra metrics form the new api. 
I can certainly include tokenizer time or even have it separately.

tuning/utils/import_utils.py
```diff
@@ -15,8 +15,6 @@
 # Third Party
 from transformers.utils.import_utils import _is_package_available
 
-_is_aim_available = _is_package_available("aim")
 
-
-def is_aim_available():
-    return _is_aim_available
+def is_package_available(pkg):
+    return _is_package_available(pkg)
```
kmehant, line 20: Let's not do this, `_is_package_available` function from `transformers` is an expensive operation as hits the filesystem and can be unnecessary performance overhead if `is_package_available` is used widely in the code. Lets revert the changes.

kmehant (changes requested):

tuning/sft_trainer.py
dushyantbehl: @Ssukriti point taken but with this new API do you think that FileLoggingCallback should just be classified as another tracker which tracks logs to a file system. 
In that case we can set the callbacks default to a FileLoggingTracker.

On the logger arg in init of FileLoggingCallback, if I remove it the linter on my system complains of a `no-value-for-parameter` in that line.

```
tuning/sft_trainer.py:349:27: E1120: No value for argument 'logger' in constructor call (no-value-for-parameter)
```

Maybe we could make the logger argument optional? or altogether remove it as its not being called in the callback.

tuning/utils/import_utils.py
dushyantbehl: Not sure if what you mean is clear to me. The package check for `aim` or any other tracker should be done before using them so cannot be avoided.
Do you just mean that people should not be using `is_package_available` where it is not absolutely necessary? 

Shouldn't we rather document the expensive nature of this operation here so that anyone who is using this in code knows what they are calling.

kmehant: > The package check for aim or any other tracker should be done before using them so cannot be avoided.

I agree with you on this. 

> Do you just mean that people should not be using is_package_available where it is not absolutely necessary?

I did not mean this.

The previous code does a one time check for aim package and the function `is_aim_available` pulls from the cache for that process. Which is much more efficient than the current implementation. So please revert the current changes and use `is_aim_available` function in your code.

kmehant (changes requested):

tuning/utils/import_utils.py
dushyantbehl: The current code only registers tracks once and hence calls `_is_package_available` only once.

kmehant: Right, I am just worried the function is now made `is_package_available ` general purpose for any package and this can be used at multiple places going forward. Just to support my point, same design is followed by HuggingFace transformers library to overcome the obvious overhead. However, I am okay with your current change.

dushyantbehl: Sure @kmehant the intent what you say is okay with me...I concur....maybe then I will just change this in another way to not have this problem

tuning/config/tracker_configs.py
```diff
@@ -0,0 +1,33 @@
...
+    #    or a remote repository hosted on a server.
+    # When 'aim_remote_server_ip' or 'aim_remote_server_port' is set,
+    #    it designates a remote aim repo.
+    # Otherwise, 'repo' specifies the directory, with default of None meaning '.aim'.
+    #
+    # See https://aimstack.readthedocs.io/en/latest/using/remote_tracking.html
+    #     for documentation on Aim remote server tracking.
+    aim_repo: str = ".aim"
+    aim_remote_server_ip: str = None
+    aim_remote_server_port: int = None
```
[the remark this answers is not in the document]
dushyantbehl: Sure...i was going to add unit test but this also works

tuning/sft_trainer.py
```diff
@@ -113,12 +114,17 @@ def train(
...
     """
 
     logger = logging.get_logger("sft_trainer")
 
     # Validate parameters
-    if (not isinstance(train_args.num_train_epochs, (float, int))) or (
+    if (not isinstance(train_args.num_train_epochs, float)) or (
         train_args.num_train_epochs <= 0
     ):
         raise ValueError("num_train_epochs has to be an integer/float >= 1")
```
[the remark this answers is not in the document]
dushyantbehl: This was by mistake...thanks for catching.

tuning/sft_trainer.py
```diff
@@ -284,6 +307,7 @@ def main(**kwargs):  # pylint: disable=unused-argument
             configs.TrainerControllerArguments,
             peft_config.LoraConfig,
             peft_config.PromptTuningConfig,
+            Union[tracker_configs.AimConfig],
```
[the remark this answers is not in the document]
dushyantbehl: Added as an example to move forward with more trackers...not strictly needed right now as we currently have only one...will make more sense when we add other trackers.

tuning/trackers/tracker_factory.py
```diff
@@ -0,0 +1,64 @@
...
+        T = meta["tracker"]
+        tracker_config = C(**dataclasses.asdict(super_configs))
+        if tracker_config is not None:
+            return T(tracker_config)
+    else:
+        logger.warning(
+            "Requested Tracker %s not found.\n"
+            "Please check the argument before proceeding."
+        )
+    return Tracker()
```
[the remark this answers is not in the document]
dushyantbehl: 👍🏽

dushyantbehl:
@Ssukriti @alex-jw-brooks some more changes and fixes as requested by @kmehant 

PR is ready to be reviewed. @tharapalanivel I will run some more e2e tests and bump you, if you have any comments in the meantime on the design, happy to take them.

dushyantbehl:
Thanks @tharapalanivel should be easy to change it like that.

tuning/sft_trainer.py
```diff
@@ -242,9 +249,11 @@ def train(
             "Validation dataset length is %s", len(formatted_validation_dataset)
         )
 
-    callbacks = [FileLoggingCallback(logger)]
-    if is_aim_available():
-        callbacks.append(get_aimstack_callback())
+    file_logger_callback = FileLoggingCallback()
+    if callbacks:
```
Ssukriti, line 253: Should changing 253 like this be better?
   ` if callbacks and file_logger_callback not in callbacks:
`
we only want to append if not passed already?

tuning/sft_trainer.py
Ssukriti: I see you have now handled it in code by setting default callback in train() . What is done is fine and solves the purpose. You can alternatively create a FileLoggingTracker too and pass that as default, but to keep things simple whats done now also works

tuning/sft_trainer.py
Ssukriti: yes I understand its an example, but since you did the example maybe lets add tokenizer time in it too  for completeness :) . From what I see online model load time includes loading of model + associated tokenizer. Thanks

tuning/sft_trainer.py
```diff
@@ -268,6 +277,22 @@ def train(
         peft_config=peft_config,
     )
 
+    # We track additional metrics and experiment metadata after
+    # Trainer object creation to ensure that this is not repeated
+    # multiple times for FSDP runs.
+    if trackers is not None and trainer.is_world_process_zero():
```
Ssukriti, line 283: @dushyantbehl @kmehant just have one question - since we are leaving it to user to pass tracker in train() and pass its associated callbacks as well, will any errors happen in case user passes aim tracker but doesnt pass aim callback? - its ok if errors are returned and user understands they need to correct arguments

I know in main() we extract callbacks from tracker and pass them to train(), but I just thinking if a user misses that step and calls train with incomplete set of callbacks, will it silently be ignored or would there be an indication to user?

tuning/sft_trainer.py
Ssukriti: you can mark this conversation as resolved if you want to stick to existing implementation as it solves thw purpose. Thanks

.gitignore
Ssukriti: I have approved 140 as well and ran the checks, once checks pass , @kmehant feel free to merge 140 as you have write access

Ssukriti: also thanks for catching this , I missed it :)

tuning/sft_trainer.py
Ssukriti: do you think we should move that code here to extract callbacks from a tracker in train function ? and in callbacks arguments, users should give any additional not specified by tracker? 
or do you want users to manually pass callbacks to train() that are needed for tracker as separate argument - is there a use case where users may not want to use callbacks of a tracker/ optionally select some?

tuning/sft_trainer.py
dushyantbehl: @Ssukriti added and updated the code.

tuning/sft_trainer.py
dushyantbehl: I did just that and updated the code.

tuning/sft_trainer.py
```diff
@@ -128,20 +87,60 @@ def train(
...
+
+    # We add file logging tracker as default
+    if "file_logger" not in requested_trackers:
+        requested_trackers.add("file_logger")
+
+    # Now initialize trackers one by one
+    for name in requested_trackers:
+        t = get_tracker(name, tracker_config_args)
+        cb = t.get_hf_callback()
+        if cb is not None:
```
Ssukriti, line 109: if cb is not None and cb not in  trainer_callbacks:

tuning/sft_trainer.py
```diff
@@ -113,6 +67,11 @@ def train(
             The peft configuration to pass to trainer
         trainer_control_args: configs.TrainerControllerArguments \
             for controlling the training loop using policy rules
+        callbacks: List of callbacks to attach with SFTtrainer.
```
Ssukriti, line 70:
```suggestion
        additional_callbacks: List of callbacks to attach with SFTtrainer.
```

tuning/sft_trainer.py
Ssukriti: looks good to me! Great change!

tuning/sft_trainer.py
```diff
@@ -113,6 +67,11 @@ def train(
             The peft configuration to pass to trainer
         trainer_control_args: configs.TrainerControllerArguments \
             for controlling the training loop using policy rules
+        additional_callbacks: List of callbacks to attach with SFTtrainer , besides those associated with experiment trackers or TrainerControllers. Callbacks associated with tracker with automatically be added.
+        trackers: List of the available trackers in trackers.tracker_factory.REGISTERED_TRACKERS
```
Ssukriti, line 71: there doesnt seem to be an argument `trackers` anymore

tuning/config/configs.py
```diff
@@ -90,6 +90,14 @@ class TrainingArguments(transformers.TrainingArguments):
             'steps' (logging is done every `logging_steps`)"
         },
     )
+    trackers: Optional[List[str.lower]] = field(
+        default=None,
```
Ssukriti, line 94: the default here is None, but docstring says `file_logger(default),`

tuning/sft_trainer.py
```diff
@@ -292,23 +296,63 @@ def main(**kwargs):  # pylint: disable=unused-argument
...
+            logger.error(
+                "failed while parsing extra metadata. pass a valid json %s", repr(e)
+            )
+
+    combined_tracker_configs = {"file_logger": file_logger_config, "aim": aim_config}
+
+    train(
+        model_args=model_args,
+        data_args=data_args,
+        train_args=training_args,
```
Ssukriti, line 349: I still have doubt on what default tracker is passed with training_args here. wby default if users use command line, we want the logs written to log file like we were earlier.

Ssukriti: oh I see you added default tracker inside code and not through arguments

tuning/sft_trainer.py
```diff
@@ -100,6 +51,9 @@ def train(
         Union[peft_config.LoraConfig, peft_config.PromptTuningConfig]
     ] = None,
     trainer_controller_args: configs.TrainerControllerArguments = None,
+    tracker_config_args: Optional[Dict] = None,
```
Ssukriti, line 54: if default tracker is filelogger inside train() , dont you need fileloggingConfig default config here? why not set default value in training_args.tracker=FilElOggingTracker and default value of `tracker_config_args` to FileLoggingConfig.

inside code we need not set it then . If users have not overriden default arguments, they will get fileLogging by default. If they have overriden, we will use what user sets ? if they have explictyly overriden with None, then we can let it be None?

Ssukriti (commented):
comments only on default value of arguments.

By default I want to ensure existing behavior is retained - any users that call .train() or main() function using command line, should see log files like before. So filelogger has to be default for train() and through command line. 
Only if aim is explictly passed, it should be used.

tuning/sft_trainer.py
dushyantbehl: @Ssukriti sure ... I made the Filelogging tracker explitct and passed from user point of view...default in the training args and tracker config default is also FileloggerTrackerConfig.

tuning/sft_trainer.py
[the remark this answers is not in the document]
dushyantbehl: fixed.

tuning/config/configs.py
dushyantbehl: fixed.

tuning/sft_trainer.py
dushyantbehl: fixed

kmehant (approved):

Issue #87, opened 2024-03-13
Switch to accelerate for Multi GPU 


Ssukriti:
## Is your feature request related to a problem? Please describe.

Accelerate https://huggingface.co/docs/transformers/en/accelerate is created by HF  to help users easily train a Transformers model on any type of distributed setup, whether it is multiple GPU’s on one machine or multiple GPU’s across several machines. We should leverage the library for its ease of use. 

## Describe the solution you'd like

1. update the [README.md](https://github.com/foundation-model-stack/fms-hf-tuning?tab=readme-ov-file#multiple-gpus-with-fsdp) removing instructions for torch.run and replace with accelerate.launch
2. replace the [FSDP JSON](https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/tuning/config/fsdp_config.json) with a [config yaml like this](https://huggingface.co/docs/accelerate/en/basic_tutorials/launch#custom-configurations).
 3. move fsdp_config.json out of tuning/config (which houses code) into somewhere which only houses config fixtures.

## Describe alternatives you've considered

torchrun used currently can be less user friendly . See https://github.com/foundation-model-stack/fms-hf-tuning/issues/80

(@fabianlim thanks for the suggestions)

fabianlim:
If we consider the constrained problem of running gpu jobs within a single pod, where each gpu is handled by a single process. There are the following options where mutiple processes are run:
1. all together in a  *single* container, within a  *single* Kube pod
2. individually each within a container, all containers housed within a *single* Kube pod.

There is also the third option where the processes are distributed across *multiple* Kube pods, but this may be over-complex.

Huggingface's recommendation is to run distributed training jobs using [accelerate](https://huggingface.co/docs/accelerate/en/index). 
- build on top of `torchrun`. The main process will spawn multiple worker processes for distributed training.
- `torchrun` has a [watchdog agent](https://github.com/pytorch/pytorch/blob/main/torch/distributed/elastic/agent/server/local_elastic_agent.py) that handles things like [fault tolerance](https://pytorch.org/docs/stable/elastic/agent.html).
- processes communicates via various backends (e.g. `static` or `c10d`).
- GPU communicates over GPU-network interfaces (e.g., NCCL).


Option 1:
- is docker compliant to run [multiple processes to parallelise](https://docs.docker.com/config/containers/multi-service_container/) work (e.g., running workers to parallelize an SQL query). This is very much like a master process spawning many child processes for distributed training.
[...]

Issue #80, opened 2024-03-06
bug: Boolean values are represented as strings in default fsdp config defaults to True


kmehant:
## Describe the bug

Boolean values in fsdp config (https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/tuning/config/fsdp_config.json#L4-L6) are represented as string values. This does not translate to actual boolean values when loaded in hf/transformers library https://github.com/huggingface/transformers/blob/2a002d073a337051bdc3fbdc95ff1bc0399ae2bb/src/transformers/training_args.py#L1654

## Solution
Use json boolean values instead of string representation. Meanwhile, I have as well raised an issue for discussion on transformers library to add a dataclass and argument parser with a motivating example (https://github.com/huggingface/transformers/issues/29476)

fabianlim:
@kmehant while the fix is as you described, now that #53 is merged, I think it may be best to switch to accelerate, which uses a yaml config defaults file. With yaml explicit encasement of strings using `"` and `'` are not necessary, and it will be more robust to such issues. I suggest the following changes
1. update the [README.md](https://github.com/foundation-model-stack/fms-hf-tuning?tab=readme-ov-file#multiple-gpus-with-fsdp) removing instructions for `torch.run` and replace with `accelerate.launch`
2. replace the [FSDP JSON](https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/tuning/config/fsdp_config.json) with a [config yaml like this](https://huggingface.co/docs/accelerate/en/basic_tutorials/launch#custom-configurations).
   - BTW I suggest to move `fsdp_config.json` out of `tuning/config` (which houses code) into somewhere which only houses config fixtures.

@Ssukriti

Ssukriti:
Thanks Fabian, created issue for README updates https://github.com/foundation-model-stack/fms-hf-tuning/issues/87 . We will prioritize it at earliest

kmehant:
https://github.com/foundation-model-stack/fms-hf-tuning/issues/80#issuecomment-1989734269

> I think it may be best to switch to accelerate, which uses a yaml config defaults file. 

Thanks @fabianlim, I am aware of this, isn't accelerate a wrapper over torch.distributed?

> I suggest the following changes

I guess @Ssukriti is tracking them in a different issue https://github.com/foundation-model-stack/fms-hf-tuning/issues/87

Ssukriti:
[...]

PR #91, opened 2024-03-13
Multi-GPU switch from TorchRun to Accelerate

fabianlim:
### Description of the change

The current codebase recommends `torchrun` for multi-gpu training. While `torchrun` is the de-facto way to launch distributed jobs, if used wrongly, it may be incompatible with huggingface trainers and models. The huggingface recommended approach is to use `accelerate` (a layer built on top of `torchrun`) to launch multi-gpu jpbs.

### Related issue number

Addresses #87 . Also in part addresses #80. 

### How to verify the PR

Test the run instructions given in the updated README.md

### Was the PR tested

Had @VassilisVassiliadis and @fabianlim  tested the scripts.


Ssukriti:
merged as part of other PR

PR #95, opened 2024-03-18
Update README.md

Ssukriti:
small change to update a variable name , reviewers - ignore, just merging

### Was the PR tested


anhuong (approved):
Thanks for catching the missed change!

PR #97, opened 2024-03-21
Add unit tests for various edge cases

alex-jw-brooks:
This PR adds some extra unit tests for various edge cases. It also fixes (one of) the edge cases that has validation which is currently unreachable. It also changes a few stray `sys.exit()` calls to just raise errors, since I don't think there's a special reason to have those there at the moment.

NOTE: Some of these tests are testing errors that are thrown by HF libraries for things we don't explicitly validate ourselves to ensure the behavior is stable with version bumps.

### Related issue number
This is a follow-up on: https://github.com/foundation-model-stack/fms-hf-tuning/issues/74

We made the decision to split the PR into two parts, to get the tests covering the base functionality it earlier.

### Was the PR tested

### DIFF
--- a/tests/test_sft_trainer.py
+++ b/tests/test_sft_trainer.py
@@ -409,19 +409,6 @@ def test_empty_data():
         sft_trainer.train(model_args, data_args, training_args, tune_config)
 
 
-def test_data_path_does_not_exist():
-    """Ensure that we get a FileNotFoundError if the data is missing completely."""
-    TRAIN_KWARGS = {
-        **BASE_PEFT_KWARGS,
-        **{"training_data_path": "/foo/bar/foobar", "output_dir": "foo/bar/baz"},
-    }
-    model_args, data_args, training_args, tune_config = causal_lm_train_kwargs(
-        TRAIN_KWARGS
-    )
-    with pytest.raises(FileNotFoundError):
-        sft_trainer.train(model_args, data_args, training_args, tune_config)
-
-
 def test_data_path_is_a_directory():
     """Ensure that we get FileNotFoundError if we point the data path at a dir, not a file."""
     with tempfile.TemporaryDirectory() as tempdir:
--- a/tests/test_sft_trainer.py
+++ b/tests/test_sft_trainer.py
@@ -419,8 +419,9 @@ def test_data_path_is_a_directory():
         model_args, data_args, training_args, tune_config = causal_lm_train_kwargs(
             TRAIN_KWARGS
         )
-        # Confusingly, if we pass a directory for our data path, it will throw a FileNotFoundError saying
-        # "unable to find '<data_path>'", since it can't find a matchable file in the path.
+        # Confusingly, if we pass a directory for our data path, it will throw a
+        # FileNotFoundError saying "unable to find '<data_path>'", since it can't
+        # find a matchable file in the path.
         with pytest.raises(FileNotFoundError):
             sft_trainer.train(model_args, data_args, training_args, tune_config)
 
--- a/tests/test_sft_trainer.py
+++ b/tests/test_sft_trainer.py
@@ -489,7 +489,8 @@ def test_no_packing_needs_reponse_template():
 
 ### Tests for model dtype edge cases
 @pytest.mark.skipif(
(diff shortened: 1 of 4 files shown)


tests/test_sft_trainer.py
```diff
@@ -323,3 +327,191 @@ def _validate_adapter_config(adapter_config, peft_type, base_kwargs):
...
+        model_args, data_args, training_args, tune_config = causal_lm_train_kwargs(
+            TRAIN_KWARGS
+        )
+        with pytest.raises(ValueError):
+            sft_trainer.train(model_args, data_args, training_args, tune_config)
+
+
+### Tests for model dtype edge cases
+@pytest.mark.skipif(
+    not (torch.cuda.is_available() and torch.cuda.is_bf16_supported()),
```
alex-jw-brooks, line 488: `cuda` check is here needed because the bf16 check throws if no Nvidia drivers are available

tests/test_sft_trainer.py
```diff
@@ -323,3 +327,191 @@ def _validate_adapter_config(adapter_config, peft_type, base_kwargs):
...
+    """Ensure that we get a FileNotFoundError if the data is missing completely."""
+    TRAIN_KWARGS = {
+        **BASE_PEFT_KWARGS,
+        **{"training_data_path": "/foo/bar/foobar", "output_dir": "foo/bar/baz"},
+    }
+    model_args, data_args, training_args, tune_config = causal_lm_train_kwargs(
+        TRAIN_KWARGS
+    )
+    with pytest.raises(FileNotFoundError):
+        sft_trainer.train(model_args, data_args, training_args, tune_config)
```
anhuong, line 418: this test already exists -- https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/tests/test_sft_trainer.py

tests/test_sft_trainer.py
```diff
@@ -323,3 +327,191 @@ def _validate_adapter_config(adapter_config, peft_type, base_kwargs):
...
+            sft_trainer.train(model_args, data_args, training_args, tune_config)
+
+
+### Tests for model dtype edge cases
+@pytest.mark.skipif(
+    not (torch.cuda.is_available() and torch.cuda.is_bf16_supported()),
+    reason="Only runs if bf16 is unsupported",
+)
+def test_bf16_still_tunes_if_unsupported():
+    """Ensure that even if bf16 is not supported, tuning still works without problems."""
```
anhuong, line 492: interesting test case! can you explain why it doesn't fail and tuning still works and why this is the preferred expected behavior?

anhuong (commented):
Great test cases, thanks Alex! Left a few questions...

tests/test_sft_trainer.py
alex-jw-brooks: As far as I understand, in devices where bfloat16 is unsupported, there is usually fallback behavior to a supported data type, which is usually float32 since bfloat16 and float32 have the same exponent size!

tuning/sft_trainer.py
```diff
@@ -194,6 +193,26 @@ def train(
...
+    if train_args.packing:
+        logger.info("Packing is set to True")
+        data_collator = None
+        packing = True
+    else:
+        logger.info("Packing is set to False")
+        if data_args.response_template is None:
+            # TODO: Fix this, currently unreachable due to crashing in batch encoding tokenization
+            # We should do this validation up front, then do the encoding, then handle the collator
+            raise ValueError("Response template is None, needs to be set for training")
```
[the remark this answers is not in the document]
alex-jw-brooks: Yeah, you can't encode a None type with a tokenizer since tokenizers generally expect an input of type ` Union[TextInputSequence, Tuple[InputSequence, InputSequence]]`. It would be best to do that in a separate PR to keep things atomic even though it's a simple change, since some of the validation logic is a little bit delicate

anhuong (approved):
Other than the linter error for a line being too long, the tests look good! Thanks Alex
```
tests/test_sft_trainer.py:422:0: C0301: Line too long (105/100) (line-too-long)
```

anhuong (approved):
Thanks Alex!

Issue #88, opened 2024-03-13
Update launch training for multi GPU training


Ssukriti:
## Is your feature request related to a problem? Please describe.

Currently l[aunch training used in dockerfile ](https://github.com/foundation-model-stack/fms-hf-tuning/tree/main/build)for integration with rest of stack only supports python jobs for single GPU training. This issue is to update the script with accelerate for launching multi GPU tuning. 

## Describe the solution you'd like

1. As issue #87 is done, also update launch_training.py to use the same
2. Test docker image with multi GPU training

PR #98, opened 2024-03-25
feat: update launch training with accelerate for multi-gpu

anhuong:
### Description of the change

- Adds additional script that parses params for `accelerate launch` which calls original launch_training to support multi (and single) GPU tuning
- Dockerfile changes:
   - Add new script and default FSDP config to image
   - Give tuning user ownership of /tmp dir - helps with running fms-hf-tuning code
   - Update default command to run accelerate launch script
- Update logging messages launch_training

### Related issue number

closes: #88 

### How to verify the PR

Tested script running fine tuning on multi-GPU cluster for large models that required loading on multi-GPUs to run. Ran with the default FSDP config provided.

Testing with FSDP, 3 GPUs, FT with granite-13b-instruct-v2' model:

```sh
# ran script in pod connected to 3 GPUs with debug logging set
$ python /app/accelerate_launch.py 

# Input params parsed
INFO:root:Using multi-GPU configs: {'num_machines': 1, 'num_processes': 3, 'main_process_port': 1234, 'config_file': '/tmp/fms-hf-tuning/fixtures/accelerate_fsdp_defaults.yaml'}
DEBUG:root:multi_gpu_args: ['--num_machines', '1', '--num_processes', '3', '--main_process_port', '1234', '--config_file', '/tmp/fms-hf-tuning/fixtures/accelerate_fsdp_defaults.yaml', '/app/launch_training.py']
/usr/local/lib/python3.11/site-packages/accelerate/utils/launch.py:232: FutureWarning: `fsdp_backward_prefetch_policy` is deprecated and will be removed in version 0.27.0 of 🤗 Accelerate. Use `fsdp_backward_prefetch` instead
  warnings.warn(
INFO:root:Using nproc_per_node=3.

# note that below you see launch_training script triggered 3 times
INFO:root:Attempting to launch training script
INFO:root:Attempting to launch training script
INFO:root:Attempting to launch training script

# note that below is also printed 3 times for each process running launch_training script
DEBUG:root:Input params parsed: {'multiGPU': {'num_machines': 1, 'num_processes': 3, 'main_process_port': 1234, 'config_file': '/tmp/fms-hf-tuning/fixtures/accelerate_fsdp_defaults.yaml'}, 'model_name_or_path': '/granite-pvc/granite-13b-instruct-v2', 'training_data_path': 'twitter_complaints.json', 'output_dir': 'granite-13b-instruct-v2-ft-multigpu', 'num_train_epochs': 5.0, 'per_device_train_batch_size': 4, 'per_device_eval_batch_size': 4, 'gradient_accumulation_steps': 4, 'evaluation_strategy': 'no', 'save_strategy': 'epoch', 'learning_rate': 0.03, 'weight_decay': 0.0, 'lr_scheduler_type': 'cosine', 'logging_steps': 1.0, 'packing': False, 'include_tokens_per_second': True, 'response_template': '\n### Label:', 'dataset_text_field': 'output', 'use_flash_attn': False, 'torch_dtype': 'bfloat16', 'tokenizer_name_or_path': '/granite-pvc/granite-13b-instruct-v2'}

# includes lots of logs on cuda fsdp resource usage
[rank0]:[2024-03-22 21:21:45,850] torch.distributed.fsdp._debug_utils: [WARNING] FSDP _optim_state_dict() profiling:  defaultdict(<class 'float'>, {'preprocessing': 0.04680833197198808, 'preprocessing_with_comm': 0.003385562915354967, <Type.ALLGATHER_OBJ: 'all_gather_object'>: 0.11370941274799407, <Type.ALLGATHER: 'all_gather'>: 0.40186795219779015, <Type.D2H: 'D2H'>: 22.668807602720335, <Type.RESHARDING: 'resharding'>: 23.291323721176013, 'state_converting': 23.29437973885797, <Type.ALL: 'all'>: 23.34716020594351})

....
# successful tuning
INFO:root:Copying last checkpoint checkpoint-5 into output dir /platform-testing-cos/anhuong/granite-13b-instruct-v2-ft-multigpu
{'train_runtime': 502.18, 'train_samples_per_second': 0.498, 'train_steps_per_second': 0.01, 'train_tokens_per_second': 10.833, 'train_loss': 109.41771049499512, 'epoch': 4.0}
INFO:root:Copying last checkpoint checkpoint-5 into output dir /platform-testing-cos/anhuong/granite-13b-instruct-v2-ft-multigpu
100%|███████████████████████████████████████████████████████████████████████████████| 5/5 [08:21<00:00, 70.52s/it]DEBUG:aim.sdk.run:Closing resource <aim.ext.resource.tracker.ResourceTracker object at 0x7fd54ce67710>

$ ls granite-13b-instruct-v2-ft-multigpu
config.json			  model-00005-of-00006.safetensors  rng_state_1.pth	     train_loss.jsonl
generation_config.json		  model-00006-of-00006.safetensors  rng_state_2.pth	     trainer_state.json
model-00001-of-00006.safetensors  model.safetensors.index.json	    scheduler.pt	     training_args.bin
model-00002-of-00006.safetensors  optimizer.bin			    special_tokens_map.json
model-00003-of-00006.safetensors  pytorch_model_fsdp.bin	    tokenizer.json
model-00004-of-00006.safetensors  rng_state_0.pth		    tokenizer_config.json
```

multi-gpu shown in nvidia-smi output:
```sh
# note 3 processes running launch_training script
$ ps aux
USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
tuning         1  0.0  0.0   5320  2520 ?        Ss   20:51   0:00 /usr/bin/coreutils --coreutils-prog-shebang=sle
tuning        19  0.0  0.0   4920  4092 pts/0    Ss   20:52   0:00 bash
tuning       378  1.7  0.0 5948140 459388 pts/0  Sl+  21:11   0:03 python /app/accelerate_launch.py
tuning       444 51.0  2.0 127945092 27616120 ?  Rsl  21:11   1:44 /bin/python -u /app/launch_training.py
tuning       445 41.8  0.1 104744860 1518380 ?   Rsl  21:11   1:25 /bin/python -u /app/launch_training.py
tuning       446 41.7  0.1 101277092 1509280 ?   Rsl  21:11   1:25 /bin/python -u /app/launch_training.py
tuning       491  0.0  0.0   4840  4020 pts/1    Ss   21:12   0:00 bash
tuning       585  0.0  0.0   7536  3480 pts/1    R+   21:15   0:00 ps aux

# initial loading of the model
Fri Mar 22 21:15:15 2024       
+---------------------------------------------------------------------------------------+
| NVIDIA-SMI 535.104.05             Driver Version: 535.104.05   CUDA Version: 12.2     |
|-----------------------------------------+----------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |         Memory-Usage | GPU-Util  Compute M. |
|                                         |                      |               MIG M. |
|=========================================+======================+======================|
|   0  NVIDIA A100-SXM4-80GB          On  | 00000000:0A:04.0 Off |                    0 |
| N/A   30C    P0              84W / 400W |  75997MiB / 81920MiB |     22%      Default |
|                                         |                      |             Disabled |
+-----------------------------------------+----------------------+----------------------+
|   1  NVIDIA A100-SXM4-80GB          On  | 00000000:0C:05.0 Off |                    0 |
| N/A   32C    P0             105W / 400W |  75717MiB / 81920MiB |    100%      Default |
|                                         |                      |             Disabled |
+-----------------------------------------+----------------------+----------------------+
|   2  NVIDIA A100-SXM4-80GB          On  | 00000000:0C:06.0 Off |                    0 |
| N/A   30C    P0             113W / 400W |  75809MiB / 81920MiB |     99%      Default |
|                                         |                      |             Disabled |
+-----------------------------------------+----------------------+----------------------+

# running fine tuning
Fri Mar 22 21:22:03 2024       
+---------------------------------------------------------------------------------------+
| NVIDIA-SMI 535.104.05             Driver Version: 535.104.05   CUDA Version: 12.2     |
|-----------------------------------------+----------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |         Memory-Usage | GPU-Util  Compute M. |
|                                         |                      |               MIG M. |
|=========================================+======================+======================|
|   0  NVIDIA A100-SXM4-80GB          On  | 00000000:0A:04.0 Off |                    0 |
| N/A   29C    P0              71W / 400W |  75999MiB / 81920MiB |      0%      Default |
|                                         |                      |             Disabled |
+-----------------------------------------+----------------------+----------------------+
|   1  NVIDIA A100-SXM4-80GB          On  | 00000000:0C:05.0 Off |                    0 |
| N/A   32C    P0              90W / 400W |  75717MiB / 81920MiB |    100%      Default |
|                                         |                      |             Disabled |
+-----------------------------------------+----------------------+----------------------+
|   2  NVIDIA A100-SXM4-80GB          On  | 00000000:0C:06.0 Off |                    0 |
| N/A   29C    P0              95W / 400W |  75809MiB / 81920MiB |    100%      Default |
|                                         |                      |             Disabled |
+-----------------------------------------+----------------------+----------------------+
```

### Was the PR tested

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -116,6 +116,9 @@ COPY launch_training.py /app
 RUN chmod +x /app/launch_training.py
 COPY accelerate_launch.py /app
 RUN chmod +x /app/accelerate_launch.py
+COPY accelerate_fsdp_defaults.yaml /app
+
+ENV FSDP_DEFAULTS_FILE_PATH="/app/accelerate_fsdp_defaults.yaml"
 
 # Need a better way to address this hack
 RUN touch /.aim_profile && \
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -114,6 +114,8 @@ COPY LICENSE /licenses/
 
 COPY launch_training.py /app
 RUN chmod +x /app/launch_training.py
+COPY accelerate_launch.py /app
+RUN chmod +x /app/accelerate_launch.py
 
 # Need a better way to address this hack
 RUN touch /.aim_profile && \
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -125,8 +125,8 @@ RUN touch /.aim_profile && \
 
 # create tuning user and give ownership to dirs
 RUN useradd -u $USER_UID tuning -m -g 0 --system && \
-    chown -R $USER:0 /app && \
-    chmod -R g+rwX /app
+    chown -R $USER:0 /app /tmp/fms-hf-tuning && \
+    chmod -R g+rwX /app /tmp/fms-hf-tuning
 
 WORKDIR /app
 USER ${USER}
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -109,7 +109,7 @@ RUN mkdir -p /licenses
 COPY LICENSE /licenses/
 
 # Copy scripts and default configs
-COPY launch_training.py accelerate_launch.py accelerate_fsdp_defaults.yaml /app/
+COPY build/launch_training.py build/accelerate_launch.py fixtures/accelerate_fsdp_defaults.yaml /app/
 RUN chmod +x /app/launch_training.py /app/accelerate_launch.py
 
 ENV FSDP_DEFAULTS_FILE_PATH="/app/accelerate_fsdp_defaults.yaml"
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -112,11 +112,9 @@ RUN git clone https://github.com/foundation-model-stack/fms-hf-tuning.git && \
 RUN mkdir -p /licenses
 COPY LICENSE /licenses/
 
-COPY launch_training.py /app
-RUN chmod +x /app/launch_training.py
-COPY accelerate_launch.py /app
-RUN chmod +x /app/accelerate_launch.py
-COPY accelerate_fsdp_defaults.yaml /app
+# Copy scripts and default configs
+COPY launch_training.py accelerate_launch.py accelerate_fsdp_defaults.yaml /app/
+RUN chmod +x /app/launch_training.py /app/accelerate_launch.py
 
 ENV FSDP_DEFAULTS_FILE_PATH="/app/accelerate_fsdp_defaults.yaml"
 
@@ -128,8 +126,8 @@ RUN touch /.aim_profile && \
 
 # create tuning user and give ownership to dirs
 RUN useradd -u $USER_UID tuning -m -g 0 --system && \
-    chown -R $USER:0 /app /tmp/fms-hf-tuning && \
-    chmod -R g+rwX /app /tmp/fms-hf-tuning
+    chown -R $USER:0 /app /tmp && \
+    chmod -R g+rwX /app /tmp
 
 WORKDIR /app
 USER ${USER}
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -132,4 +132,4 @@ RUN useradd -u $USER_UID tuning -m -g 0 --system && \
 WORKDIR /app
 USER ${USER}
 
-CMD [ "tail", "-f", "/dev/null" ]
+CMD [ "python", "/app/accelerate_launch.py" ]
--- a/build/README.md
+++ b/build/README.md
@@ -1,0 +1,164 @@
+# Building fms-hf-tuning as an Image
+
+The Dockerfile provides a way of running fms-hf-tuning SFT Trainer. It installs the dependencies needed and adds two additional scripts that helps to parse arguments to pass to SFT Trainer. The `accelerate_launch.py` script is run by default when running the image to trigger SFT trainer for single or multi GPU by parsing arguments and running `accelerate launch launch_training.py`. 
+
+## Configuration
+
+The scripts accept a JSON formatted config which are set by environment variables. `SFT_TRAINER_CONFIG_JSON_PATH` can be set to the mounted path of the JSON config. Alternatively, `SFT_TRAINER_CONFIG_JSON_ENV_VAR` can be set to the encoded JSON config using the below function:
+
+```py
+import base64
+
+ def encode_json(my_json_string):
+    base64_bytes = base64.b64encode(my_json_string.encode("ascii"))
+    txt = base64_bytes.decode("ascii")
+    return txt
+```
+
+The keys for the JSON config are all of the flags available to use with [SFT Trainer](https://huggingface.co/docs/trl/sft_trainer#trl.SFTTrainer).
+
+For configuring `accelerate launch`, use key `multiGPU` and pass the set of flags accepted by [accelerate launch](https://huggingface.co/docs/accelerate/package_reference/cli#accelerate-launch).
+
+For example, the below config is used for running with two GPUs and FSDP for PEFT tuning:
+
+Note that `num_processes` which is the total number of processes to be launched in parallel,should match the number of GPUs to run on.
+
+```json
+{
+    "multiGPU": {
+        "num_machines": 1,
+        "main_process_port": 1234,
+        "num_processes": 2,
+        "use_fsdp": true,
+        "fsdp_backward_prefetch_policy": "TRANSFORMER_BASED_WRAP",
+        "fsdp_sharding_strategy": 1,
+        "fsdp_state_dict_type": "FULL_STATE_DICT",
+        "fsdp_cpu_ram_efficient_loading": true,
+        "fsdp_sync_module_states": true
+    },
+    "model_name_or_path": "/llama/7B",
+    "training_data_path": "/data/twitter_complaints.json",
+    "output_dir": "/output/llama-7b-pt-multigpu",
+    "num_train_epochs": 5.0,
+    "per_device_train_batch_size": 4,
+    "per_device_eval_batch_size": 4,
+    "gradient_accumulation_steps": 4,
+    "save_strategy": "epoch",
+    "learning_rate": 0.03,
+    "weight_decay": 0.0,
+    "lr_scheduler_type": "cosine",
+    "logging_steps": 1.0,
+    "packing": false,
+    "include_tokens_per_second": true,
+    "response_template": "\n### Label:",
+    "dataset_text_field": "output",
+    "use_flash_attn": false,
+    "torch_dtype": "bfloat16",
+    "tokenizer_name_or_path": "/llama/7B"
+}
+```
+
+When `multiGPU` is set, the [FSDP config](https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/fixtures/accelerate_fsdp_defaults.yaml) is used by default. Any of these values can be overwritten by passing in flags via the JSON config or by passing in your own config file using key `config_file`.
+
+If `multiGPU` is set and `num_processes` is not explicitly set, the number of processes/GPUs will be determined by the number of GPUs available via `torch.cuda.device_count()`.
+
+If `multiGPU` is not set, the script will assume single-GPU and run with `num_processes=1`. The number of GPUs used can also be set by setting environment variable `CUDA_VISIBLE_DEVICES`.
+
+
+## Building the Image
+
+With docker, build the image with:
+
+```sh
+docker build . -t sft-trainer:mytag
+```
+
+## Running the Image
+
+Run sft-trainer-image with the JSON env var and mounts set up.
+
+```sh
+docker run -v config.json:/app/config.json -v $MODEL_PATH:/model -v $TRAINING_DATA_PATH:/data/twitter_complaints.json --env SFT_TRAINER_CONFIG_JSON_PATH=/app/config.json sft-trainer:mytag
+```
+
+This will run `accelerate_launch.py` with the JSON config passed.
+
+An example Kubernetes Pod for deploying sft-trainer which requires creating PVCs with the model and input dataset and any mounts needed for the outputted tuned model:
+
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+name: sft-trainer-config
+data:
+config.json: |
+    {
+      "multiGPU": {
+            "num_machines": 1,
+            "main_process_port": 1234,
+            "num_processes": 2,
+            "use_fsdp": true,
+            "fsdp_backward_prefetch_policy": "TRANSFORMER_BASED_WRAP",
+            "fsdp_sharding_strategy": 1,
+            "fsdp_state_dict_type": "FULL_STATE_DICT",
+            "fsdp_cpu_ram_efficient_loading": true,
+            "fsdp_sync_module_states": true
+        },
+        "model_name_or_path": "/data/input/llama/7B",
+        "training_data_path": "/data/input/twitter_complaints.json",
+        "output_dir": "/data/output/llama-7b-pt-multigpu",
+        "num_train_epochs": 5.0,
+        "per_device_train_batch_size": 4,
+        "per_device_eval_batch_size": 4,
+        "gradient_accumulation_steps": 4,
+        "save_strategy": "epoch",
+        "learning_rate": 0.03,
+        "weight_decay": 0.0,
+        "lr_scheduler_type": "cosine",
+        "logging_steps": 1.0,
+        "packing": false,
+        "include_tokens_per_second": true,
+        "response_template": "\n### Label:",
+        "dataset_text_field": "output",
+        "use_flash_attn": false,
+        "torch_dtype": "bfloat16",
+        "tokenizer_name_or_path": "/data/input/llama/7B"
+    }
+---
+apiVersion: v1
+kind: Pod
+metadata:
+name: sft-trainer-test
+spec:
+containers:
+    env:
+        - name: SFT_TRAINER_CONFIG_JSON_PATH
+        value: /config/config.json
+    image: sft-trainer:mytag
+    imagePullPolicy: IfNotPresent
+    name: tuning-test
+    resources:
+        limits:
+        nvidia.com/gpu: "1"
+        requests:
+        nvidia.com/gpu: "1"
+    volumeMounts:
+        - mountPath: /data/input
+        name: input-data
+        - mountPath: /data/output
+        name: output-data
+        - mountPath: /config
+        name: sft-trainer-config
+restartPolicy: Never
+terminationGracePeriodSeconds: 30
+volumes:
+    - name: input-data
+    persistentVolumeClaim:
+        claimName: input-pvc
+    - name: output-data
+    persistentVolumeClaim:
+        claimName: output-pvc
+    - name: sft-trainer-config
+    configMap:
+        name: sft-trainer-config
+```
--- a/build/README.md
+++ b/build/README.md
@@ -9,23 +9,26 @@ The scripts accept a JSON formatted config which are set by environment variable
 ```py
 import base64
 
- def encode_json(my_json_string):
+def encode_json(my_json_string):
     base64_bytes = base64.b64encode(my_json_string.encode("ascii"))
     txt = base64_bytes.decode("ascii")
     return txt
+
+with open("test_config.json") as f:
+    contents = f.read()
+
+encode_json(contents)
 ```
 
 The keys for the JSON config are all of the flags available to use with [SFT Trainer](https://huggingface.co/docs/trl/sft_trainer#trl.SFTTrainer).
 
-For configuring `accelerate launch`, use key `multiGPU` and pass the set of flags accepted by [accelerate launch](https://huggingface.co/docs/accelerate/package_reference/cli#accelerate-launch).
+For configuring `accelerate launch`, use key `accelerate_launch_args` and pass the set of flags accepted by [accelerate launch](https://huggingface.co/docs/accelerate/package_reference/cli#accelerate-launch). Since these flags are passed via the JSON config, the key matches the long formed flag name. For example, to enable flag `--quiet`, use JSON key `"quiet"`, using the short formed `"q"` will fail.
 
-For example, the below config is used for running with two GPUs and FSDP for PEFT tuning:
-
-Note that `num_processes` which is the total number of processes to be launched in parallel,should match the number of GPUs to run on.
+For example, the below config is used for running with two GPUs and FSDP for fine tuning:
 
 ```json
 {
-    "multiGPU": {
+    "accelerate_launch_args": {
         "num_machines": 1,
         "main_process_port": 1234,
         "num_processes": 2,
@@ -36,7 +39,8 @@ Note that `num_processes` which is the total number of processes to be launched 
         "fsdp_cpu_ram_efficient_loading": true,
         "fsdp_sync_module_states": true
     },
-    "model_name_or_path": "/llama/7B",
+    "multi_gpu": true,
+    "model_name_or_path": "/llama/13B",
     "training_data_path": "/data/twitter_complaints.json",
     "output_dir": "/output/llama-7b-pt-multigpu",
     "num_train_epochs": 5.0,
@@ -52,17 +56,19 @@ Note that `num_processes` which is the total number of processes to be launched 
     "include_tokens_per_second": true,
     "response_template": "\n### Label:",
     "dataset_text_field": "output",
-    "use_flash_attn": false,
+    "use_flash_attn": true,
     "torch_dtype": "bfloat16",
-    "tokenizer_name_or_path": "/llama/7B"
+    "tokenizer_name_or_path": "/llama/13B"
 }
 ```
 
-When `multiGPU` is set, the [FSDP config](https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/fixtures/accelerate_fsdp_defaults.yaml) is used by default. Any of these values can be overwritten by passing in flags via the JSON config or by passing in your own config file using key `config_file`.
+When `multi_gpu` is set to true, the [FSDP config](https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/fixtures/accelerate_fsdp_defaults.yaml) is used by default. Any of these values can be overwritten by passing in flags via the JSON config or by passing in your own config file using key `config_file`.
 
-If `multiGPU` is set and `num_processes` is not explicitly set, the number of processes/GPUs will be determined by the number of GPUs available via `torch.cuda.device_count()`.
+If `multi_gpu` is set and `num_processes` is not explicitly set, the number of processes/GPUs will be determined by the number of GPUs available via `torch.cuda.device_count()`.
 
-If `multiGPU` is not set, the script will assume single-GPU and run with `num_processes=1`. The number of GPUs used can also be set by setting environment variable `CUDA_VISIBLE_DEVICES`.
+If `multi_gpu` is not set, the script will assume single-GPU and run with `num_processes=1`.
+
+Note that `num_processes` which is the total number of processes to be launched in parallel, should match the number of GPUs to run on. The number of GPUs used can also be set by setting environment variable `CUDA_VISIBLE_DEVICES`.
 
 
 ## Building the Image
@@ -93,7 +99,7 @@ name: sft-trainer-config
 data:
 config.json: |
     {
-      "multiGPU": {
+        "accelerate_launch_args": {
             "num_machines": 1,
             "main_process_port": 1234,
             "num_processes": 2,
@@ -104,9 +110,10 @@ config.json: |
             "fsdp_cpu_ram_efficient_loading": true,
             "fsdp_sync_module_states": true
         },
-        "model_name_or_path": "/data/input/llama/7B",
-        "training_data_path": "/data/input/twitter_complaints.json",
-        "output_dir": "/data/output/llama-7b-pt-multigpu",
+        "multi_gpu": true,
+        "model_name_or_path": "/llama/13B",
+        "training_data_path": "/data/twitter_complaints.json",
+        "output_dir": "/output/llama-7b-pt-multigpu",
         "num_train_epochs": 5.0,
         "per_device_train_batch_size": 4,
         "per_device_eval_batch_size": 4,
@@ -120,9 +127,9 @@ config.json: |
         "include_tokens_per_second": true,
         "response_template": "\n### Label:",
         "dataset_text_field": "output",
-        "use_flash_attn": false,
+        "use_flash_attn": true,
         "torch_dtype": "bfloat16",
-        "tokenizer_name_or_path": "/data/input/llama/7B"
+        "tokenizer_name_or_path": "/llama/13B"
     }
 ---
 apiVersion: v1
--- a/build/README.md
+++ b/build/README.md
@@ -39,7 +39,6 @@ For example, the below config is used for running with two GPUs and FSDP for fin
         "fsdp_cpu_ram_efficient_loading": true,
         "fsdp_sync_module_states": true
     },
-    "multi_gpu": true,
     "model_name_or_path": "/llama/13B",
     "training_data_path": "/data/twitter_complaints.json",
     "output_dir": "/output/llama-7b-pt-multigpu",
@@ -62,13 +61,9 @@ For example, the below config is used for running with two GPUs and FSDP for fin
 }
 ```
 
-When `multi_gpu` is set to true, the [FSDP config](https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/fixtures/accelerate_fsdp_defaults.yaml) is used by default. Any of these values can be overwritten by passing in flags via the JSON config or by passing in your own config file using key `config_file`.
+Users should always set `num_processes` to be explicit about the number of processes to run tuning on. When `num_processes` is greater than 1, the [FSDP config](https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/fixtures/accelerate_fsdp_defaults.yaml) is used by default. Any of these values can be overwritten by passing in flags via `accelerate_launch_args` in the JSON config or by passing in your own config file using key `config_file`.
 
-If `multi_gpu` is set and `num_processes` is not explicitly set, the number of processes/GPUs will be determined by the number of GPUs available via `torch.cuda.device_count()`.
-
-If `multi_gpu` is not set, the script will assume single-GPU and run with `num_processes=1`. This can be overwritten by setting `num_processes` or a `config_file` which contains `num_processes`.
-
-Note that `num_processes` which is the total number of processes to be launched in parallel, should match the number of GPUs to run on. The number of GPUs used can also be set by setting environment variable `CUDA_VISIBLE_DEVICES`.
+Note that `num_processes` which is the total number of processes to be launched in parallel, should match the number of GPUs to run on. The number of GPUs used can also be set by setting environment variable `CUDA_VISIBLE_DEVICES`. If ``num_processes=1`, the script will assume single-GPU.
 
 
 ## Building the Image
--- a/build/README.md
+++ b/build/README.md
@@ -67,10 +67,10 @@ If `multiGPU` is not set, the script will assume single-GPU and run with `num_pr
 
 ## Building the Image
 
-With docker, build the image with:
+With docker, build the image at the top level with:
 
 ```sh
-docker build . -t sft-trainer:mytag
+docker build . -t sft-trainer:mytag -f build/Dockerfile
 ```
 
 ## Running the Image
--- a/build/README.md
+++ b/build/README.md
@@ -66,7 +66,7 @@ When `multi_gpu` is set to true, the [FSDP config](https://github.com/foundation
 
 If `multi_gpu` is set and `num_processes` is not explicitly set, the number of processes/GPUs will be determined by the number of GPUs available via `torch.cuda.device_count()`.
 
-If `multi_gpu` is not set, the script will assume single-GPU and run with `num_processes=1`.
+If `multi_gpu` is not set, the script will assume single-GPU and run with `num_processes=1`. This can be overwritten by setting `num_processes` or a `config_file` which contains `num_processes`.
 
 Note that `num_processes` which is the total number of processes to be launched in parallel, should match the number of GPUs to run on. The number of GPUs used can also be set by setting environment variable `CUDA_VISIBLE_DEVICES`.
 
@@ -146,9 +146,9 @@ containers:
     name: tuning-test
     resources:
         limits:
-        nvidia.com/gpu: "1"
+            nvidia.com/gpu: "2"
         requests:
-        nvidia.com/gpu: "1"
+            nvidia.com/gpu: "2"
     volumeMounts:
         - mountPath: /data/input
         name: input-data
--- a/build/README.md
+++ b/build/README.md
@@ -61,7 +61,7 @@ For example, the below config is used for running with two GPUs and FSDP for fin
 }
 ```
 
-Users should always set `num_processes` to be explicit about the number of processes to run tuning on. When `num_processes` is greater than 1, the [FSDP config](https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/fixtures/accelerate_fsdp_defaults.yaml) is used by default. Any of these values can be overwritten by passing in flags via `accelerate_launch_args` in the JSON config or by passing in your own config file using key `config_file`.
+Users should always set `num_processes` to be explicit about the number of processes to run tuning on. When `num_processes` is greater than 1, the [FSDP config](https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/fixtures/accelerate_fsdp_defaults.yaml) is used by default. You can also set your own default values by specifying your own config file using key `config_file`. Any of these values in configs can be overwritten by passing in flags via `accelerate_launch_args` in the JSON config.
 
 Note that `num_processes` which is the total number of processes to be launched in parallel, should match the number of GPUs to run on. The number of GPUs used can also be set by setting environment variable `CUDA_VISIBLE_DEVICES`. If ``num_processes=1`, the script will assume single-GPU.
 
@@ -105,7 +105,6 @@ config.json: |
             "fsdp_cpu_ram_efficient_loading": true,
             "fsdp_sync_module_states": true
         },
-        "multi_gpu": true,
         "model_name_or_path": "/llama/13B",
         "training_data_path": "/data/twitter_complaints.json",
         "output_dir": "/output/llama-7b-pt-multigpu",
--- a/build/accelerate_launch.py
+++ b/build/accelerate_launch.py
@@ -1,9 +1,31 @@
-from accelerate.commands.launch import launch_command_parser, launch_command
+# Copyright The FMS HF Tuning Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Script wraps launch_training to run with accelerate for multi and single GPU cases.
+Read multiGPU configuration via environment variable `SFT_TRAINER_CONFIG_JSON_PATH`
+for the path to the JSON config file with parameters or `SFT_TRAINER_CONFIG_JSON_ENV_VAR`
+for the encoded config string to parse.
+"""
+
+# Standard
 import json
 import os
 import base64
 import pickle
+import logging
 
+# Third Party
+from accelerate.commands.launch import launch_command_parser, launch_command
 
 def txt_to_obj(txt):
     base64_bytes = txt.encode("ascii")
@@ -16,6 +38,9 @@ def txt_to_obj(txt):
         return pickle.loads(message_bytes)
 
 def main():
+    LOGLEVEL = os.environ.get("LOG_LEVEL", "WARNING").upper()
+    logging.basicConfig(level=LOGLEVEL)
+
     json_configs = {}
     json_path = os.getenv("SFT_TRAINER_CONFIG_JSON_PATH")
     json_env_var = os.getenv("SFT_TRAINER_CONFIG_JSON_ENV_VAR")
@@ -28,21 +53,24 @@ def main():
         json_configs = txt_to_obj(json_env_var)
 
     # parse multiGPU args
-    print("json_configs[multiGPU]", json_configs["multiGPU"])
-
-    multigpu_args = []
+    multi_gpu_args = []
     if json_configs.get("multiGPU"):
+        logging.info("Using multi-GPU configs: %s", json_configs.get("multiGPU"))
         for key, val in json_configs["multiGPU"].items():
-            multigpu_args.append(f"--{key}")
-            multigpu_args.append(str(val))
+            multi_gpu_args.append(f"--{key}")
+            multi_gpu_args.append(str(val))
     
+        # add FSDP config
+        fsdp_filepath = os.getenv("FSDP_DEFAULTS_FILE_PATH", "/app/accelerate_fsdp_defaults.yaml")
+        multi_gpu_args.append("--config_file")
+        multi_gpu_args.append(fsdp_filepath)
+
     # add training_script
-    multigpu_args.append("/app/launch_training.py")
+    multi_gpu_args.append("/app/launch_training.py")
     
-    print("multigpu_args", multigpu_args)
+    logging.debug("multi_gpu_args: %s", multi_gpu_args)
     parser = launch_command_parser()
-    args = parser.parse_args(args=multigpu_args)
-    print(args)
+    args = parser.parse_args(args=multi_gpu_args)
     launch_command(args)
 
 if __name__ == "__main__":
--- a/build/accelerate_launch.py
+++ b/build/accelerate_launch.py
@@ -26,6 +26,7 @@ import logging
 
 # Third Party
 from accelerate.commands.launch import launch_command_parser, launch_command
+import torch
 
 def txt_to_obj(txt):
     base64_bytes = txt.encode("ascii")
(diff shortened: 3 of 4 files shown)


build/accelerate_launch.py
```diff
@@ -0,0 +1,94 @@
...
+                multi_gpu_args.append("--config_file")
+                multi_gpu_args.append(fsdp_filepath)
+
+        # add num_processes to overwrite config file set one
+        if not json_configs.get("multiGPU").get("num_processes"):
+            num_gpus = torch.cuda.device_count()
+            if num_gpus > 1:
+                logging.info(f"Setting accelerate num processes to {num_gpus}")
+                multi_gpu_args.append("--num_processes")
+                multi_gpu_args.append(str(num_gpus))
```
anhuong, line 82: Note that here I have set that we are using `accelerate_fsdp_defaults.yaml` as the default config file, and this file requires setting `num_processes`. So we have to overwrite this value which is set to 1 by default by sending a flag with the value.

If we didn't have to set the FSDP config file, then the accelerate script knows to calculate num of GPUs available and to set that as the num_processeses. accelerate launch script does this calculation [here](https://github.com/huggingface/accelerate/blob/main/src/accelerate/commands/launch.py#L974-L988) and I simplified this to just use `torch.cuda.device_count()`

build/accelerate_launch.py
```diff
@@ -0,0 +1,97 @@
...
+        # add num_processes to overwrite config file set one
+        if not json_configs.get("multiGPU").get("num_processes"):
+            num_gpus = torch.cuda.device_count()
+            if num_gpus > 1:
+                logging.info("Setting accelerate num processes to %s", num_gpus)
+                multi_gpu_args.append("--num_processes")
+                multi_gpu_args.append(str(num_gpus))
+    else:
+        multi_gpu_args.append("--num_processes")
+        multi_gpu_args.append("1")
```
anhuong, line 85: Another note on usage, if multiGPU is not set as an imput param, I set `num_processes` explicitly to be 1 assuming single GPU. If this is not explicitly set, then accelerate launch will try to figure out how many GPUs are available and set that value for num_processes. So even if multiGPU is not set, it can still run multiGPU if a pod has 2 GPUs available. This requires users to be explicit when they want to run multi-GPU.

build/README.md
```diff
@@ -0,0 +1,164 @@
...
+
+For configuring `accelerate launch`, use key `multiGPU` and pass the set of flags accepted by [accelerate launch](https://huggingface.co/docs/accelerate/package_reference/cli#accelerate-launch).
+
+For example, the below config is used for running with two GPUs and FSDP for PEFT tuning:
+
+Note that `num_processes` which is the total number of processes to be launched in parallel,should match the number of GPUs to run on.
+
+```json
+{
+    "multiGPU": {
```
anhuong, line 28: This is technically for the set of flags/input params for `accelerate launch` which runs for single and multi-GPU. Is there a reason someone would want to set the flags for `accelerate launch` but for single-GPU? Because if so, then this key should be changed to `accelerate_launch_params` or something

Ssukriti:
@kmehant @fabianlim for visibility into full stack , could you review this if you have time? 

1. The PR is adding a python script to call accelerate launch ( we know we can do it in multiple ways, we just chose to write a python script)
2. the only parameters that will be passed per job are the number of processes == number of gPUs. Rest will be fixed as described in sample config in README . (sharding strategy etc will be fixed by us. we are using the default FSDP config in the repository) 
3. we have tested the image by itself, and are actively testing integration with Training Operator - which will use image built here

build/accelerate_launch.py
fabianlim: what if `json_configs.get("multiGPU")` returns `None`?

build/accelerate_launch.py
```diff
@@ -0,0 +1,94 @@
...
+    if json_path:
+        with open(json_path, "r", encoding="utf-8") as f:
+            json_configs = json.load(f)
+
+    elif json_env_var:
+        json_configs = txt_to_obj(json_env_var)
+
+    # parse multiGPU args
+    multi_gpu_args = []
+    if json_configs.get("multiGPU"):
```
fabianlim: I think its better to just do a null pattern here

```python
_config = json_configs.get("multiGPU", {})
```

and then below just do `_config.get("num_processes")` etc..

build/accelerate_launch.py
fabianlim: @anhuong btw if you are linking to HF you should use a permalink, because HF merges everyday.

build/README.md
kmehant: I second on moving away from `multiGPU` key name to something like `accelerate_launch_params` This might also alleviate confusion on usage of multi_gpu word for DDP in accelerate distributed type param.

build/README.md
```diff
@@ -0,0 +1,164 @@
...
+        "num_machines": 1,
+        "main_process_port": 1234,
+        "num_processes": 2,
+        "use_fsdp": true,
+        "fsdp_backward_prefetch_policy": "TRANSFORMER_BASED_WRAP",
+        "fsdp_sharding_strategy": 1,
+        "fsdp_state_dict_type": "FULL_STATE_DICT",
+        "fsdp_cpu_ram_efficient_loading": true,
+        "fsdp_sync_module_states": true
```
kmehant, line 37: Trying to think out loud, should we call out that these accelerate parameters are flattened like typically provided to accelerate cli? (I see CLI arg parser is used in the code https://github.com/foundation-model-stack/fms-hf-tuning/pull/98/files#diff-3918b6c0fe504e55b66f13d0cbc2958364e228b0bd7244c67efc5732c5f067f5R91-R92). The reason being users might get confused with this being in JSON to the typical accelerate YAML config (https://github.com/foundation-model-stack/fms-hf-tuning/blob/4cc09cbe4cedff71ef2260ce4847de836df45c90/fixtures/accelerate_fsdp_defaults.yaml)  Where arguments are not flattened and can be a dict type

https://github.com/foundation-model-stack/fms-hf-tuning/blob/4cc09cbe4cedff71ef2260ce4847de836df45c90/fixtures/accelerate_fsdp_defaults.yaml#L11-L15

kmehant: https://github.com/foundation-model-stack/fms-hf-tuning/pull/98/files#diff-4b71ec736aee936feb417e5193e800a1014a8e2df2a5b5b92fd78fbd3d781dbdR61 

May be this point is calling out that difference.

build/Dockerfile
```diff
@@ -112,8 +112,11 @@ RUN git clone https://github.com/foundation-model-stack/fms-hf-tuning.git && \
 RUN mkdir -p /licenses
 COPY LICENSE /licenses/
 
-COPY launch_training.py /app
-RUN chmod +x /app/launch_training.py
+# Copy scripts and default configs
+COPY launch_training.py accelerate_launch.py accelerate_fsdp_defaults.yaml /app/
```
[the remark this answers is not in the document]
kmehant: @anhuong  Since, `accelerate_fsdp_defaults.yaml` not in the current directory where Dockefile is present the provided image build command in the docs would fail isn't it?

https://github.com/foundation-model-stack/fms-hf-tuning/pull/98/files#diff-4b71ec736aee936feb417e5193e800a1014a8e2df2a5b5b92fd78fbd3d781dbdR71-R74

build/accelerate_launch.py
```diff
@@ -0,0 +1,97 @@
...
+        if not json_configs.get("multiGPU").get("config_file"):
+            fsdp_filepath = os.getenv(
+                "FSDP_DEFAULTS_FILE_PATH", "/app/accelerate_fsdp_defaults.yaml"
+            )
+            if os.path.exists(fsdp_filepath):
+                logging.info("Setting accelerate config file to: %s", fsdp_filepath)
+                multi_gpu_args.append("--config_file")
+                multi_gpu_args.append(fsdp_filepath)
+
+        # add num_processes to overwrite config file set one
```
kmehant, line 76: @anhuong You only want to overwrite when default config is being used isn't it? What if the config file from the user is intact and we dont want to touch it? Should we check this in conjunction with the config file and the flags passed in the json config together irrespective of the config file being default or user provided?

kmehant (commented):
Thank you @anhuong, have added few comments for discussion thank you.

build/README.md
fabianlim: @kmehant I see, I think maybe just put a link to the [accelerate docs](https://huggingface.co/docs/accelerate/en/package_reference/cli#accelerate-launch), it should be clear what command line arguments are available. When we talk about command line args, it should be unambiguous that it is a flattened dict

I think if by @kmehant 's suggestion, simply replacing `multGPU` with `accelerate_launch_args` it would make it even clearer.

build/accelerate_launch.py
fabianlim: See my comment below.

build/accelerate_launch.py
fabianlim: `_validate_launch_command` is called because you already used `launch_command`, see [here](https://github.com/huggingface/accelerate/blob/4d415f2129f97da4b16f29b9bf0535fe5f2fb0e8/src/accelerate/commands/launch.py#L1053).

kmehant: @fabianlim correct me if I am wrong, are we dealing with `args.num_machines` argument at all, it should always be 1 isn't it? I thought we were not dealing with multi-node training at this point isn't it? Also, I see  @anhuong is mentioning about num_processes, isn't it?

fabianlim: @kmehant my bad, I have deleted my previous comments.

build/accelerate_launch.py
anhuong: @fabianlim In the bullets do you mean `args.num_processes` since I am not touching the `args.num_machines` in the script? 

`accelerate launch` is a smart script and does a lot of processing of the params passed in. If `num_processes` is never specified, then it can determine how many resources are available and assign `num_processes` appropriately. However, since the `config_file` specifies `num_processes` to be 1,  I am overriding that value to the number of GPUs available if multiGPU is set. If multi-GPU is not set, I explicitly set `num_processes` to be 1 in case the pod has more than 1 GPU allocated, in which if `num_processes` is not passed in as a flag, `accelerate launch` would assume multiGPU if multiple GPUs are available given that no config_file is passed and no flag is passed but multiple GPUs exist.

. This can be overwritten if a flag is passed in

anhuong: Ah I see this comment is further discussed below and seems resolved. Let me know if you have more questions about it

build/Dockerfile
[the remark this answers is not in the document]
anhuong: Yes you are both correct. When I run my image build I have to explicitly copy the file over but I think I have a way to fix this....

build/README.md
```diff
@@ -0,0 +1,164 @@
...
+
+Note that `num_processes` which is the total number of processes to be launched in parallel,should match the number of GPUs to run on.
+
+```json
+{
+    "multiGPU": {
+        "num_machines": 1,
+        "main_process_port": 1234,
+        "num_processes": 2,
+        "use_fsdp": true,
```
anhuong, line 32: no value needed here, the flag is passed is as just `--use_fsdp`. Passing it in as `--use_fsdp True` parses incorrectly. hmm how to parse this as a flag appropriately... --> good catch by @tharapalanivel !

anhuong: there are many params like this (search for `action="store_true"` in https://github.com/huggingface/accelerate/blob/4d415f2129f97da4b16f29b9bf0535fe5f2fb0e8/src/accelerate/commands/launch.py) where the flag is passed with no value. How to best handle this for JSON passing in params...

options:
1. Since these params are not exposed to users widely, only allow a subset of the parameters to be used in BAM/SW (num_processes, master_port, etc). Note that if a user deviates from this, the script will break. A workaround is the user can pass all of these params in as their own config_file. This would overwrite our default config_file. 
2. Hard code the set of params that are of this type and parse them accordingly -- only pass in flag and not value.

Thara and I are leaning towards option 1 especially to start but @Ssukriti @fabianlim let us know your thoughts

kmehant: Adding another approach to solve this.

Given that we are getting the cli params in a well structured format (JSON) we can add one more step of cleaning on top of `launch_command_parser` to handle those specific flags that are `_StoreTrueAction` type class. 

You can do specific handling of such arguments/flags from the `accelerate launch` command parser as

```python
from accelerate.commands.launch import launch_command_parser
parser = launch_command_parser()
for action in parser._actions:
    if (type(action).__name__ == "_StoreTrueAction":
    # This way you know which flags are store true actions and then you can simply include 
    # or ignore them in the args list based on the boolean value in the json and they being store true action type or not.
```

@anhuong

build/README.md
```diff
@@ -0,0 +1,164 @@
...
+
+ def encode_json(my_json_string):
+    base64_bytes = base64.b64encode(my_json_string.encode("ascii"))
+    txt = base64_bytes.decode("ascii")
+    return txt
+```
+
+The keys for the JSON config are all of the flags available to use with [SFT Trainer](https://huggingface.co/docs/trl/sft_trainer#trl.SFTTrainer).
+
+For configuring `accelerate launch`, use key `multiGPU` and pass the set of flags accepted by [accelerate launch](https://huggingface.co/docs/accelerate/package_reference/cli#accelerate-launch).
```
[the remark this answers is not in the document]
anhuong: updated in readme

build/README.md
anhuong: added this in https://github.com/foundation-model-stack/fms-hf-tuning/pull/98/files#diff-3918b6c0fe504e55b66f13d0cbc2958364e228b0bd7244c67efc5732c5f067f5R58-R63

build/accelerate_launch.py
fabianlim: good point @kmehant

anhuong:
@kmehant @fabianlim please provide another review, thank you!

build/accelerate_launch.py
```diff
@@ -55,41 +55,53 @@ def main():
...
+            store_action_params[action.dest] = action
+
+    # parse accelerate_launch_args
+    accelerate_launch_args = []
+    accelerate_config = json_configs.get("accelerate_launch_args", {})
+    if accelerate_config:
+        logging.info("Using accelerate_launch_args configs: %s", accelerate_config)
+        for key, val in accelerate_config.items():
+            # For flags that don't have value, ie. --quiet, only add if value is true
+            if store_action_params.get(key) and val:
```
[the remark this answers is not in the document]
anhuong: i initially was thinking if the value is False that it was not worth adding in. But @tharapalanivel caught that there are StoreActionTrue and StoreActionFalse params. adding in change to address....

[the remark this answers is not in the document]
fabianlim: Right but `StoreActionFalse` just means that if `--key` is specified, then `parse_args.key == False`. 

How about this?

```python

map = {act.dest: action for act in parser._actions}

launch_args = []
for key, val in accelerate_config.items():
   if isinstance(val, bool):
         if (
                (isinstance(map[key], _StoreTrueAction) and val)
                or (isinstance(map[key], _StoreFalseAction) and not val)
         ): 
               launch_args.append(f"--{key}")

         continue
   # ..  handle the non bool args
```

Is abit bloated, and could be refactored to be more compact. But the pattern is just 
1. decide what parser action is is
2. based on `val`, decide how to append launch_args.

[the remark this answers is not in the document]
fabianlim: ok i think @tharapalanivel's solution was very nice.

build/README.md
```diff
@@ -0,0 +1,171 @@
...
+        - name: SFT_TRAINER_CONFIG_JSON_PATH
+        value: /config/config.json
+    image: sft-trainer:mytag
+    imagePullPolicy: IfNotPresent
+    name: tuning-test
+    resources:
+        limits:
+        nvidia.com/gpu: "1"
+        requests:
+        nvidia.com/gpu: "1"
```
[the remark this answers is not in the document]
kmehant: I think we might want to fix indentation here and set number of gpus to 2 as we are having num_processes as 2 (https://github.com/foundation-model-stack/fms-hf-tuning/pull/98/files#diff-4b71ec736aee936feb417e5193e800a1014a8e2df2a5b5b92fd78fbd3d781dbdR104-R105)

build/accelerate_launch.py
```diff
@@ -0,0 +1,111 @@
...
+                accelerate_launch_args.append("--config_file")
+                accelerate_launch_args.append(fsdp_filepath)
+
+            # add num_processes to overwrite config file set one
+            if not accelerate_config.get("num_processes"):
+                num_gpus = torch.cuda.device_count()
+                if num_gpus > 1:
+                    logging.info("Setting accelerate num processes to %s", num_gpus)
+                    accelerate_launch_args.append("--num_processes")
+                    accelerate_launch_args.append(str(num_gpus))
```
kmehant, line 96: Got confused here, do we intend to execute this block only when there is no `config_file` given?

kmehant: Should we not execute it when there is a config file from the user, in that case as well should not we have to prepare `accelerate launch` params isn't it?

kmehant (changes requested):

build/accelerate_launch.py
```diff
@@ -0,0 +1,111 @@
...
+        for key, val in accelerate_config.items():
+            if actions_type_map.get(key) == "_AppendAction":
+                for param_val in val:
+                    accelerate_launch_args.extend([f"--{key}", str(param_val)])
+            else:
+                accelerate_launch_args.append(f"--{key}")
+                # Only need to add key for params that aren't flags ie. --quiet
+                if actions_type_map.get(key) == "_StoreAction":
+                    accelerate_launch_args.append(str(val))
+
```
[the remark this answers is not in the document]
anhuong: If a user set the ipex flag, then it would be passed as `--ipex` which would be setting it to True. But we are assuming these singular flags are always being passed, for example if the user passes in `{"ipex": false}` then we would still pass the flag through `accelerate launch --ipex` which would be setting it to true

build/README.md
```diff
@@ -0,0 +1,171 @@
...
+            "main_process_port": 1234,
+            "num_processes": 2,
+            "use_fsdp": true,
+            "fsdp_backward_prefetch_policy": "TRANSFORMER_BASED_WRAP",
+            "fsdp_sharding_strategy": 1,
+            "fsdp_state_dict_type": "FULL_STATE_DICT",
+            "fsdp_cpu_ram_efficient_loading": true,
+            "fsdp_sync_module_states": true
+        },
+        "multi_gpu": true,
```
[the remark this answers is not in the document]
anhuong: I'm okay with either, whichever is more easily understandable. Distributed technically is more accurate

build/README.md
```diff
@@ -0,0 +1,171 @@
...
+        "accelerate_launch_args": {
+            "num_machines": 1,
+            "main_process_port": 1234,
+            "num_processes": 2,
+            "use_fsdp": true,
+            "fsdp_backward_prefetch_policy": "TRANSFORMER_BASED_WRAP",
+            "fsdp_sharding_strategy": 1,
+            "fsdp_state_dict_type": "FULL_STATE_DICT",
+            "fsdp_cpu_ram_efficient_loading": true,
+            "fsdp_sync_module_states": true
```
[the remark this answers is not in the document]
anhuong: I think currently for the use case of this image running out of the box, it would be difficult for users to know what values to set these to, so by default multi-GPU is using FSDP. This can be overwritten by a new config_file being passed in for DPP or for more advanced/individual use cases, instead of running this script a user could run `accelerate launch` as the command of the image.

build/README.md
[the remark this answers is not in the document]
fabianlim: distributed can mean both multi-gpu or multi-mode. If we dont allow user to set `num_machines`, then I think the `accelerate` accurate term is `multi_gpu`

build/accelerate_launch.py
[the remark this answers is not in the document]
anhuong: @tharapalanivel what do you think of this edge case, we would have to check if the flag is of store_true/store_false and then check the value given that....

build/README.md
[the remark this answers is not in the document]
anhuong: Note that I am not passing `--multi_gpi` in as a flag to `accelerate launch`. I know this flag exists but I saw that passing in the flag gave poor behavior and I believe it's because the `distributed_type` gets set to `MULTI_GPU` instead of `FSDP` so overriding the config_file value. And I believe MULTI_GPU uses DPP instead of FSDP so I was hitting into memory issues when adding flag `--mutli_gpu`

Ssukriti:
firstly thanks everyone on the discussions on PR and for thorough review . We need to provide an updated image with this change to BAM early next week, so may I ask to prioritze this PR and bring to completion by Tuesday next week ? Does that sound reasonable given time difference between India team and US ?

build/accelerate_launch.py
fabianlim: @anhuong sorry im unclear about this, because it is a valid use case that one will profile a `config_file` that contains all of the defaults, but while running experiments, wants to change one key without making a completely new defaults file.

build/README.md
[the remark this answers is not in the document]
fabianlim: @anhuong yes you are right, if `args.mult_gpu == True`, then `args.use_fsdp` will not be set, see [here](https://github.com/huggingface/accelerate/blob/b634388ef1e5d3631e88fdc08eda321c1c69e874/src/accelerate/commands/launch.py#L900).

build/README.md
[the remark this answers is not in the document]
kmehant: Sure @anhuong, its clear to me now, thank you.

build/README.md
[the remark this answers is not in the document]
kmehant: I think lets stay with `multi_gpu` (instead of distributed), thanks for the discussion on this!

build/accelerate_launch.py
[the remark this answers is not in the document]
kmehant: @anhuong  @tharapalanivel 

>  But we are assuming these singular flags are always being passed, for example if the user passes in {"ipex": false} then we would still pass the flag through accelerate launch --ipex which would be setting it to true


Do you think it would not create some confusion while usage? I mean for both `{"ipex": false}` and `{"ipex": true}` from the user, we are passing the flag to the `accelerate launch --ipex`, this is not factoring in the value right? 
Just trying to confirm on the thought process as this would be the interface for the users. 

@fabianlim second thoughts on this thread?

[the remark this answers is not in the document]
fabianlim: @kmehant I think it can be written better, but maybe for now we can settle to skip the key apppend logic in the case of `_StoreTrueAction` or `StoreFalseAction`?

```python
if actions_type_map.get(key) == "_AppendAction":
   for param_val in val:
      accelerate_launch_args.extend([f"--{key}", str(param_val)])
elif not (
   (actions_type_map.get(key) == "_StoreTrueAction" and not val)
   or 
   (actions_type_map.get(key) == "_StoreFalseAction" and val)
):
   accelerate_launch_args.append(f"--{key}")
   if actions_type_map.get(key) == "_StoreAction":
      accelerate_launch_args.append(str(val))
```

build/accelerate_launch.py
Ssukriti: agreed that this coditional block needs to be refactored a bit. This should be the pseudocode-

```
if multi_gpu:
     if no config_file given:
            use default FSDP file
     if num_processes not provided by user:
         num_processes = num_gpus # to ensure it overrides the default file
else:
   #now we know its single GPU
   num_processes=1
```

for the case where it is multi_gpu and num_processes is provided by user, it should have been set in code above, in accelerate_launch_args already

Ssukriti:
@fabianlim @kmehant could you resolve the comments that have been addressed by Anh? Let us leave only ones that need code changes as unresolved. Thank you!

build/accelerate_launch.py
kmehant: Thanks @Ssukriti Above pseudocode looks good to me, thank you!

build/accelerate_launch.py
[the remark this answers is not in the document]
kmehant: Thanks @fabianlim and @tharapalanivel 

@tharapalanivel The code snippet looks good to me, Lets go with this, thank you.

This thread resolved once @tharapalanivel pushes the changes. cc: @Ssukriti

build/README.md
kmehant: This thread is resolved. 
cc: @Ssukriti

build/Dockerfile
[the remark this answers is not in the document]
kmehant: This thread is resolved.
cc: @Ssukriti

build/accelerate_launch.py
kmehant: This thread is resolved.
cc: @Ssukriti

build/accelerate_launch.py
kmehant: This thread is resolved once changes are pushed.
cc: @Ssukriti

kmehant: Thanks @tharapalanivel thats should be enough from my end. Thank you.

anhuong: My logic has it that if a config_file is passed by the user, we do not override with the default config_file or num_processes **we** set. However, we always passed all of the user provided arguments so that any individual field can be ovewritten in the config_file that a user passes and/or the default one set by us. 

So in my logic the user provided arguments is always passed in as `accelerate_launch_args`. If a user passes their own config_file, num_processes must be included for it to be valid.

Please correct me if i'm missing something

build/README.md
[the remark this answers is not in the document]
anhuong: also important to note that I get num_processes warning for multiGPU when single GPU set. I passed in `num_processes=1` into a pod that had 2 GPUs available and got below warning message
```
   WARNING:accelerate.commands.launch:The following values were not passed to `accelerate launch` and had defaults used instead:
		More than one GPU was found, enabling multi-GPU training.
		If this was unintended please pass in `--num_processes=1`.
	`--num_machines` was set to a value of `1`
	`--mixed_precision` was set to a value of `'no'`
	`--dynamo_backend` was set to a value of `'no'`
```
This means that internally multi_gpu is being set, however the value of num_processes is not being overridden otherwise we would see message `--num_machines was set to a value of 2`. However, I think this means DPP is being used. This is an important usage note on if multi-gpu is being used and is on the user to set resources appropriately. Thoughts on this behavior?

build/accelerate_launch.py
anhuong: Note that we have simplified this logic to put the onus on users to pass num_processes explicitly. This determines if running for multi-gpu or single gpu:
```
if num_processes set by user:
   if num_processes > 1:  # multi-gpu
      if not config_file:
         set FSDP config
   else if num_processes == 1:   # single-gpu
      os.environ[CUDA_VISIBLE_DEVICES] = "0"
else:   # num_processes not set
   warn num processes not passed in
   # this means num_processes will be pulled from config_file user provided or will be determined by accelerate launch
```

anhuong:
Tested running a few diff use cases

1. Single-GPU passing in `'accelerate_launch_args': {'num_processes': 1}`

Logs showing parsing:
```sh
INFO:root:Using accelerate_launch_args configs: {'num_processes': 1}
INFO:root:num_processes=1 so setting env var CUDA_VISIBLE_DEVICES=0
DEBUG:root:accelerate_launch_args: ['--num_processes', '1', '/app/launch_training.py']
DEBUG:root:accelerate launch parsed args: Namespace(config_file=None, quiet=None, cpu=False, multi_gpu=False, tpu=False, ipex=False, mixed_precision=None, num_processes=1, num_machines=None, num_cpu_threads_per_process=None, dynamo_backend=None, dynamo_mode='default', dynamo_use_fullgraph=False, dynamo_use_dynamic=False, use_deepspeed=False, use_fsdp=False, use_megatron_lm=False, use_xpu=False, gpu_ids=None, same_network=False, machine_rank=None, main_process_ip=None, main_process_port=None, tee='0', role='default', rdzv_backend='static', rdzv_conf='', max_restarts=0, monitor_interval=5, module=None, no_python=None, tpu_use_cluster=False, tpu_use_sudo=False, vm=None, env=None, main_training_function=None, downcast_bf16=False, deepspeed_config_file=None, zero_stage=None, offload_optimizer_device=None, offload_param_device=None, offload_optimizer_nvme_path=None, offload_param_nvme_path=None, gradient_accumulation_steps=None, gradient_clipping=None, zero3_init_flag=None, zero3_save_16bit_model=None, deepspeed_hostfile=None, deepspeed_exclusion_filter=None, deepspeed_inclusion_filter=None, deepspeed_multinode_launcher=None, fsdp_offload_params='false', fsdp_min_num_params=100000000.0, fsdp_sharding_strategy='FULL_SHARD', fsdp_auto_wrap_policy=None, fsdp_transformer_layer_cls_to_wrap=None, fsdp_backward_prefetch_policy=None, fsdp_backward_prefetch=None, fsdp_state_dict_type=None, fsdp_forward_prefetch='false', fsdp_use_orig_params='true', fsdp_cpu_ram_efficient_loading='true', fsdp_sync_module_states='true', megatron_lm_tp_degree=1, megatron_lm_pp_degree=1, megatron_lm_num_micro_batches=None, megatron_lm_sequence_parallelism=None, megatron_lm_recompute_activations=None, megatron_lm_use_distributed_optimizer=None, megatron_lm_gradient_clipping=1.0, aws_access_key_id=None, aws_secret_access_key=None, debug=None, training_script='/app/launch_training.py', mpirun_hostfile=None, mpirun_ccl=1, training_script_args=[])

WARNING:accelerate.commands.launch:The following values were not passed to `accelerate launch` and had defaults used instead:
	`--num_machines` was set to a value of `1`
	`--mixed_precision` was set to a value of `'no'`
	`--dynamo_backend` was set to a value of `'no'`
To avoid this warning pass in values for each of the problematic parameters or run `accelerate config`.
INFO:root:Initializing launch training script
```
Ran this on a pod with 2 GPUs avail but note that there is no mention of multi-gpu in the accelerate launch logs. Ran tuning successfully on single GPU for llama-7b model.

2. Multi-GPU with setting `'accelerate_launch_args': {'num_processes': 2}`

Logs showing parsing:
```sh
INFO:root:Using accelerate_launch_args configs: {'num_processes': 2}
INFO:root:Using accelerate config file: /app/accelerate_fsdp_defaults.yaml
DEBUG:root:accelerate_launch_args: ['--num_processes', '2', '--config_file', '/app/accelerate_fsdp_defaults.yaml', '/app/launch_training.py']
DEBUG:root:accelerate launch parsed args: Namespace(config_file='/app/accelerate_fsdp_defaults.yaml', quiet=None, cpu=False, multi_gpu=False, tpu=False, ipex=False, mixed_precision=None, num_processes=2, num_machines=None, num_cpu_threads_per_process=None, dynamo_backend=None, dynamo_mode='default', dynamo_use_fullgraph=False, dynamo_use_dynamic=False, use_deepspeed=False, use_fsdp=False, use_megatron_lm=False, use_xpu=False, gpu_ids=None, same_network=False, machine_rank=None, main_process_ip=None, main_process_port=None, tee='0', role='default', rdzv_backend='static', rdzv_conf='', max_restarts=0, monitor_interval=5, module=None, no_python=None, tpu_use_cluster=False, tpu_use_sudo=False, vm=None, env=None, main_training_function=None, downcast_bf16=False, deepspeed_config_file=None, zero_stage=None, offload_optimizer_device=None, offload_param_device=None, offload_optimizer_nvme_path=None, offload_param_nvme_path=None, gradient_accumulation_steps=None, gradient_clipping=None, zero3_init_flag=None, zero3_save_16bit_model=None, deepspeed_hostfile=None, deepspeed_exclusion_filter=None, deepspeed_inclusion_filter=None, deepspeed_multinode_launcher=None, fsdp_offload_params='false', fsdp_min_num_params=100000000.0, fsdp_sharding_strategy='FULL_SHARD', fsdp_auto_wrap_policy=None, fsdp_transformer_layer_cls_to_wrap=None, fsdp_backward_prefetch_policy=None, fsdp_backward_prefetch=None, fsdp_state_dict_type=None, fsdp_forward_prefetch='false', fsdp_use_orig_params='true', fsdp_cpu_ram_efficient_loading='true', fsdp_sync_module_states='true', megatron_lm_tp_degree=1, megatron_lm_pp_degree=1, megatron_lm_num_micro_batches=None, megatron_lm_sequence_parallelism=None, megatron_lm_recompute_activations=None, megatron_lm_use_distributed_optimizer=None, megatron_lm_gradient_clipping=1.0, aws_access_key_id=None, aws_secret_access_key=None, debug=None, training_script='/app/launch_training.py', mpirun_hostfile=None, mpirun_ccl=1, training_script_args=[])
/usr/local/lib/python3.11/site-packages/accelerate/utils/launch.py:232: FutureWarning: `fsdp_backward_prefetch_policy` is deprecated and will be removed in version 0.27.0 of 🤗 Accelerate. Use `fsdp_backward_prefetch` instead
  warnings.warn(
INFO:root:Using nproc_per_node=2.
INFO:root:Initializing launch training script
INFO:root:Initializing launch training script
```

As can be sees, num_processes=2 is passed accordingly and the default FSDP config_file is also passed. Also two processes are triggered as seen from the two launch training script printed at the end.

3. Multi-GPU passing many diff types of params

Logs showing parsing:
```sh
INFO:root:Using accelerate_launch_args configs: {'num_machines': 1, 'main_process_port': 1234, 'num_processes': 2, 'use_fsdp': True, 'fsdp_backward_prefetch_policy': 'TRANSFORMER_BASED_WRAP', 'fsdp_sharding_strategy': 1, 'fsdp_state_dict_type': 'FULL_STATE_DICT', 'fsdp_cpu_ram_efficient_loading': True, 'fsdp_sync_module_states': True}
INFO:root:Using accelerate config file: /app/accelerate_fsdp_defaults.yaml
DEBUG:root:accelerate_launch_args: ['--num_machines', '1', '--main_process_port', '1234', '--num_processes', '2', '--use_fsdp', '--fsdp_backward_prefetch_policy', 'TRANSFORMER_BASED_WRAP', '--fsdp_sharding_strategy', '1', '--fsdp_state_dict_type', 'FULL_STATE_DICT', '--fsdp_cpu_ram_efficient_loading', 'True', '--fsdp_sync_module_states', 'True', '--config_file', '/app/accelerate_fsdp_defaults.yaml', '/app/launch_training.py']
DEBUG:root:accelerate launch parsed args: Namespace(config_file='/app/accelerate_fsdp_defaults.yaml', quiet=None, cpu=False, multi_gpu=False, tpu=False, ipex=False, mixed_precision=None, num_processes=2, num_machines=1, num_cpu_threads_per_process=None, dynamo_backend=None, dynamo_mode='default', dynamo_use_fullgraph=False, dynamo_use_dynamic=False, use_deepspeed=False, use_fsdp=True, use_megatron_lm=False, use_xpu=False, gpu_ids=None, same_network=False, machine_rank=None, main_process_ip=None, main_process_port=1234, tee='0', role='default', rdzv_backend='static', rdzv_conf='', max_restarts=0, monitor_interval=5, module=None, no_python=None, tpu_use_cluster=False, tpu_use_sudo=False, vm=None, env=None, main_training_function=None, downcast_bf16=False, deepspeed_config_file=None, zero_stage=None, offload_optimizer_device=None, offload_param_device=None, offload_optimizer_nvme_path=None, offload_param_nvme_path=None, gradient_accumulation_steps=None, gradient_clipping=None, zero3_init_flag=None, zero3_save_16bit_model=None, deepspeed_hostfile=None, deepspeed_exclusion_filter=None, deepspeed_inclusion_filter=None, deepspeed_multinode_launcher=None, fsdp_offload_params='false', fsdp_min_num_params=100000000.0, fsdp_sharding_strategy='1', fsdp_auto_wrap_policy=None, fsdp_transformer_layer_cls_to_wrap=None, fsdp_backward_prefetch_policy='TRANSFORMER_BASED_WRAP', fsdp_backward_prefetch=None, fsdp_state_dict_type='FULL_STATE_DICT', fsdp_forward_prefetch='false', fsdp_use_orig_params='true', fsdp_cpu_ram_efficient_loading='True', fsdp_sync_module_states='True', megatron_lm_tp_degree=1, megatron_lm_pp_degree=1, megatron_lm_num_micro_batches=None, megatron_lm_sequence_parallelism=None, megatron_lm_recompute_activations=None, megatron_lm_use_distributed_optimizer=None, megatron_lm_gradient_clipping=1.0, aws_access_key_id=None, aws_secret_access_key=None, debug=None, training_script='/app/launch_training.py', mpirun_hostfile=None, mpirun_ccl=1, training_script_args=[])
/usr/local/lib/python3.11/site-packages/accelerate/utils/launch.py:232: FutureWarning: `fsdp_backward_prefetch_policy` is deprecated and will be removed in version 0.27.0 of 🤗 Accelerate. Use `fsdp_backward_prefetch` instead
  warnings.warn(
INFO:root:Using nproc_per_node=2.
INFO:root:Initializing launch training script
INFO:root:Initializing launch training script
```

Can see the configs in the first log output and can see how its parsed as args. All of the above cases ran tuning successfully.

4.  Failure case -- not pass num_processes, do not pass `'accelerate_launch_args'`

Since I ran this on a pod with 2 GPUs, by not setting num_processes, `accelerate launch` automatically set num_processes=2 and distributed_type=MULTI_GPU. 

This failed because llama-13b model failed to run tuning, hitting into OOM. This is because FSDP needs to be used to load the model, not DPP.

Logs:
```sh
WARNING:root:num_processes param was not passed in. Value from config file (if available) will                 be used or accelerate launch will determine number of processes automatically
DEBUG:root:accelerate_launch_args: ['/app/launch_training.py']
DEBUG:root:accelerate launch parsed args: Namespace(config_file=None, quiet=None, cpu=False, multi_gpu=False, tpu=False, ipex=False, mixed_precision=None, num_processes=None, num_machines=None, num_cpu_threads_per_process=None, dynamo_backend=None, dynamo_mode='default', dynamo_use_fullgraph=False, dynamo_use_dynamic=False, use_deepspeed=False, use_fsdp=False, use_megatron_lm=False, use_xpu=False, gpu_ids=None, same_network=False, machine_rank=None, main_process_ip=None, main_process_port=None, tee='0', role='default', rdzv_backend='static', rdzv_conf='', max_restarts=0, monitor_interval=5, module=None, no_python=None, tpu_use_cluster=False, tpu_use_sudo=False, vm=None, env=None, main_training_function=None, downcast_bf16=False, deepspeed_config_file=None, zero_stage=None, offload_optimizer_device=None, offload_param_device=None, offload_optimizer_nvme_path=None, offload_param_nvme_path=None, gradient_accumulation_steps=None, gradient_clipping=None, zero3_init_flag=None, zero3_save_16bit_model=None, deepspeed_hostfile=None, deepspeed_exclusion_filter=None, deepspeed_inclusion_filter=None, deepspeed_multinode_launcher=None, fsdp_offload_params='false', fsdp_min_num_params=100000000.0, fsdp_sharding_strategy='FULL_SHARD', fsdp_auto_wrap_policy=None, fsdp_transformer_layer_cls_to_wrap=None, fsdp_backward_prefetch_policy=None, fsdp_backward_prefetch=None, fsdp_state_dict_type=None, fsdp_forward_prefetch='false', fsdp_use_orig_params='true', fsdp_cpu_ram_efficient_loading='true', fsdp_sync_module_states='true', megatron_lm_tp_degree=1, megatron_lm_pp_degree=1, megatron_lm_num_micro_batches=None, megatron_lm_sequence_parallelism=None, megatron_lm_recompute_activations=None, megatron_lm_use_distributed_optimizer=None, megatron_lm_gradient_clipping=1.0, aws_access_key_id=None, aws_secret_access_key=None, debug=None, training_script='/app/launch_training.py', mpirun_hostfile=None, mpirun_ccl=1, training_script_args=[])
WARNING:accelerate.commands.launch:The following values were not passed to `accelerate launch` and had defaults used instead:
	`--num_processes` was set to a value of `2`
		More than one GPU was found, enabling multi-GPU training.
		If this was unintended please pass in `--num_processes=1`.
	`--num_machines` was set to a value of `1`
	`--mixed_precision` was set to a value of `'no'`
	`--dynamo_backend` was set to a value of `'no'`
To avoid this warning pass in values for each of the problematic parameters or run `accelerate config`.
INFO:root:Using nproc_per_node=2.
INFO:root:Initializing launch training script
INFO:root:Initializing launch training script
...
...
Loading checkpoint shards:   0%|          | 0/3 [00:00<?, ?it/s]You are attempting to use Flash Attention 2.0 with a model not initialized on GPU. Make sure to move the model to GPU after initializing it on CPU with `model.to('cuda')`.
Loading checkpoint shards: 100%|██████████| 3/3 [04:35<00:00, 91.71s/it] 
Loading checkpoint shards: 100%|██████████| 3/3 [04:35<00:00, 91.88s/it] 
INFO:sft_trainer:Max sequence length is 4096
INFO:sft_trainer:Max sequence length is 4096
Map:   0%|          | 0/50 [00:00<?, ? examples/s]DEBUG:fsspec.local:open file: /.cache/huggingface/datasets/json/default-1938351542cb9295/0.0.0/ab573428e7a11a7e23eebd41a2a71665ac3789ce311cbad7049572034a9bda05/tmpawgphxxq
Map: 100%|██████████| 50/50 [00:00<00:00, 6481.29 examples/s]
INFO:sft_trainer:Training dataset length is 50
INFO:sft_trainer:Packing is set to False
Map:   0%|          | 0/50 [00:00<?, ? examples/s]DEBUG:fsspec.local:open file: /.cache/huggingface/datasets/json/default-1938351542cb9295/0.0.0/ab573428e7a11a7e23eebd41a2a71665ac3789ce311cbad7049572034a9bda05/tmpjv138ss1
Map: 100%|██████████| 50/50 [00:00<00:00, 8143.01 examples/s]
...
...
  File "/app/launch_training.py", line 183, in <module>
    main()
  File "/app/launch_training.py", line 138, in main
    sft_trainer.train(model_args, data_args, training_args, tune_config)
  File "/usr/local/lib/python3.11/site-packages/tuning/sft_trainer.py", line 269, in train
    trainer.train()
  File "/usr/local/lib/python3.11/site-packages/trl/trainer/sft_trainer.py", line 360, in train
    output = super().train(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/transformers/trainer.py", line 1780, in train
    return inner_training_loop(
           ^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/transformers/trainer.py", line 2181, in _inner_training_loop
    self.optimizer.step()
  File "/usr/local/lib/python3.11/site-packages/accelerate/optimizer.py", line 149, in step
    self.optimizer.step(closure)
  File "/usr/local/lib64/python3.11/site-packages/torch/optim/lr_scheduler.py", line 75, in wrapper
    return wrapped(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib64/python3.11/site-packages/torch/optim/optimizer.py", line 385, in wrapper
    out = func(*args, **kwargs)
          ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib64/python3.11/site-packages/torch/optim/optimizer.py", line 76, in _use_grad
    ret = func(self, *args, **kwargs)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib64/python3.11/site-packages/torch/optim/adamw.py", line 176, in step
    has_complex = self._init_group(
                  ^^^^^^^^^^^^^^^^^
  File "/usr/local/lib64/python3.11/site-packages/torch/optim/adamw.py", line 127, in _init_group
    state["exp_avg_sq"] = torch.zeros_like(
                          ^^^^^^^^^^^^^^^^^
torch.cuda.OutOfMemoryError: CUDA out of memory. Tried to allocate 136.00 MiB. GPU 0 has a total capacity of 79.15 GiB of which 104.38 MiB is free. Process 1637710 has 79.05 GiB memory in use. Of the allocated memory 75.95 GiB is allocated by PyTorch, and 754.45 MiB is reserved by PyTorch but unallocated. If reserved but unallocated memory is large try setting PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True to avoid fragmentation.  See documentation for Memory Management  (https://pytorch.org/docs/stable/notes/cuda.html#environment-variables)
...
[2024-04-01 19:34:27,724] torch.distributed.elastic.multiprocessing.api: [ERROR] failed (exitcode: 1) local_rank: 0 (pid: 72) of binary: /bin/python
  File "/app/accelerate_launch.py", line 112, in <module>
    main()
  File "/app/accelerate_launch.py", line 108, in main
    launch_command(args)
  File "/usr/local/lib/python3.11/site-packages/accelerate/commands/launch.py", line 1048, in launch_command
    multi_gpu_launcher(args)
  File "/usr/local/lib/python3.11/site-packages/accelerate/commands/launch.py", line 702, in multi_gpu_launcher
    distrib_run.run(args)
  File "/usr/local/lib64/python3.11/site-packages/torch/distributed/run.py", line 803, in run
    elastic_launch(
  File "/usr/local/lib64/python3.11/site-packages/torch/distributed/launcher/api.py", line 135, in __call__
    return launch_agent(self._config, self._entrypoint, list(args))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib64/python3.11/site-packages/torch/distributed/launcher/api.py", line 268, in launch_agent
    raise ChildFailedError(
torch.distributed.elastic.multiprocessing.errors.ChildFailedError: 
```

build/accelerate_launch.py
```diff
@@ -0,0 +1,112 @@
...
+                accelerate_launch_args.append(f"--{key}")
+            else:
+                accelerate_launch_args.append(f"--{key}")
+                # Only need to add key for params that aren't flags ie. --quiet
+                if actions_type_map.get(key) == "_StoreAction":
+                    accelerate_launch_args.append(str(val))
+
+    num_processes = accelerate_config.get("num_processes")
+    if num_processes:
+        if num_processes > 1 and not accelerate_config.get("config_file"):
```
Ssukriti, line 84: this check should be moved out of num_processes if loop , it has nothing to do with mumber of processes. Config file can be set even with one process. The warning you are seeing with single GPU training will go away. In our config we have set num_machines to 1 , which is default for single GPU anyway

build/accelerate_launch.py
```diff
@@ -0,0 +1,111 @@
...
+        json_configs = txt_to_obj(json_env_var)
+
+    parser = launch_command_parser()
+    # determine which flags are store true actions (don't require a value to be set)
+    actions_type_map = {
+        action.dest: type(action).__name__ for action in parser._actions
+    }
+
+    # parse accelerate_launch_args
+    accelerate_launch_args = []
```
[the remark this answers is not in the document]
anhuong: yup this can be done, but @tharapalanivel has a refactor of this already in her unit tests PR. So perhaps we can wait to add it in then

build/README.md
```diff
@@ -0,0 +1,166 @@
...
+        },
+        "multi_gpu": true,
```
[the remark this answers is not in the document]
anhuong: nope! good catch

Ssukriti (approved):

Ssukriti:
@fabianlim @kmehant  letting you know, we refactored PR a bit -
1. num_processes will be passed by user and will override what is config
2. if not passed, we will use whats in config
Just like how accelerate does it . We will ask BAM and product to set num_processes = num_gpus explicitly.

Now the script is simple and passes every argument to accelerate launch

kmehant (approved):
https://github.com/foundation-model-stack/fms-hf-tuning/pull/98#issuecomment-2030776744

@Ssukriti Thank you, I have no further review comments from my end.

Ssukriti (approved):

PR #100, opened 2024-03-28
fix: not passing PEFT argument should default to full parameter finetuning

### LINKED ISSUE #48: Not passing `--peft_method` should fallback to a method which is *not* parameter efficient like full model param tuning
Currently, not setting `--peft_method` defaults to prompt tuning. However, intuitively, not passing `--peft_method` means not to use any parameter efficient method like LoRA and prompt tuning. Therefore, its good to fallback to full parameter tuning rather prompt tuning.

kmehant:
### Description of the change

Currently, not setting `--peft_method` defaults to prompt tuning. However, intuitively, not passing --peft_method should mean not to use any parameter efficient method like LoRA and prompt tuning. Therefore, its good to fallback to full parameter tuning rather prompt tuning. PR adds this.

### Related issue number

Closes #48 

### Was the PR tested

Yes ✅

```
fmstuning ❯ python sft_trainer.py --output_dir ./output
# no peft method is given so it defaults to none (full parameter tuning)
Namespace(peft_method='none')
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,11 @@ pip install -e ".[flash-attn]"
 ```
 [FlashAttention](https://github.com/Dao-AILab/flash-attention) requires the [CUDA Toolit](https://developer.nvidia.com/cuda-toolkit) to be pre-installed.
 
+If you wish to use [aim](https://github.com/aimhubio/aim), then you need to install it:
+```
+pip install -e ".[aim]"
+```
+
 ## Data format
 The data format expectation is a single column text. The trainer is configured to expect a response template as a string. For example, if one wants to prepare the `alpaca` format data to feed into this trainer, it is quite easy and can be done with the following code.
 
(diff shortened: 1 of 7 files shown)


kmehant:
Please do not merge, I will add conditional importing of peft and aim dependencies based on the user options. suggestion from @fabianlim

Ssukriti (commented):
I am ok with the PR checking if peft and aim are installed and importing accordingly, but for them to be optional dependencies , they need to be specified here https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/pyproject.toml#L24
and removed from requirements. txt .
- please make aim optional
- I am on the fence about peft since it is neeeded for majority of capability - but whatever @ashokponkumar and @fabianlim decide is fine
- please update README accordingly - installation section should specify the optional deps 
- if you do decide to make peft optional, you ll have to ensure CI/CD installs it for unit tests. . Please do not bypass unit tests in CI/CD by skipping the tests - skipping if not installed for local runs is ok, in CI/CD please install as part of test requirements

fabianlim:
@Ssukriti I agree with your point that most users will not do `pip install -e . [peft]` and then try to run peft but have a failure. Since peft is not a huge package, i think lets just leave it as main dep.

ashokponkumar:
I agree, given ease of use is an important factor, we can leave peft as a mandatory dependency.

Ssukriti (approved):
Thank you!

PR #104, opened 2024-03-30
Setting default values in training job config

tharapalanivel:
### Description of the change

Setting the following defaults for training jobs, will get overwritten if user explicitly passes in a value

- `save_strategy` = `epoch`
- `logging_strategy` = `epoch`

### Was the PR tested


tuning/config/configs.py
```diff
@@ -34,8 +34,8 @@
 class ModelArguments:
     model_name_or_path: Optional[str] = field(default="facebook/opt-125m")
     use_flash_attn: bool = field(
-        default=True,
-        metadata={"help": "Use Flash attention v2 from transformers, default is True"},
+        default=False,
+        metadata={"help": "Use Flash attention v2 from transformers, default is False"},
     )
     torch_dtype: Optional[Union[torch.dtype, str]] = torch.bfloat16
```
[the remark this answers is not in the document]
Ssukriti: no it should still be bfloat16 since we are using A100s and it will work

tuning/utils/config_utils.py
```diff
@@ -87,3 +89,55 @@ def get_hf_peft_config(task_type, tuning_config):
         hf_peft_config = None  # full parameter tuning
 
     return hf_peft_config
+
+
+def post_process_job_config(job_config_dict):
```
Ssukriti, line 94: can we keep this function inside the build folder since its to do with KTO config only? these utils are actually for our tuning config structures https://github.com/foundation-model-stack/fms-hf-tuning/tree/main/tuning/config .

Things related to docker build and KTO, we can keep inside build folder as its not relevant to library users and contributors

tests/utils/test_config_utils.py
```diff
@@ -0,0 +1,78 @@
...
+import os
+
+# Third Party
+import pytest
+
+# Local
+from tuning.config.peft_config import LoraConfig, PromptTuningConfig
+from tuning.utils.config_utils import post_process_job_config
+
+HAPPY_PATH_DUMMY_CONFIG_PATH = os.path.join(
```
Ssukriti, line 27: these can also be moved to tests/build/test_job_config.py ?

Ssukriti (changes requested):
ok with code changes, suggestions around refactoring files and functions.

we can keep all build related code inside build folder and tests can be inside tests/build as well

tests/dummy_job_config.json
```diff
@@ -0,0 +1,34 @@
+{
```
Ssukriti, line 1: this file should also be moved into tests/build/dummy_job_config.json ?

tuning/config/configs.py
```diff
@@ -34,8 +34,8 @@
 class ModelArguments:
     model_name_or_path: Optional[str] = field(default="facebook/opt-125m")
     use_flash_attn: bool = field(
-        default=True,
-        metadata={"help": "Use Flash attention v2 from transformers, default is True"},
+        default=False,
```
Ssukriti, line 37: for library users would have been good to leave it as True, we need the system testing team to keep it True. 

can we set it to False temporarily in launch_training script instead until we figure out what the issue was?

build/utils.py
```diff
@@ -0,0 +1,74 @@
...
+    parser = transformers.HfArgumentParser(
+        dataclass_types=(
+            configs.ModelArguments,
+            configs.DataArguments,
+            configs.TrainingArguments,
+            peft_config.LoraConfig,
+            peft_config.PromptTuningConfig,
+        )
+    )
+    peft_method_parsed = "pt"
```
Ssukriti, line 45: hmm do we need to set this here to pt? I know it was done earlier in launch training too, but now I am confused why.. @anhuong ?
we are reading it below - peft_method_parsed = job_config_dict.get("peft_method")

and if not set, it should remain None for full fine tuning

tests/build/test_utils.py
```diff
@@ -0,0 +1,80 @@
...
+
+# Third Party
+import pytest
+
+# Local
+from tuning.config.peft_config import LoraConfig, PromptTuningConfig
+from build.utils import process_launch_training_args
+
+HAPPY_PATH_DUMMY_CONFIG_PATH = os.path.join(
+    os.path.dirname(__file__), "..", "dummy_job_config.json"
```
Ssukriti, line 28: if we move the file as requested below the path will change

Ssukriti (commented):
some questions - also @anhuong have you reviewed the refactor to launch_training as well?

tuning/config/configs.py
Ssukriti: thanks

Ssukriti: you are yet to commit right?

tests/dummy_job_config.json
Ssukriti: is there a commit not showing up?

Ssukriti (dismissed):
approved from my end, just a remaining question on default for peft_method^. - I want to ensure we are testing fine tuning correctly, peft_method should be None for fine tuning.
1.  Does peft_method need to be passed explicitly as None then for fine tuning? 
2. or should we keep default to None, and pass "pt"/"Lora" explictly for prompt tuning/LoRA? 
I think second one makes more sense. 

When @anhuong approves, we can merge

build/utils.py
anhuong: I agree that this doensn't need to be set here. I think we used to set the peft_method to be PT by default but I know a recent PR went in to update that. In addition, as you said since we pull the value from the JSON, this gets updated to `None` appropriately. I verified in my tests that I ran which printed out the params parsed and passed into sft_trainer, that `tune_config` gets set to None appropriately

build/launch_training.py
```diff
@@ -67,69 +66,31 @@ def main():
...
-        ) = parser.parse_json_file(json_path, allow_extra_keys=True)
-
-        contents = ""
         with open(json_path, "r", encoding="utf-8") as f:
-            contents = json.load(f)
-        peft_method_parsed = contents.get("peft_method")
-        logging.debug("Input params parsed: %s", contents)
+            job_config_dict = json.load(f)
     elif json_env_var:
         job_config_dict = txt_to_obj(json_env_var)
```
anhuong, line 78: Note to self: perhaps we should refactor this into build/utils as well since accelerate_launch and launch_training use the same method to parse the JSON

anhuong (dismissed):
LGTM!

anhuong (approved):

PR #106, opened 2024-04-02
docs: add note on ephemeral storage

anhuong:
Provide more specific details and better kube use case of tuning LLM with sft_trainer.


build/README.md
```diff
@@ -163,3 +167,5 @@ volumes:
     configMap:
         name: sft-trainer-config
 ```
+
+The above kube resource values are not hard-defined. However, they are useful when running a LLaMa-13b model. If ephemeral storage is not defined, you will likely hit into error `The node was low on resource: ephemeral-storage. Container was using 1498072868Ki, which exceeds its request of 0.` where the pod runs low on storage while tuning the model.
```
Ssukriti, line 171:
```suggestion
The above kube resource values are not hard-defined. However, they are useful when running some models (such as LLaMa-13b model). If ephemeral storage is not defined, you will likely hit into error `The node was low on resource: ephemeral-storage. Container was using 1498072868Ki, which exceeds its request of 0.` where the pod runs low on storage while tuning the model.
```

Ssukriti (dismissed):
minor suggestion

Ssukriti (approved):

PR #115, opened 2024-04-10
Evaluation Data Format Scripts

alex-jw-brooks:
### Description of the change

This PR adds scripts for:

1. Pulling and formatting tone/entities/TSA data in Alpaca format
2. Converting the Alpaca format data to SFT format - this can use the "verbose" instruction template, or the (default) "simple" template.

It also some docs for how to run them.

### Was the PR tested


scripts/cluster_k8s.yaml
```diff
@@ -0,0 +1,73 @@
+apiVersion: v1
+kind: ConfigMap
```
anhuong, line 2: Probably don't have to commit this file

anhuong, line 2: Probably don't have to commit this file

scripts/cluster_stuff.md
```diff
@@ -0,0 +1,140 @@
+# Formatting / Tuning / Evaluating Notes
```
anhuong, line 1: Did you mean to add this as well to the PR? There is useful info here on how to run the scripts but there are also some parts, like running on a fork and specifying IBM datasets that may not be needed?

scripts/pull_and_format_datasets.py
```diff
@@ -0,0 +1,228 @@
...
+            Path to tone file to be formatted.
+        export_path: str
+            Path to export the formatted data to.
+        num_samples: Optional[int]
+            Number of samples to be included in the formatted file.
+    """
+
+    def get_entites_output_text(datum):
+        mentions = datum["mentions"]
+        # TODO: check this for TSA, but seems like it is the same as entities
```
anhuong, line 79: Can remove TODO since this is true

scripts/pull_and_format_datasets.py
```diff
@@ -0,0 +1,228 @@
+"""Pulls and formats data into alpaca format.
```
anhuong, line 1: Might want to specify it pulls specific datasets since this script isn't generic for any dataset

scripts/pull_and_format_datasets.py
```diff
@@ -0,0 +1,228 @@
...
+            os.path.join(download_subdir, filename)
+            for filename in os.listdir(download_subdir)
+        ]
+        for data_file in data_files:
+            # Apply this datasets formatter to the data file
+            export_path = os.path.join(
+                EXPORT_DIR, data_file[data_file.index(os.sep) + 1 :]
+            )
+            dataset_info[SUBSAMPLE_KEY]
+            # This is silly :)
```
anhuong, line 211: haha like the note

anhuong (changes requested):
Few comments on some additional cleanup, but have been using the scripts and they work well! Making the pull_and_format script generic is a nice future todo

scripts/cluster_stuff.md
alex-jw-brooks: I _did_ mean to commit this, as otherwise it's a bit unclear what the role of each script actually is, and these scripts are only really applicable to IBM datasets anyway, since the format that they're consuming is internal formats that aren't standardized (i.e., if they were standardized, we wouldn't probably would not need this script to begin with).

I am open to changing the guidance here if you have specific thoughts though! Or, we could reduce this by a bit by not including any information on how to run in a cluster / k8s, but IMO that is relatively useful knowledge, at least within our group

alex-jw-brooks:
Thanks @anhuong, I'll make the changes! I am curious what your thoughts are in terms of making the dataset pulling / formatting generic - since the intent is really to format the dataset _into_ something that is generic (i.e., alpaca, which can then be converted to SFT), the lambdas to do the actual conversion are pretty dataset / task specific. The script itself could be expanded, e.g., to make the lambdas/COS paths pickable through a CLI or something like that, but given that the input format of what is being consumed is arbitrary, the main thing that could potentially be standardized is output formats for alike tasks, which isn't really worth it IMO

anhuong:
In terms of making the dataset pulling / formatting generic, I feel like this would be nice but since it's so specific to the dataset on how to parse and reformat it, I agree that it's not worth it and this is a nice reference that someone can use to help format their datasets as needed.

scripts/cluster_stuff.md
alex-jw-brooks: Cool, thanks for the links! I updated the docs to point it at the wiki for some guidance on the cluster stuff & deleted the k8s config, but left some small notes in about how you might go about running things in a pod, and also added a short section on how to add new datasets

scripts/pull_and_format_datasets.py
alex-jw-brooks: Sure sounds good!

scripts/cluster_k8s.yaml
alex-jw-brooks: Removed it and pointed it at the wiki for how to run something in the cluster

anhuong (approved):
Awesome, thanks Alex!

PR #119, opened 2024-04-12
Propose ADR for Training Acceleration

fabianlim:
### Description of the change

This ADR proposes an framework to integrate training **accelerations** into `tuning.sft_trainer.py`. 

Recently, it has been observed that new training techniques are released with an incomplete "preview" version. These "preview" versions tend to be not be fully integrated into OSS. Therefore, using new techniques typically involve additional work. This framework aims to allow timely integrations of such techniques into `sft_trainer.py`. A short list of powerful training techniques but are "preview"-only include:
- Huggingface integration of [AutoGPTQ](https://github.com/AutoGPTQ/AutoGPTQ).
    * 4-bit quantization kernels to reduce memory storage of the base weights.
- [Unsloth](https://github.com/unslothai/unsloth).
    * Fused operation kernels
    * Kernels for common model architectures (e.g., cross-entropy losses, RoPE embeddings and RMS norms).
- [megablocks](https://github.com/databricks/megablocks).
    * acceleration package for distributing mixture-of-experts that improves upon FSDP sharding.

With these tools, users will be able to train models with less GPU resources and/or quicker, resulting in quicker turnaround and improved user experience. 

### Related issue number

NA

### How to verify the PR

NA

### Was the PR tested


fabianlim:
Closing the PR for now as this may not be the appropriate time to raise

architecture_records/002-acceleration-framework.md
```diff
@@ -0,0 +1,302 @@
...
+The proposal satisfies the following desiredata:
+- Unified configuration YAML for all plugins. Fine configuration details abstracted away from Maintainers and other plugin developers.
+- Modular design allows new methods plugins to be added / removed / deactivated seemlessly.
+- Modular design enforces that plugins interact with `sft_trainer.py` at controlled points, and throw appropriate exceptions.
+- Generic enough for most use cases of interest (e.g., quantization, distributed training, etc).
+- Unobstrusive design that only *modifies the model*, and leaves [`SFTTrainer`] unmodified. Minimal inversion-of-control maintained through [`TrainerCallbacks`].
+
+### Only the Model is Modified
+
+Since the [`SFTTrainer`] is designed to work with generic pytorch models, modifying the model is much less intrusive to the training pipeline, then say, modifying [`SFTTrainer`] itself. The hope is then if we constrain ourselves to modify only the model, that we can implement all the method plugins (e.g., quantization, distributed training, etc) that we hope for.
```
nairbv, line 78: > Since the [`SFTTrainer`] is designed to work with generic pytorch models

is this true? it inherits from transformers.Trainer and I see a lot of HF-specific code in Trainer.train().

I know the PEFT code used in SFTTrainer doesn't work nicely with non-HF models

architecture_records/002-acceleration-framework.md
```diff
@@ -0,0 +1,302 @@
+# Training Enhancements Framework
```
nairbv, line 1: as mentioned in another PR, it might be better to track requirements/proposals in github issues as we've done previously in these fms-related repos.

architecture_records/002-acceleration-framework.md
```diff
@@ -0,0 +1,302 @@
...
+- [`SFTTrainer`] also performs model augmentation internally (e.g., it installs PEFT adapters if `peft_config` is passed in). 
+    * However, [`SFTTrainer`]'s model augmentation should be passed through if configs are omitted (e.g., if `peft_config = None`).
+- [`SFTTrainer`] will prepare model for distributed training (e.g. wrap with `FSDP`) internally. 
+    * thus Plugin implementers need to be aware that `TuningAccelerationPlugin.augmentation` should not interfere with any model preperation that [`SFTTrainer`] will perform.
+
+![Framework](imgs/002-framework.png)
+
+### Acceleration Methods 
+
+A top priority is to incorporate methods that enchance PEFT. While PEFT is known to be memory efficient, it is known to be slower than full-finetuning if not *properly optimized*. Also, another topic of interest is to add support for 4D masks to enable packing while instruction tuning; this acceleration may require some adjustments to the data processing.
```
nairbv, line 181: might help to clarify what we mean by "slower." slower to converge? slower tokens per GPU per second?

architecture_records/002-acceleration-framework.md
```diff
@@ -0,0 +1,302 @@
...
+
+Since the [`SFTTrainer`] is designed to work with generic pytorch models, modifying the model is much less intrusive to the training pipeline, then say, modifying [`SFTTrainer`] itself. The hope is then if we constrain ourselves to modify only the model, that we can implement all the method plugins (e.g., quantization, distributed training, etc) that we hope for. 
+
+The framework is designed to only modify them model at two integration points in `sft_trainer.py`. The primary motivation for this is easy code maintenance:
+1. an *optional* `loading` method that acts as a drop-in replacement for `AutoModel.from_pretrained`. 
+2. an *optional* `agumentation` method that provides a way to perform *minor* adjustments to an already instantiated model 
+3. an *optional* `callback` method to install `TrainerCallbacks` (if needed, e.g. custom save logic).
+
+```python
+class TuningAccelerationPlugin:
```
nairbv, line 86: Maybe it would be helpful to have an example, e.g., what would an implementation of this look like if Lora or PrefixTuning were implemented using this plugin architecture?

If I'm understanding correctly, I'd expect that `augmentation` would be used to do things like replacing Linear layers with LoraLinear layers, and `loading` would be used to load a Lora checkpoint (but why 'loading' instead of just 'load'?). I'm not sure where `callbacks` would be applied, and not sure how the checkpoint saving code would be modified.

architecture_records/002-acceleration-framework.md
```diff
@@ -0,0 +1,302 @@
...
+    model, (peft_config,) = framework.augmentation(
+        model, trainer.accelerator,
+        train_args, modifiable_args=(peft_config,),
+    )
+
+# instantiate trainer. Pass in model (with training enchancements)
+trainer = Trainer(model, ...)
+
+# Minor Change 4: add trainer callbacsk
+trainer.add_callbacks(framework.callbacks())
```
nairbv, line 164: isn't `Trainer` part of HF transformers? does it have an `add_callbacks` method and, if not, how do we add that?

architecture_records/002-acceleration-framework.md
fabianlim: Thanks for the comment! yes I have now attached an ICLR paper that has a very clear graphic that Full Finetuning actually requires the least tuning time across various PEFT methods.!

architecture_records/002-acceleration-framework.md
```diff
@@ -0,0 +1,302 @@
...
+
+### Minimal and Controlled Changes to Training Script
+
+All proposed code changes to `sft_trainer.py` contained in minimal lines of code:
+- Plugins loaded by discovery; transparent to `sft_trainer.py`.
+- Plugin configuration automatically parsed.
+- Passthrough to original operation if `Framework` is disabled.
+
+```python
+from tuning.acceleration import AccelerationFramework
```
Ssukriti, line 129: things I am not very clear on:
1. How does example above `TuningAccelerationPlugin:` relate here ?It is not even imported in this example, so is it not used here?  Can we get one complete example in ADR? What does `AccelerationFramework` contain? 

2. what happens when there are multiple frameworks all imported in main code? like what if theer is one for quantization and one for something else? at train time what flag will determine which framework to use? Here it seems very hard coded to AccelartionFramework `if framework_args.config_file is not None:
    framework = AccelerationFramework(framework_args.config_file)`
  
Will there not be more than one?

 3. I think I may not be understanding differenec between Framework and plugin in this ADR. From what I understood is plugin is the class that has custom loaders/augmentors . What is Framework? Is Framework the yaml file + Plugin? I see it has the flags `framework.requires_custom_loading:` , but it also has the loaders `framework.model_loader ` 
 
 4. did you mean `framework.loading()` or `framework.model_loader` in line 139 ? Should framework have a model_loader property?

architecture_records/002-acceleration-framework.md
fabianlim: Yes! the `Trainer` aluded to here indeed is the `tranformers.Trainer`, which does [indeed have an `add_callback` method](https://github.com/huggingface/transformers/blob/e34da3ee3c9d2d628fdbeb60cee45c4f8f32945a/src/transformers/trainer.py#L718) (sorry there was a minor typo).

And as you know `trl.SFTTrainer` inherits from `tranformers.Trainer`,

architecture_records/002-acceleration-framework.md
fabianlim: Yes your comments on `augmentation` and `loading` are spot on. Yes we could consider to change to `load` (however in the revision I changed to `model_loader`, but that could be further changed.

An example of custom model saving logic could be something like as follows. Back in the days when PEFT was new, the HFTrainer would not know how to detect that it is a PEFT-Model, and save the full model weights. So 
we used to have some logic to delete and replace with the adapter weights only. 

In the examples I have now, this is not the case, but I included it in the design as it could be something that is needed in the future. 

```
class SavePeftModelCallback(TrainerCallback):
    def on_save(
        self,
        args: TrainingArguments,
        state: TrainerState,
        control: TrainerControl,
        **kwargs,
    ):
        checkpoint_folder = os.path.join(args.output_dir, f"{PREFIX_CHECKPOINT_DIR}-{state.global_step}")

        peft_model_path = os.path.join(checkpoint_folder, "adapter_model")
        kwargs["model"].save_pretrained(peft_model_path)

        pytorch_model_path = os.path.join(checkpoint_folder, "pytorch_model.bin")
        if os.path.exists(pytorch_model_path):
            os.remove(pytorch_model_path)
        return control
```

architecture_records/002-acceleration-framework.md
fabianlim: Ah ok, Will be happy to comply if you can send me an example of such a proposal @Ssukriti .

architecture_records/002-acceleration-framework.md
fabianlim: @Ssukriti thanks for your careful reading. Have attempted to address in https://github.com/foundation-model-stack/fms-hf-tuning/pull/119/commits/da8c9f8905cd50bcf9d59b8590f39b86cb5c5bdf

Just take note in the rewrite there are renames. I shall refer to them with the new names
- `TuningAccelerationPlugin` -> `AccelerationPlugin`
- `Framework` -> `AccelerationFramework`

> How does example above TuningAccelerationPlugin: relate here ?It is not even imported in this example, so is it not used here? Can we get one complete example in ADR? What does AccelerationFramework contain?

I take it that you mean `AccelerationPlugin` was not imported `sft_trainer.py` example. Yes it was omission on my part. I have now included:
- (almost) full implementation of `AccelerationFramework`. See new section **Implmentation of Acceleration Framework Class**. This is the thing that is imported in `sft_trainer.py` and it will construct and manage all the `AccelerationPlugin`. This management is *transparent* to `sft_trainer.py`, that is why it does not appear. 
- Also in **AccelerationPlugin Base Class**, I have further fleshed out `AccelerationPlugin`, and show that all implemented plugins need to be registered to `PLUGIN_REGISTRATIONS`. That is how `AccelerationFramework` will know which plugins are registered and available for loading.
- In section **Detailed Design**.  this shows a concrete implemntation of `AutoGPTQAccelerationPlugin` that inherits from `AccelerationPlugin`. And the end of the class definition, you can see we make a `register_plugin` call to register the plugin. When registering, we leave a bread crumb to inform `AccelerationFramework` that if you see this key in the configuration file, then promote the plugin to active status.
    ```
    configuration_paths=["peft.quantization.auto_gptq"], 
    ```
- In the following subsection *Configuration To Load AutoGPTQ LoRA Plugin**, I show an example of a configuraiton file used to activate `AutoGPTQAccelerationPlugin`

fabianlim: > what happens when there are multiple frameworks all imported in main code? like what if theer is one for quantization and one for something else? at train time what flag will determine which framework to use? Here it seems very hard coded to AccelartionFramework if framework_args.config_file is not None: framework = AccelerationFramework(framework_args.config_file)

@Ssukriti No I intend `AccelartionFramework` to be a *singleton*. On referring to the newly updated `sft_trainer.py`, `AccelartionFramework` will be created only if `AccelerationFrameworkArguments.acceleration_framework_config_file` is specified.
- plugins are transparent to user. The `AccelartionFramework` hides logic for managing and loading plugins to create a simple interfacing point.
- Inside that config file, we could have multiple plugin descriptions, that loads multiple plugins. I give another example below, to show something we have in our `dev` code to that will load 2 plugins, the second plugin installs fast kernels on top of the first plugin that loads an auto-gptq (LoRA) trainable model.
  ```yaml
  plugins:
  
    # PEFT-related acceleration
    peft:
  
      # quantization-releated acceleration
      # e.g., kernels for quantized base weights
      quantization: 
  
        # AutoGPTQ quantized base weights.
        auto_gptq:
          kernel: triton_v2
          from_quantized: True
  
        unsloth:
          stackable: 
            base_layer: auto_gptq # bitsandbytes
            fused_lora: True
            fast_loss: True
            fast_rsm_layernorm: True
            fast_rope_embeddings: True
  ```

fabianlim: > I think I may not be understanding differenec between Framework and plugin in this ADR. From what I understood is plugin is the class that has custom loaders/augmentors . What is Framework? Is Framework the yaml file + Plugin? I see it has the flags framework.requires_custom_loading: , but it also has the loaders framework.model_loader 

Sorry again for the confusion, the user only interfaces with `AccelerationFramework`. The `AccelerationPlugin` also has `model_loader` and `augmentation` functinos, but we rely on the `AccelerationFramework` to call them; the user does not call them directly

I hope that the full implementation of `AccelerationFramework` provided under the new section **Implmentation of Acceleration Framework Class** can help clarify what im saying here

fabianlim: > did you mean framework.loading() or framework.model_loader in line 139 ? Should framework have a model_loader property?

Oh, i did mean it as is. If i reprise the code here:

```
# Minor Change 2: custom loader (if necessary)
_model_loader = AutoModelForCausalLM.from_pretrained # default
if framework is not None and framework.requires_custom_loading:
    _model_loader = framework.model_loader # drop in replacement

# will passthrough the default loader if framework is disabled
model = _model_loader(
    attn_implementation="flash_attention_2" if model_args.use_flash_attn else None,
)
```

- if `framework is not None and framework.requires_custom_loading` is False, then 
  ```
  _model_loader = AutoModelForCausalLM.from_pretrained 
  ```
  which means that on the line `model = _model_loader(...)` then `model` will be loaded by `AutoModelForCausalLM.from_pretrained`.
- similarly otherwise then `model` will be loaded by `framework.model_loader `, which is a bona-fide function, and internally has logic to call the plugin's model_loader to load the model. See the full implementation of `AccelerationFramework` that is now provided in https://github.com/foundation-model-stack/fms-hf-tuning/pull/119#:~:text=%E2%80%A6-,da8c9f8,-fabianlim%20force%2Dpushed

architecture_records/002-acceleration-framework.md
fabianlim: Thanks for pointing this out. Ah ok yes `Trainer` is more flexible than `SFTTrainer`; the `SFTTrainer` has some additional requirements (e.g., cannot pass `tokenizer=None`) but these are bare minimum.

I updated the text to be more careful. But you can verify that the below will actually train. This is a ridiculous example, but it does show how `SFTTrainer` can train a `torch.nn.Module`.

Furthermore, if `AccelerationFramework` loads a model, typically it will be one that resembles a HF model. 

```python
iimport numpy as np
import torch

N = 800 # number of examples
S = 10 # sequence length 

# create dataset 
from torch.utils.data.dataset import TensorDataset
rng = np.random.default_rng(1)
data = torch.tensor(rng.random((N, S, 5)), dtype=torch.float)
labels = torch.tensor(rng.random((N, S, 2)), dtype=torch.float)
dataset = TensorDataset(data, labels)

# create data collator
def data_collator(features):
    # import pdb; pdb.set_trace()
    x, y = zip(*features)
    return {'x': torch.stack(x), 'y': torch.stack(y)}

# create model
class Model(torch.nn.Module):
    def __init__(self):
        super().__init__()
        self.weight = torch.nn.Linear(5, 2)
        self.loss = torch.nn.MSELoss()
        
    def forward(self, x, y):
        return (self.loss(self.weight(x), y), )

model = Model()

# trainer
from transformers import Trainer
from trl import SFTTrainer
from transformers import TrainingArguments
args = TrainingArguments('results', logging_steps=50, save_strategy='steps', save_steps=50)

# The dataset in this example is overly simple, so just pardon some stubbing of the tokenizer and the
# formatting func
class DummyTokenizer:
    padding_side = None
    save_pretrained = lambda x,y: x
    model_max_length = S

formatting_func = lambda x: x

trainer = SFTTrainer(
    model, args=args, tokenizer=DummyTokenizer(), 
    train_dataset=dataset, data_collator=data_collator,
    formatting_func=lambda x:x,
)
```

architecture_records/002-acceleration-framework.md
nairbv: I don't think we have a fixed template, just regular github issues, but pasting the text of an ADR into a github issue would be fine I think

architecture_records/002-acceleration-framework.md
nairbv: hmm...  Currently PEFT is handled by PEFT-specific code inside SFTTrainer (https://github.com/huggingface/trl/blob/v0.8.6/trl/trainer/sft_trainer.py), including code in the superclass (https://github.com/huggingface/transformers/blob/v4.40.1/src/transformers/trainer.py#L2500) that's specific to how adapters are checkpointed. Is none of that code needed and something like this can be implemented entirely external to the trainer class?

If I go look at the docs for callbacks (https://huggingface.co/docs/transformers/en/main_classes/callback) it says `Callbacks are “read only” pieces of code`, but I guess "read only" only means from the perspective of modifying the behavior of the training loop, so I guess writing checkpoints is still within the idea of "read only?"

architecture_records/002-acceleration-framework.md
```diff
@@ -0,0 +1,480 @@
...
+
+Describe the forces at play, including technological, political, social, and project local. These forces are likely in tension, and should be called out as such. The language in this section is value-neutral. It is simply describing facts.
+-->
+
+### Motivation
+
+Currently `sft_trainer.py` only can access those tools already integrated in HF. Due to rapid developments in AI training methodologies, these technologies are quickly considered "standard", for example:
+1. LoRA adapters from [PEFT](https://github.com/huggingface/peft).
+2. Prefix tuning from [PEFT](https://github.com/huggingface/peft).
+3. FSDP training from [accelerate](https://github.com/huggingface/accelerate).
```
nairbv, line 32: These are examples of features that _are_ part of HF training. I'm not clear if the intent is to say that these are features that could have been implemented outside HF and still used in HF trainer? If so I think it should be a bit more explicit

architecture_records/002-acceleration-framework.md
```diff
@@ -0,0 +1,480 @@
...
+
+Below are various reasons for a framework to integrate custom training tools into `sft_trainer.py`. 
+* Enable quick integrations of open-source techniques that have yet to be integrated into Huggingface.
+* Enable integrations of custom techniques developed by IBM researchers, that are not planned be integrated into Huggingface.
```
nairbv, line 36: I think more specific example use-cases might help clarify the motivation here, or at least the nature of those use cases. What are some of the open source techniques this could be used to implement?

And conversely what are examples of use-cases that would not be supported?

I think anything that modifies the logic of the training loop would not be supported (In the docs for Callbacks it says "For customizations that require changes in the training loop, you should subclass [Trainer](https://huggingface.co/docs/transformers/v4.40.1/en/main_classes/trainer#transformers.Trainer) and override the methods you need")

... and so I think the idea is this is all about modifying just the model, and then callbacks would just be for alternate checkpointing mechanisms?

raghukiran1224:
@fabianlim I have a meta question here, if we add `unsloth` kernels -- do we do that by creating a dependency on `unsloth`?

architecture_records/002-acceleration-framework.md
fabianlim: Understand, will see what @Ssukriti thinks. 

By rendering issues, do you mean some broken links, as well as inconsistencies between inline code (e.g., `SFTTrainer` vs `[SFTTrainer]`)?

architecture_records/002-acceleration-framework.md
fabianlim: > If I go look at the docs for callbacks (https://huggingface.co/docs/transformers/en/main_classes/callback) it says Callbacks are “read only” pieces of code, but I guess "read only" only means from the perspective of modifying the behavior of the training loop, so I guess writing checkpoints is still within the idea of "read only?"

Correct, I believe because "read-only" is in quotes, it is not to be taken literally. For example
- `class transformers.integrations.TensorBoardCallback` may write to some location to populate data for tensorboard
- [`PrinterCallback`](https://github.com/huggingface/transformers/blob/9fe3f585bb4ea29f209dc705d269fbe292e1128f/src/transformers/trainer_callback.py#L220-L224)  modifies the `log` object.

And the fact that `TrainerCallbacks` [provide an `on_save` event](https://github.com/huggingface/transformers/blob/9fe3f585bb4ea29f209dc705d269fbe292e1128f/src/transformers/trainer_callback.py#L289-L293) suggests it is meant to allow custom save logic to be implemented, hence the example above.

> hmm... Currently PEFT is handled by PEFT-specific code inside SFTTrainer (https://github.com/huggingface/trl/blob/v0.8.6/trl/trainer/sft_trainer.py), including code in the superclass (https://github.com/huggingface/transformers/blob/v4.40.1/src/transformers/trainer.py#L2500) that's specific to how adapters are checkpointed. Is none of that code needed and something like this can be implemented entirely external to the trainer class?

If you move down a couple of lines you would see [that this loading logic works only if the adapters are saved properly](https://github.com/llohann-speranca/transformers/blob/5a35cb9c12724f97b3866edd260cf1f27613f026/src/transformers/trainer.py#L2077-L2081), and suggest that in order to save properly, we should use a [custom callback](https://github.com/huggingface/peft/issues/96). The loading logic was fixed around mid 2023, and only on Jan this year, that they introduced a [`save_adapter` into `save_fsdp_model` in accelerate](https://github.com/huggingface/transformers/pull/28297). So from Jan onwards, this callback is no more needed, but within those 6 months, you would have not been able to resume your trainings without a proper callback.

This is a situtation that no longer exists, but within the context of `tuning.sft_trainer.py`, I believe we need an organized way of putting in such fixes without always changing the `sft_trainer.py`. Hence this Framework is envisioned to do that. One year ago PEFT was envisioned as the method to train cheaply, but it wouldnt have worked well without having properly saved checkpoints.

architecture_records/002-acceleration-framework.md
fabianlim: Sorry for the confusion! The intent was not to say that innovations come very quickly, and while these 3 already have some form of integration in HF, and thus are readily accessible, there are others that are currently not. I have listed those  a couple of lines below. 

> Recently, it has been observed that new training techniques are released with an incomplete "preview" version. These "preview" versions tend to be not be fully integrated into OSS. Therefore, using new techniques typically involve additional work. This framework aims to allow timely integrations of such techniques into `sft_trainer.py`. A short list of powerful training techniques but are "preview"-only include:
> - Huggingface integration of [AutoGPTQ](https://github.com/AutoGPTQ/AutoGPTQ).
    * 4-bit quantization kernels to reduce memory storage of the base weights.
> - [Unsloth](https://github.com/unslothai/unsloth).
    * Fused operation kernels
    * Kernels for common model architectures (e.g., cross-entropy losses, RoPE embeddings and RMS norms).
> - [megablocks](https://github.com/databricks/megablocks).
    * acceleration package for distributing mixture-of-experts that improves upon FSDP sharding.

The concern of this ADR is then, how to get these advancements (any many more in the future) working with `tuning.sft_trainer.py` in a timely fashion, where we can integrate, fix issues and test, without having to make a PRs to constantly change , and most invariably bloat, `tuning.sft_trainer.py`. 

I believe having something like this will greatly increase the value of `tuning.sft_trainer.py`.

architecture_records/002-acceleration-framework.md
fabianlim: Following up on the previous comment, those 3 examples i have listed are quite canonical. Basically these would be use cases that 
- need to load the model differently (i.e., cannot use AutoModel.from_pretrained), like loading a quantized checkpoint, or some custom sharding strategy no supported OOTB in FSDP like in the megablocks use case.
- need to replace layers in ways not supported (e.g., like when there is a new triton kernel release not yet integrated into HF).
- need to modify some FSDP configuration. This is needed in megablocks, and [there is already one example of such thing being done currently in sft_trainer.py](https://github.com/fabianlim/fms-hf-tuning/blob/2c19b708accc5704c01c01efe0f62d1a4735752a/tuning/sft_trainer.py#L277-L280). 

> ... and so I think the idea is this is all about modifying just the model, and then callbacks would just be for alternate checkpointing mechanisms? Does model loading happen outside of the trainer class?

Callbacks may not be always used for checkpointing. If Im thinking ahead of the megablocks use case, then I plan to use `TrainerCallback.on_init_end` to modify the configuration of the `fsdp_plugin`, similar to [what is done here](https://github.com/fabianlim/fms-hf-tuning/blob/2c19b708accc5704c01c01efe0f62d1a4735752a/tuning/sft_trainer.py#L277-L280).

fabianlim:
> @fabianlim I have a meta question here, if we add unsloth kernels -- do we do that by creating a dependency on unsloth?

@raghukiran1224 thanks for the question. Few of these my thoughts regarding this outlined in the *Dependency* section of the ADR, but to clarify more.

1. That is the motivation of the plugin design. If the unsloth plugin is not desired, then it should not be installed, and any potential dependencies on unsloth are unwarranted. 
2. If the unsloth plugin is desired, then we need to load the code/kernels from somewhere. In this case, while it could be by having unsloth as an optional dep of `fms-hf-tuning`, Im not so keen as this will bloat `fms-hf-tuning`. Hence, my preference is to have `AccelerationFramework` manage the dep on behalf of `fms-hf-tuning`. My thinking is that the unsloth dep will be tied to the installation of the plugin, and probably the most preferred approach is to extract out parts of the unsloth that we need, in the implemntation of that plugin.

architecture_records/002-acceleration-framework.md
```diff
@@ -0,0 +1,480 @@
...
+#### Implmentation of Acceleration Framework Class
+
+The role of `AccelerationFramework` is to manage implemented plugins. In particular:
+- parse `configuration_file`, see below, and based on contents, decide the `AccelerationPlugin`'s to promote to `active_plugins`. See [implementaiton of AutoGPTQ LoRA Plugin](#detailed-design) for more description on configuration logic.
+- handle *plugin stacking*, i.e., when we have more than one `active_plugins`, apply their `model_loader` / `augmentation` logic in appropriate succession. 
+    * This is very useful, e.g., loading a LoRA-trainable AutoGPTQ model first, then applying additional optimized fused kernels to further improve training speeds.
+- enforce that `AccelerationFramework.model_loader` witll call `AccelerationPlugin.model_loader` of some `active_plugins` *at most once*.
+    * Prevents potential complication as multiple `active_plugins` could load models in conflicting ways.
+- enforce that `AccelerationFramework.augmentation` would apply `AccelerationPlugin.augmentation` in appropriate succession.
+    * c.f. previous example, where first a LoRA-trainable AutoGPTQ model is loaded, then fused_kernels on applied on top.
```
Ssukriti, line 155: so we dont want to limit number of augmentation techniques that can be applied? there can be multiple one after the other?

raghukiran1224:
> > @fabianlim I have a meta question here, if we add unsloth kernels -- do we do that by creating a dependency on unsloth?
> 
> @raghukiran1224 thanks for the question. Few of these my thoughts regarding this outlined in the _Dependency_ section of the ADR, but to clarify more.
> 
> In what follows below "unsloth plugin" means "the implementation of the unsloth version of`AccelerationPlugin`"
> 
> 1. That is the motivation of the plugin design. If the unsloth plugin is not desired, then it should not be installed, and any potential dependencies on unsloth are unwarranted.
> 2. If the unsloth plugin is desired, then we need to load the code/kernels from somewhere. In this case, while it could be by having unsloth as an optional dep of `fms-hf-tuning`, Im **not so keen** as this will bloat `fms-hf-tuning`. Hence, my preference is to have `AccelerationFramework` manage the dep on behalf of `fms-hf-tuning`. My thinking is that the unsloth dep will be tied to the installation of the plugin, and probably the most preferred approach is to extract out parts of the unsloth that we need, in the implemntation of that plugin.

The normal approach here is to copy the kernels that are most important into our repo based on the licenses and then use it on our own. We should not create any dependencies.

architecture_records/002-acceleration-framework.md
```diff
@@ -0,0 +1,480 @@
...
+
+    def model_loader(self, model_name: str, **kwargs):
+
+        if len(self.plugins_require_custom_loading) == 0:
+            raise NotImplementedError(
+                f"Attempted modeling loading, but none of activated plugins \'{list(self.active_plugins.keys())}\' "
+                "require custom loading."
+            )
+
+        # otherwise there should be exactly 1
```
Ssukriti, line 203: @fabianlim I undertsand the intent, but I feel this logic is making the code a bit complicated :
1. only one plugin can say it requires loading - as contributors add multiple plugins, to avoid errors they need to check other plugins to see if they have custom load implemented. If yes, then they cannot implement load. Error is thrown only at runtime, while creating plugin contributors wont even know the restriction.
2. they will have to go back and change plugin definition in code to fix error 

Instead of this , how about we do following:
1. allow plugins to define whatever they want in classes
2. at runtime, when user passes the config, restrict custom load to 1 by adding a field in yaml
```
plugins :

   peft
   ..... register all

allow_custom_load: plugin name
```

that way user can specify in config at runtime which plugin's load they want to use and the code can just read loader of that pluginClass. Only check will be : if specified plugin does not have a load(), throw warning and say default loader will be used

for augmentation and other callbacks, logic remains same

Ssukriti: but I guess the reason you might have done it this way is because if a plugib requires custom load, it doesnt make sense to just ignore the load() and apply augmentation?

the above design will only make sense if plugins can have their custom load functions defined, but dont absolutely require them - we can ignore load() and apply augmentation () . If this scenario is not true from use cases we know so far, then your design is fine and might be the only way :)

Ssukriti: hmm I guess you can have a unit test in the framework that can check all registered plugins and will fail if more than 1 plugin has requires_load property set 

That will prevent any contributions of multiple plugins that require custom load. Mostly I am double checking if this is what you really want to do

architecture_records/002-acceleration-framework.md
Ssukriti: @nairbv there are several advantages to writing an architecture decision record:
https://www.redhat.com/architect/architecture-decision-records

1. It clarifies to new contributors why a certain decision was made, helps larger community understand that . We will be regularly onboarding new members and finding multiple issues for each design made will be harder than keeping all docs in one place. 
2. At later points if we want to revisit why a certain decision was made, its easier to read a doc. The purpose of ADR is to clarify the architecture decision and why it was made , what were tradeoffs and why we selected this. 
3. its easier for us to review high level design first and propose changes as needed, before everyone wastes time creating code  and then we disagree with design. in issues we may not be able to capture so many details


As far as rending is considered, using local refernces in yml usually works . I did not find broken links in the md file.. were there any particular broken URLs? they can be fixed though

architecture_records/002-acceleration-framework.md
wynterl: @raghukiran1224 there is another issue with unsloth, I am discussing with an IP lawyer now on it, but although they use an apache2 license, they have some strange lines of text in 2 of their files asking user not to use those files on 'more than 4 gpu' or else send them money. Current plan is to take those 2 files, and rewrite them.

architecture_records/002-acceleration-framework.md
fabianlim: correct, we want to have a way to compose plugins. One concrerte use case i have is
- have one plugin to load an autoGPTQ prepared model for lora
- have an option to additional load a plugin that installs the lora merge ops, loss kernels, etc, on top of the model loaded by first plugin

That way we can have multiple options, and see how the different plugins / acceleration impact the speed. Also it has implications on needs, if user wants it simple, and does not need the additional speed (because dont want to install additional cuda kernels), can opt out of the second plugin

architecture_records/002-acceleration-framework.md
fabianlim: @Ssukriti 

> like if 2 plugins requested - peft and something else say XYZ , both require custom load, we cant use both together and the user has to pick only one?
we cant say - use load from peft, but use augmentation from XYZ without load from XYZ?

Yes correct. The current plan is to keep things simple, and if the user opts to chose 2 plugins that both require model, load, I will raise, and ask user to only pick one.

architecture_records/002-acceleration-framework.md
nairbv: @Ssukriti aren't those all purposes of github issues too though? I just think it's better to manage that doc through the github issues UI vs committing it to the repo. Github is a purpose built system with features for supporting a variety of development processes. Circumventing that system makes it harder to take advantage of those features.

It's great to have a template to use when creating docs. We can even add a template for github issue submission form so that if a user selects ADR it pre-populates in the UI (via an md file in .github/ISSUE_TEMPLATE). We could also add a label in the UI to differentiate these from bugs and other issues.

We also already track most design decisions and planned features through github issues, across many repos. There are currently 20 open and 25 closed issues just on this repo. Introducing a new process (that so-far only has 2-3 work items being tracked) bifurcates work tracking/planning. I think this would make it much harder to find information, since one would need to both search the UI and grep the directory. I think it would be better to copy these couple of ADRs into github issues and remove this directory to reduce confusion, rather than try to get everyone else to move or to maintain two divergent processes.

With committed files, I'm not sure how we even do things like track which are open/closed, or link to them from relevant PRs. We'd have to introduce a lot of conventions to adopt this process, conventions the rest of the OSS community isn't familiar with.

architecture_records/002-acceleration-framework.md
nairbv: > The intent was not to say that innovations come very quickly

"not" looks like a typo?

I think what I'm suggesting is just a slight tweak to the wording of the doc to be more explicit/clear about what you're saying this feature will/won't be able to do. (not for my sake, but for the sake of readers in general).

Since you don't say "these are techniques that could be / could have been implemented with this plugin framework" this can be interpreted as _some_ could have been ... or since you say "aims to" and "such techniques" can be interpreted as "some similar ones maybe, but not necessarily these specific ones at all and not really sure."

Also do we have examples of any yet-unimplemented features we would hope to use this for?

fabianlim: Hi @nairbv . Sorry for the typo in my [comment above](https://github.com/foundation-model-stack/fms-hf-tuning/pull/119#discussion_r1580240177). I have since corrected it. 

Also I have attempted to address your concern in my latest update.

architecture_records/002-acceleration-framework.md
Ssukriti: @nairbv we use issues to track work as you said. Example issue for creating this ADR https://github.com/foundation-model-stack/fms-hf-tuning/issues/144 , which is a work item. ADR can be of many forms - markdown, box notes, various other ways of documenting architecture decisions.

We are storing **design records** here for major architecture changes, so we can easily find them and go through design decisions and consequences . It also helps with easy onboarding. New members will not know what ADRs exist to be able to search long list of issues with hopes of finding some documentation in issue comments. 

Stpring ADRs as GitHub markdowns is not a new thing for open source community. Several examples exist and multiple projects are moving towards adopting ADRs
1. https://github.com/json-schema-org/json-schema-spec/tree/main/adr
2. there are several blogs and articles on benefits of ADR and where to store them, example: https://ctaverna.github.io/adr/ -> where they are stored is main;y wikipages or GitHub Mrakdowns

we have only small number of issues so far for thsi repo, as all of us have been using internal IBM trackers . we are trying to move everyone to using open source repo for issues. We have hundreds of issues for this project in internal tracker and no one is going to search each to find all relevant ADRs. 


Anyways, decision to support ADRs for this repo and contribution guidelines was discussed extensively and the need for it was raised by executives like Raghu. Hence, even if you disagree, please agree to disagree and accept this as a decision made by project owners and let us give it a chance. Thank you

architecture_records/002-acceleration-framework.md
```diff
@@ -0,0 +1,493 @@
...
+
+The framework is designed to only modify them model at two integration points in `sft_trainer.py`. The primary motivation for this is easy code maintenance:
+1. an *optional* `model_loader` method that acts as a drop-in replacement for `AutoModel.from_pretrained`. 
+2. an *optional* `agumentation` method that provides a way to perform *minor* adjustments to an already instantiated model 
+3. an *optional* `callback` method to install `TrainerCallbacks` (if needed, e.g. custom save logic).
+
+In what follows, we provide:
+- a description of an abstract base class that all plugins must inherit and conform to.
+- a brief description of the framework class that is responsible for managing plugins (e.g., loading, executing).
```
Ssukriti, line 100:
```suggestion
- a brief description of the framework class that is responsible for managing plugins (e.g., loading, executing).

NOTE: We want to note that the implementation of frameworks and plugins may be moved to a separate open source repository . fms-hf-tuning will then include the framework library as an optional dependency and call respective loaders and augmentation techniques as specified. 
```

Ssukriti (approved):
I  suggested one change @fabianlim to note that frameworks implementation may lie in another repository as we discussed, which will then be installed in fms-tuning . Otherwise fine with the ADR. 
We should also be aware in general ADRs are meant to capture major architecture decisions - here the primary point is that frameworks will be created in a way with custom loaders and augmentors , to avoid having to change SFT Trainer . Smaller implementation details may be adjusted as you start coding and create pull requests.

@nairbv thank you for the detailed review and thoughtful comments. Once your comments are resolved and we have an approval from you, we can merge ADR.

architecture_records/002-acceleration-framework.md
nairbv: @Ssukriti I think there was some confusion on my part re ownership of this repo/project and longer term plans for it. I'd like to avoid having ADRs (committed to) the fms and fms-extras repos but I don't still have a reason to object here.

fabianlim:
@Ssukriti after your approval i have rebased and fixed the DCO issue. 

@nairbv i have resolved the comments that you had given a "thumbs up" to. Your [most recent comment](https://github.com/foundation-model-stack/fms-hf-tuning/pull/119#discussion_r1588203988) seems to be addressed to @Ssukriti . I hope that I have addressed your concerns, and If I have could we go ahead to merge?

nairbv (approved):

PR #120, opened 2024-04-12
Propose ADR for Model Mixing

RhuiDih:
### Description of the change

This ADR proposes to add a feature to create MOE from base models.


mixing/mix.py
```diff
@@ -0,0 +1,285 @@
...
+):
+    
+    set_seed(1399)
+
+    logging.info("loading base model...")
+    config_base = AutoConfig.from_pretrained(base_model)
+    model_base = AutoModelForCausalLM.from_pretrained(base_model)
+    model_base_type = type(model_base)
+    sd_base = model_base.state_dict()
+    MOE_MODEL_CLS, MOE_CFG_CLS = MODELS_MOE_CLASS[model_base_type]
```
nairbv, line 70: local variables in python should be lowercase

mixing/models/mixllama/configuration_mixllama.py
```diff
@@ -0,0 +1,39 @@
+import transformers
+from transformers import LlamaConfig
+from transformers.utils import logging
+
+from packaging import version
+
+logger = logging.get_logger(__name__)
+
+class MixLlamaConfig(LlamaConfig):
```
nairbv, line 9: Is this the right repo (and path) to be defining these models architectures?

There's a (currently closed) PR to add the same code to fms here, and other PRs to add MOE models in FMS as well: https://github.com/foundation-model-stack/foundation-model-stack/pull/245/files

I think if we can get this into an FMS format it would be nice to have MOE models in fms. If it's going to be pure-HF it might also even make sense to add directly to HF.

tests/mixing/test_mix.py
```diff
@@ -0,0 +1,88 @@
+import logging, traceback
+import gc, shutil
+import torch
+
```
nairbv, line 4: pytest will attempt to run test_* files during CI, but looks like this is just being run as a script. Can this be reformulated as a pytest test?

mixing/models/mixllama/configuration_mixllama.py
wynterl: @nairbv This is quite different from training a MOE architecture. This involves mixing pre-trained models to simply *create* an MOE-type model. Then it can be minimally trained via freezing all but the routers (or even not trained) and inferenced out-of-the-box. So, it need not be clubbed with the MOE PR. As for the closed PR to the FMS repo, it is indeed the same, we are trying to find the right repo for this. 
We want to get this on BAM so the internal community can play with this first.

architecture_records/00x-model-merging-&-mixing.md
```diff
@@ -0,0 +1,50 @@
+# Model Merging & Mixing
```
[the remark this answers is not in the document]
wynterl: No problem on going straight to PR, we followed the convention on the fms-hf-tuning repo which suggests ADR then PR

tests/mixing/test_mix.py
RhuiDih: done

mixing/mix.py
RhuiDih: done

architecture_records/00x-model-merging-&-mixing.md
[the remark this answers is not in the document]
nairbv: if we're going to stick with github issues we might just want to move the ADR portion to a github issue

mixing/mix.py
nairbv: I still see a lot of upper-case local variables throughout the code

mixing/models/mixllama/modeling_mixllama.py
```diff
@@ -0,0 +1,462 @@
...
+from transformers.utils import logging
+
+from .configuration_mixllama import MixLlamaConfig
+from mixing.utils.loss import load_balancing as load_balancing_loss_func
+
+logger = logging.get_logger(__name__)
+
+_CONFIG_FOR_DOC = "MixLlamaConfig"
+
+def make_moe_modules(moe_class_partial):
```
nairbv, line 33: I don't think this extra level of indirection is required and using partials makes the code more confusing

architecture_records/00x-model-merging-&-mixing.md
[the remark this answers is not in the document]
wynterl: We will be proposing to integrate this PR in a different way. Stay tuned

mixing/mix.py
RhuiDih: last commit is not reflected in this comment section

PR #125, opened 2024-04-16
Fix eos token suffix removal

alex-jw-brooks:
### Description of the change

Fixes a bug introduced by https://github.com/foundation-model-stack/fms-hf-tuning/pull/121 - this should be repeatedly removing the suffix and not (character level) rstripping, nice catch @anhuong 

### Was the PR tested


anhuong (approved):
Thanks Alex!

Issue #129, opened 2024-04-23
feat: standardize the format for metrics, operations, controls in the yamls used by `TrainerControllerCallback`


HarikrishnanBalagopal:
## Is your feature request related to a problem? Please describe.

Current format
```yaml
controller-metrics:
    loss:
        Loss:
  - name: sparsity
    class: Sparsity
    args:
       key3: val3
       key4: val4
```

## Describe the solution you'd like

New format

```yaml
controller-metrics:
  - name: loss-one
    class: Loss
    arguments:
       key-num-1: val1
       key-num-2: val2
  - name: sparsity-one
    class: Sparsity
    arguments:
       key-num-3: val3
       key-num-4: val4
operations:
  - name: op1
    class: MyOperation1
controllers:
  - name: loss-controller
    triggers:
      - on_log
    rule: loss-one < 1.0
    operations:
      - hfcontrols.should_training_stop
```

PR #130, opened 2024-04-23
feat: new format for the controller metrics and operations

HarikrishnanBalagopal:
### Description of the change

Standardized the format in the yaml files used by `TrainerControllerCallback`.

### Related issue number

https://github.com/foundation-model-stack/fms-hf-tuning/issues/129

### How to verify the PR

We have tests which can be run with `make test`

### Was the PR tested

Run `make test`

### DIFF
--- a/tuning/trainercontroller/callback.py
+++ b/tuning/trainercontroller/callback.py
@@ -50,11 +50,13 @@ logger = logging.get_logger(__name__)
 CONTROLLER_METRICS_KEY = "controller-metrics"
 OPERATIONS_KEY = "operations"
 CONTROLLERS_KEY = "controllers"
+ARGS_KEY = "arguments"
 
 CONTROLLER_NAME_KEY = "name"
-CONTROLLER_TRIGGERS_KEY = "triggers"
+CONTROLLER_CLASS_KEY = "class"
 CONTROLLER_RULE_KEY = "rule"
-CONTROLLER_OPERATIONS_KEY = "operations"
+CONTROLLER_TRIGGERS_KEY = "triggers"
+CONTROLLER_OPERATIONS_KEY = OPERATIONS_KEY
 
 
 # pylint: disable=too-many-instance-attributes
@@ -74,7 +76,13 @@ class TrainerControllerCallback(TrainerCallback):
         if isinstance(trainer_controller_config, str):
             if os.path.exists(trainer_controller_config):
                 with open(trainer_controller_config, "r", encoding="utf-8") as f:
-                    self.trainer_controller_config = yaml.safe_load(f)
+                    self.trainer_controller_config: dict = yaml.safe_load(f)
+                if not isinstance(self.trainer_controller_config, dict):
+                    raise TypeError(
+                        "expected the trainer controller config YAML file"
+                        "to contain a dictionary. actual type: %s"
+                        % (type(self.trainer_controller_config))
+                    )
             else:
                 raise FileNotFoundError(
                     f"Trainer controller configuration \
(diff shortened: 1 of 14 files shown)


tuning/trainercontroller/callback.py
```diff
@@ -50,11 +51,73 @@
...
+title: "Metrics and Operations"
+description: "Controller metrics and operations."
+type: object
+additionalProperties: false
+properties:
+    "{CONTROLLER_METRICS_KEY}":
+        type: array
+        items:
+            type: object
+            required: ["name", "class"]
```
kmehant, line 73: I see you are reusing the variables `*_KEY` to infix values in to the schema string. Do you also want to do that for required constraint fields as well?


```suggestion
            required: ["{CONTROLLER_NAME_KEY}", "class"]
```

Similar changes wherever needed.

kmehant (changes requested):

tuning/utils/import_utils.py
```diff
@@ -12,11 +12,14 @@
...
+
 # Third Party
 from transformers.utils.import_utils import _is_package_available
 
 _is_aim_available = _is_package_available("aim")
 
 
 def is_aim_available():
-    return _is_aim_available
+    return _is_aim_available and (os.environ.get("AIMSTACK_DB") is not None)
```
dushyantbehl, line 25: @HarikrishnanBalagopal Aim installation and setup does not depend only on the AIMSTACK_DB it depends on server as well. I think its best to keep this change separate and I can have a separate PR for this fix as this leaves the check incomplete.

HarikrishnanBalagopal:
```
I think its best to keep this change separate and I can have a separate PR for this fix as this leaves the check incomplete.
```

We had to fix this bug just for testing this PR that's why it's part of it.

```
@HarikrishnanBalagopal Aim installation and setup does not depend only on the AIMSTACK_DB it depends on server as well. 
```

Yes we can add extra checks, @kmehant suggested checking for `AIMSTACK_DB`. We can check for all 3 environment variables as well.

kmehant: @dushyantbehl We were debugging end to end tests and had to add this check tentatively using `AIMSTACK_DB` env if it has to be `AIMSTACK_SERVER` hari will update this PR. 

Please feel free to open another PR if there are additional checks that has to go into this.

HarikrishnanBalagopal: @dushyantbehl @kmehant I have separated the fix into a new PR.

HarikrishnanBalagopal:
> HI @HarikrishnanBalagopal, thanks for the contribution. I think the schema change looks reasonable, but would like to decouple the schema change and the schema validation to keep things as atomic as possible, especially since the schema validation adds a new dependency. Would you be able to split this into two PRs?

@alex-jw-brooks Thanks for the review, I have removed the jsonschema dependency and schema validation code. PTAL

tuning/trainercontroller/callback.py
```diff
@@ -263,86 +285,102 @@ def on_init_end(
...
-                metric_args = {}
+            metric_args = metric_config[ARGS_KEY] if ARGS_KEY in metric_config else {}
             # Metric handler instance is created here.
-            obj = metric_handler(name=metric_name, **metric_args, **kwargs)
+            metric_handler = metric_handler_class(
+                name=metric_name, **metric_args, **kwargs
+            )
             # Add metric instances to the events.
-            for event_name in obj.get_events():
+            for event_name in metric_handler.get_events():
```
alex-jw-brooks, line 310: just a note in general - in the future it would be best to decouple minor changes and refactors from functional changes to keep PRs as small/atomic as possible, as it keeps the PRs easier to review and less likely to be in conflict with each other 🙂

alex-jw-brooks (changes requested):
One small thing, but otherwise I think we can get this one merged, thanks!

tuning/trainercontroller/callback.py
HarikrishnanBalagopal: Sure

tuning/trainercontroller/callback.py
```diff
@@ -74,7 +76,8 @@ def __init__(self, trainer_controller_config: Union[dict, str]):
         if isinstance(trainer_controller_config, str):
             if os.path.exists(trainer_controller_config):
                 with open(trainer_controller_config, "r", encoding="utf-8") as f:
-                    self.trainer_controller_config = yaml.safe_load(f)
+                    self.trainer_controller_config: dict = yaml.safe_load(f)
+                assert isinstance(self.trainer_controller_config, dict)
```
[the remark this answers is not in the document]
HarikrishnanBalagopal: Done

Issue #103, opened 2024-03-28
Prompt Tuning returns low-quality results


weidotwisc:
## Describe the bug

Prompt Tuning model generates low-quality output

## Platform

Please provide details about the environment you are using, including the following:

- Interpreter version: python 3.11.0
- Library version: transformers 4.37.2 peft 0.8.2

## Sample Code
My launching script is
```bash
export MODEL_PATH=/dccstor/ai4code-ansible/shared/ckpt/granite-20b-code-all-yaml-2k/
export DATA_PATH=/dccstor/weiz/peft/train.json
export OUTPUT_PATH=pt_ckpts
export CUDA_VISIBLE_DEVICES=0
export PYTHONPATH=.
python tuning/sft_trainer.py  \
--model_name_or_path $MODEL_PATH  \
--training_data_path $DATA_PATH  \
--output_dir $OUTPUT_PATH  \
--peft_method "pt" \
--tokenizer_name_or_path $MODEL_PATH  \
--num_train_epochs 5  \
--per_device_train_batch_size 2  \
--per_device_eval_batch_size 1  \
--gradient_accumulation_steps 1  \
--evaluation_strategy "no"  \
--save_strategy "epoch"  \
--learning_rate 3e-2  \
--weight_decay 0.  \
--warmup_ratio 0.0  \
--lr_scheduler_type "linear"  \
--logging_steps 1  \
--include_tokens_per_second  \
--packing False  \
--response_template "\n### Response:"  \
--dataset_text_field "output" \
--use_flash_attn True  \
--torch_dtype "bfloat16" \
       --num_virtual_tokens 100 \
       --max_seq_length 512 \
--prompt_tuning_init RANDOM
```
To repeat, one needs to define MODEL_PATH to the yaml-2k path (which is shared via COS bucket) and DATA_PATH to the training data

## Expected behavior
Using the first *training* data item as the input to the model should show something close to the ground truth, i.e.,  
```
community.docker.docker_login:                                                                                                                         
    registry: "{{ hostvars[groups['docker_registry'][0]].ansible_host }}:{{ registry_port }}"                                                            
    username: "{{ registry_username }}"                                                                                                                  
    password: "{{ registry_password }}"     
``` 

## Observed behavior
When running through inference, I got
```
[...]

PR #132, opened 2024-04-23
feat: add flash attn to inference and eval scripts

anhuong:
### Description of the change

Add `use_flash_attn` flag for the inference and evaluation scripts. When used, this will load the model using flash attention which can affect the inference results.

### Related issue number

related to: #103 

### How to verify the PR

Ran the scripts loading llama-7b model and a Lora tuned llama-7b model that requires merging. Saw that when using flag, the model is loaded with flash attention as noted by the log output.

inference:
```sh
# without
$ python scripts/run_inference.py --model /llama-eval-pvc/LLaMa/models/hf/7B --out_file llama_7b --max_new_tokens 50 --text "Today is a good day. What do you want to do today?" 
No adapter config found! Loading as a merged model...
Loading checkpoint shards: 100%|██████████████████████████████████████████████████████████████| 2/2 [00:25<00:00, 12.86s/it]
Inferred device: cuda
100%|█████████████████████████████████████████████████████████████████████████████████████████| 1/1 [00:02<00:00,  2.29s/it]
Exported results to: llama_7b

# with flash attn
$ python scripts/run_inference.py --model /llama-eval-pvc/LLaMa/models/hf/7B --out_file llama_7b_flash --max_new_tokens 50 --text "Today is a good day. What do you want to do today?" --use_flash_attn
No adapter config found! Loading as a merged model...
You are attempting to use Flash Attention 2.0 with a model not initialized on GPU. Make sure to move the model to GPU after initializing it on CPU with model.to('cuda').
Loading checkpoint shards: 100%|██████████████████████████████████████████████████████████████| 2/2 [00:02<00:00,  1.35s/it]
Inferred device: cuda
100%|█████████████████████████████████████████████████████████████████████████████████████████| 1/1 [00:01<00:00,  1.80s/it]
Exported results to: llama_7b_flash

# lora tuned model with flash attn
$ python scripts/run_inference.py --model /nfs-storage-pvc/llama2-7b-twitter-lora-default-checkpoint-15/ --base_model_name_or_path /llama-eval-pvc/LLaMa/models/hf/7B --out_file llama_7b_lora --max_new_tokens 50 --text "Today is a good day. What do you want to do today?" --use_flash_attn
You are attempting to use Flash Attention 2.0 with a model not initialized on GPU. Make sure to move the model to GPU after initializing it on CPU with `model.to('cuda')`.
Loading checkpoint shards: 100%|██████████████████████████████████████████████████████████████| 2/2 [00:02<00:00,  1.03s/it]
Inferred device: cuda
100%|█████████████████████████████████████████████████████████████████████████████████████████| 1/1 [00:02<00:00,  2.02s/it]
Exported results to: llama_7b_lora
```

### Was the PR tested

### DIFF
--- a/scripts/run_inference.py
+++ b/scripts/run_inference.py
@@ -183,7 +183,7 @@ class TunedCausalLM:
                         attn_implementation="flash_attention_2"
                         if use_flash_attn
                         else None,
-                        torch_dtype=torch.float16 if use_flash_attn else None,
+                        torch_dtype=torch.bfloat16 if use_flash_attn else None,
                     )
                 except OSError as e:
                     print("Failed to initialize checkpoint model!")
@@ -194,7 +194,7 @@ class TunedCausalLM:
             model = AutoModelForCausalLM.from_pretrained(
                 checkpoint_path,
                 attn_implementation="flash_attention_2" if use_flash_attn else None,
-                torch_dtype=torch.float16 if use_flash_attn else None,
+                torch_dtype=torch.bfloat16 if use_flash_attn else None,
             )
 
         device = "cuda" if torch.cuda.is_available() else None
--- a/scripts/run_inference.py
+++ b/scripts/run_inference.py
@@ -142,7 +142,10 @@ class TunedCausalLM:
 
     @classmethod
     def load(
-        cls, checkpoint_path: str, base_model_name_or_path: str = None
+        cls,
+        checkpoint_path: str,
+        base_model_name_or_path: str = None,
+        use_flash_attn: bool = False,
     ) -> "TunedCausalLM":
         """Loads an instance of this model.
 
@@ -152,6 +155,8 @@ class TunedCausalLM:
                 adapter_config.json.
             base_model_name_or_path: str [Default: None]
                 Override for the base model to be used.
+            use_flash_attn: bool [Default: False]
+                Whether to load the model using flash attention.
 
         By default, the paths for the base model and tokenizer are contained within the adapter
         config of the tuned model. Note that in this context, a path may refer to a model to be
@@ -173,14 +178,24 @@ class TunedCausalLM:
         try:
             with AdapterConfigPatcher(checkpoint_path, overrides):
                 try:
-                    model = AutoPeftModelForCausalLM.from_pretrained(checkpoint_path)
+                    model = AutoPeftModelForCausalLM.from_pretrained(
+                        checkpoint_path,
+                        attn_implementation="flash_attention_2"
+                        if use_flash_attn
+                        else None,
+                        torch_dtype=torch.float16 if use_flash_attn else None,
+                    )
                 except OSError as e:
                     print("Failed to initialize checkpoint model!")
                     raise e
         except FileNotFoundError:
             print("No adapter config found! Loading as a merged model...")
             # Unable to find the adapter config; fall back to loading as a merged model
-            model = AutoModelForCausalLM.from_pretrained(checkpoint_path)
+            model = AutoModelForCausalLM.from_pretrained(
+                checkpoint_path,
+                attn_implementation="flash_attention_2" if use_flash_attn else None,
+                torch_dtype=torch.float16 if use_flash_attn else None,
+            )
 
         device = "cuda" if torch.cuda.is_available() else None
         print(f"Inferred device: {device}")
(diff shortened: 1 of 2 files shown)


scripts/run_inference.py
```diff
@@ -173,14 +178,24 @@ def load(
         try:
             with AdapterConfigPatcher(checkpoint_path, overrides):
                 try:
-                    model = AutoPeftModelForCausalLM.from_pretrained(checkpoint_path)
+                    model = AutoPeftModelForCausalLM.from_pretrained(
+                        checkpoint_path,
+                        attn_implementation="flash_attention_2"
+                        if use_flash_attn
+                        else None,
+                        torch_dtype=torch.float16 if use_flash_attn else None,
```
alex-jw-brooks, line 186: Is there a reason that float16 is being hardcoded here since flash attention also supports bfloat32?

alex-jw-brooks (commented):
Thanks Anh! One question

scripts/run_inference.py
anhuong: Yeah I wasn't sure about this, I think I did it based on [this comment](https://github.com/foundation-model-stack/fms-hf-tuning/issues/103#issuecomment-2036003946) and perhaps some examples I saw like [here](https://github.com/huggingface/transformers/blob/bc274a28a9a72c5fe98b070e2083fe689c8fd5e9/docs/source/en/perf_infer_gpu_one.md?plain=1#L102-L115). But I think you're very right that this should be removed

anhuong: If it's removed, what torch_dtype will be used to load the model? If trying to load a model that uses float32 for example this would fail so should this be set by the user? and default to bfloat16 as we have in [sft_trainer](https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/tuning/sft_trainer.py#L134-L135)?

For example I get warning:
```
Flash Attention 2.0 only supports torch.float16 and torch.bfloat16 dtypes, but the current dype in GPTBigCodeModel is torch.float32. You should run training or inference using Automatic Mixed-Precision via the `with torch.autocast(device_type='torch_device'):` decorator, or load the model with the `torch_dtype` argument. Example: `model = AutoModel.from_pretrained("openai/whisper-tiny", attn_implementation="flash_attention_2", torch_dtype=torch.float16)`
```
and then error `RuntimeError: FlashAttention only support fp16 and bf16 data type`

alex-jw-brooks: Ahh sorry, I meant bfloat16, not bfloat32 :) The best choice probably depends on what type the model is tuned with, but the safer choice is probably `bfloat16` since it has the a bigger exponent (same range as float32).

It would be nice to check the behavior for what happens if you try loading with bfloat16 on a device that doesn't support it, to see if it has a fallback (and if that actually modifies the loaded model `dtype` value, or just the way computations are handled) or throws. But the most reasonable choice is probably to do the same thing we are doing for tuning to select the default data type at tune time, or check if bf16 is supported and use it if we can. The easiest way to check if bf16 is available is:

`torch.cuda.is_available() and torch.cuda.is_bf16_supported()`
(the latter throws if you try to call it in torch with no cuda installed)

alex-jw-brooks: We use `bfloat16` as the default for tuning ([ref](https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/tuning/config/configs.py#L40)). So `bfloat16` is the right default here, and I assume since we don't explicitly handle it, `from_pretrained` has a sensible fallback

anhuong: Modified to load with bfloat16 and I tried testing on a v100 where `torch.cuda.is_available() == True` and `torch.cuda.is_bf16_supported() == False`. How do I check the dtype of the model? I ran...
```py
>>> model = AutoModelForCausalLM.from_pretrained("bigscience/bloom-560m", torch_dtype=torch.bfloat16)
config.json: 100%|█████████████████████████████████████████████████████████████████████████| 693/693 [00:00<00:00, 4.49MB/s]
model.safetensors: 100%|███████████████████████████████████████████████████████████████| 1.12G/1.12G [00:14<00:00, 79.3MB/s]
>>> model.config.torch_dtype
torch.bfloat16
```

alex-jw-brooks: The `config.torch_dtype` is set by the model at save time, I think! The one you should want is just the model `.dtype`, which reflects the current instance loaded on your device, I don't think conversions like `.to` (which moves the model to different devices/types etc)  would update the config based attribute but I am not sure

alex-jw-brooks (approved):
LGTM

Issue #131, opened 2024-04-23
bug: AIM package being installed causes the trainer to expect the AIM server to be running.


HarikrishnanBalagopal:
## Describe the bug

If the AIM package is installed then the AIM server is also expected to be running.

```
  File "/dccstor/rhassistant/seshapad/seshaenv/lib/python3.10/site-packages/aim/storage/migrations/utils.py", line 28, in upgrade_database
    raise subprocess.SubprocessError(f'Database upgrade failed with exit code {exit_code}')
subprocess.SubprocessError: Database upgrade failed with exit code 1
```

It should allow running tests and training in environments where the AIM package is installed but there is no AIM server running.

## Platform

Please provide details about the environment you are using, including the following:

- Interpreter version: Python 3.10.4 (main, Mar 31 2022, 08:41:55) [GCC 7.5.0] on linux
- Library version: `main` branch commit `db99b280e95c9e5822ee83beee050ba5575ab3bf`

## Sample Code

Steps to reproduce:
1. Create a `conda` or `venv` environment with the `aim` package and this library installed.
3. Run https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/tuning/sft_trainer.py

## Expected behavior

Training should complete without errors. Only expect an AIM server to be running if the appropriate environment variables like `AIMSTACK_DB` are set.

## Observed behavior

Training fails with an error talking about AIM

```

  File "/dccstor/rhassistant/seshapad/seshaenv/lib/python3.10/site-packages/aim/ext/exception_resistant.py", line 68, in wrapper
    return func(*args, **kwargs)
  File "/dccstor/rhassistant/seshapad/seshaenv/lib/python3.10/site-packages/aim/sdk/run.py", line 859, in __init__
    super().__init__(run_hash, repo=repo, read_only=read_only, experiment=experiment, force_resume=force_resume)
  File "/dccstor/rhassistant/seshapad/seshaenv/lib/python3.10/site-packages/aim/sdk/run.py", line 272, in __init__
    super().__init__(run_hash, repo=repo, read_only=read_only, force_resume=force_resume)
  File "/dccstor/rhassistant/seshapad/seshaenv/lib/python3.10/site-packages/aim/sdk/base_run.py", line 34, in __init__
    self.repo = get_repo(repo)
  File "/dccstor/rhassistant/seshapad/seshaenv/lib/python3.10/site-packages/aim/sdk/repo_utils.py", line 24, in get_repo
[...]

PR #139, opened 2024-04-25
fix: bug with AIM server availability check

HarikrishnanBalagopal:
### Description of the change

Fixes the bug in https://github.com/foundation-model-stack/fms-hf-tuning/issues/131

### Related issue number

https://github.com/foundation-model-stack/fms-hf-tuning/issues/131

### How to verify the PR

Run an end-to-end test using https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/tuning/sft_trainer.py in a `venv` or `conda` environment which has the AIM package installed. Even without an AIM server running the training should succeed.

### Was the PR tested

### DIFF
--- a/tuning/aim_loader.py
+++ b/tuning/aim_loader.py
@@ -38,6 +38,6 @@ def get_aimstack_callback():
     if aim_db:
         aim_callback = AimCallback(repo=aim_db, experiment=aim_experiment)
     else:
-        aim_callback = AimCallback(experiment=aim_experiment)
+        aim_callback = None
 
     return aim_callback
--- a/tuning/sft_trainer.py
+++ b/tuning/sft_trainer.py
@@ -244,7 +244,9 @@ def train(
 
     callbacks = [FileLoggingCallback(logger)]
     if is_aim_available():
-        callbacks.append(get_aimstack_callback())
+        aimstack_callback = get_aimstack_callback()
+        if aimstack_callback is not None:
+            callbacks.append(aimstack_callback)
 
     if (trainer_controller_args is not None) and (
         trainer_controller_args.trainer_controller_config_file is not None


dushyantbehl:
@HarikrishnanBalagopal I believe we do not need so much changes to the code. The idea is if Aim is installed but `AIMSTACK_DB` and `AIMSTACK_SERVER` are not set then we should not use Aim.

Just change this one line 

```
  if aim_server:
        aim_callback = AimCallback(
            repo="aim://" + aim_server + "/", experiment=aim_experiment
        )
    if aim_db:
        aim_callback = AimCallback(repo=aim_db, experiment=aim_experiment)
   else:
-       aim_callback = AimCallback(experiment=aim_experiment)
+       aim_callback = None
```

HarikrishnanBalagopal:
> @HarikrishnanBalagopal I believe we do not need so much changes to the code. The idea is if Aim is installed but `AIMSTACK_DB` and `AIMSTACK_SERVER` are not set then we should not use Aim.
> 
> Just change this one line
> 
> ```
>   if aim_server:
>         aim_callback = AimCallback(
>             repo="aim://" + aim_server + "/", experiment=aim_experiment
>         )
>     if aim_db:
>         aim_callback = AimCallback(repo=aim_db, experiment=aim_experiment)
>    else:
> -       aim_callback = AimCallback(experiment=aim_experiment)
> +       aim_callback = None
> ```

I think that will also require a change here to avoid adding `None`s into the list of callbacks https://github.com/foundation-model-stack/fms-hf-tuning/blob/c2f2f8c2179f38f1754d621d4bb13cd4c1eaca52/tuning/sft_trainer.py#L247

dushyantbehl:
Wait for #89 which fixes this.

HarikrishnanBalagopal:
> Wait for #89 which fixes this.

Sure, I have made the change in this PR as well.

anhuong:
if this is in #89 perhaps we can close this PR

ashokponkumar:
@HarikrishnanBalagopal If #89 fixes this issue, lets close this PR. Thanks @anhuong !

PR #143, opened 2024-04-25
fix: Handling of FSDP Wrapping for PEFT in Training Script

achew010:
### Description of the change

I discovered an issue with how PEFT with FSDP is handled in the tuning script, `sft_trainer.py`.  In this [portion](https://github.com/foundation-model-stack/fms-hf-tuning/blob/8548a6df86e0a3ea00ece11a47c3aac2971a512e/tuning/sft_trainer.py#L272C1-L274C10) of the tuning script, the FSDP wrap policy is applied on a full model. This will result in an incorrect wrapping of FSDP layers that won't be applied on the adapter layers and subsequently produce wrong gradients in training.

Instead, after `SFTTrainer` creates the PEFT model in its initialization, then apply the wrap policy on the trainer's model.

I made the change in the code to correct the behaviour. Below, i demonstrate the effect of the bug and how the fix eliminates the problem.

### Tests

| The training was observed to produce nan gradients in distributed PEFT | After the fix is introduced, we do not see the nan gradients anymore |
| -------- | ------- |
|   |     |

### DIFF
--- a/tuning/sft_trainer.py
+++ b/tuning/sft_trainer.py
@@ -276,7 +276,7 @@ def train(
 
     if trainer.is_fsdp_enabled and peft_config is not None:
         trainer.accelerator.state.fsdp_plugin.auto_wrap_policy = fsdp_auto_wrap_policy(
-            model
+            trainer.model
         )
     trainer.train()
 


fabianlim:
@achew010 I see, so to summarize
- before fix:  since `model` is not yet PEFT wrapped, calling `fsdp_auto_wrap_policy(model)`will not wrap the LoRA adapters in fsdp units
- after fix: `trainer.model` is already PEFT wrapped, so calling `fsdp_auto_wrap_policy(trainer.model)` works.

A more API-compliant way to do this. is to use the [`on_init_end` trainer Callback](https://github.com/huggingface/transformers/blob/a98c41798cf6ed99e1ff17e3792d6e06a2ff2ff3/src/transformers/trainer.py#L669). SFTTrainer calls `super().__init__` after installing PEFT adapters.

@lchu-ibm

PR #146, opened 2024-05-03
feat: Exposed the evaluation metrics for rules within trainer controller

### LINKED ISSUE #136: feat: Exposed the evaluation metrics for rules within trainer controller

The evaluation metrics based stopping criteria for training jobs is not available yet.

The evaluation metrics could be customized in the hugging face. Using these metrics within the stopping criteria rules of the trainer controller framework will help stop training if the training is not making progress.

Not applicable

## Additional context

Not applicable

seshapad:
### Description of the change

 Evaluation metrics could be customized in the hugging face. Using these metrics within the stopping criteria rules of the trainer controller framework will help stop training if the training is not making progress. This PR exposes the evaluation metrics to be used in trainer controller rules.

### Related issue number

Closes #136 

### How to verify the PR
Add trainer controller configuration for `eval metrics`.

### Was the PR tested

### DIFF
--- a/tuning/trainercontroller/controllermetrics/eval_metrics.py
+++ b/tuning/trainercontroller/controllermetrics/eval_metrics.py
@@ -1,0 +1,75 @@
+# Copyright The IBM Tuning Team
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# SPDX-License-Identifier: Apache-2.0
+# https://spdx.dev/learn/handling-license-info/
+
+# Standard
+from typing import Any
+
+# Third Party
+from transformers.utils import logging
+
+# Local
+from tuning.trainercontroller.controllermetrics.metricshandler import MetricHandler
+
+logger = logging.get_logger(__name__)
+
+
+class EvalMetrics(MetricHandler):
+    """Implements the controller metric which exposes the evaluation metrics"""
+
+    def __init__(self, **kwargs):
+        """Initializes the metric handler, by registering the event \
+            list and arguments with base handler.
+
+        Args:
+            kwargs: List of arguments (key, value)-pairs
+        """
+        source_events_to_check = {"on_evaluate", "on_predict"}
+        source_event = kwargs.get("source-event")
+        if source_event is None:
+            source_event = "on_evaluate"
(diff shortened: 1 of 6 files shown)


tuning/trainercontroller/controllermetrics/eval_metrics.py
```diff
@@ -0,0 +1,65 @@
...
+    def __init__(self, **kwargs):
+        """Initializes the metric handler, by registering the event \
+            list and arguments with base handler.
+
+        Args:
+            kwargs: List of arguments (key, value)-pairs
+        """
+        super().__init__(
+            events=[
+                kwargs.get("source-event", "on_evaluate"),
```
alex-jw-brooks, line 42: Could you explain the need for the source-event arg here, instead of just hardcoding the event to `on_evaluate`? I.e., aren't the metrics in the trainer callback only available for `on_evaluate` ([ref](https://huggingface.co/docs/transformers/en/main_classes/callback#transformers.TrainerCallback.metrics))?

tuning/trainercontroller/controllermetrics/eval_metrics.py
```diff
@@ -0,0 +1,65 @@
...
+    def compute(self, **kwargs) -> Any:
+        """Exposes the trainer state.
+
+        Args:
+            kwargs: Remaining event arguments
+
+        Returns:
+            dict. Trainer state as a dictionary
+        """
+        return kwargs["metrics"]
```
alex-jw-brooks, line 65: Related to above - if the source event is provided and not evaluation, does this need to check and inspect to make sure `metrics` is in the dict?

tuning/trainercontroller/controllermetrics/eval_metrics.py
seshapad: @alex-jw-brooks We put this line to cater to `on_predict` event as well. It seems metrics are passed to this event too (please refer to [link](https://github.com/huggingface/transformers/blob/4fdf58afb72b0754da30037fc800b6044e7d9c99/src/transformers/trainer_callback.py#L283)).

tuning/trainercontroller/controllermetrics/eval_metrics.py
seshapad: Addressed this by adding a validation for source events. This should ensure that the case of `metrics` not being available will not occur.

tuning/trainercontroller/controllermetrics/eval_metrics.py
alex-jw-brooks: Ahh that is really interesting, maybe that section of the documentation saying it is only accessible from eval is outdated, and I guess custom callbacks can kind of do what they want anyway, so the metrics provided to those events might not always be the same 🤔 Thank you for the explanation, makes sense

tuning/trainercontroller/controllermetrics/eval_metrics.py
alex-jw-brooks: Sounds good, thanks for that!

alex-jw-brooks (approved):
Thanks for the contribution! LGTM

PR #147, opened 2024-05-03
fix: replace eval with a safer alternative

HarikrishnanBalagopal:
### Description of the change

Use the AST module to evaluate the rules instead of `eval`.

1. We use the Python AST library to parse the code.
2. We use this library https://github.com/danthedeckie/simpleeval for traversing and evaluating the generated AST.
3. The output of evaluation should be a boolean value which we return indicating if the rule succeeded.

Example supported expressions: https://github.com/danthedeckie/simpleeval?tab=readme-ov-file#operators

Support for more complex types (list, dict, etc.) are also implemented:
- https://github.com/danthedeckie/simpleeval?tab=readme-ov-file#compound-types
- https://github.com/danthedeckie/simpleeval/blob/2a12b5856d6f70b78dc1ac38840c80c1be6c6c4e/simpleeval.py#L638-L642

Notes:
- PyPi https://pypi.org/project/simpleeval/ (MIT Licence)
- Array Multiplication `["hello"]*10` is allowed but limited to smaller numbers to avoid Denial of Service (DOS).
- List comprehension is allowed but limited to small number of iterations to avoid DOS.
- Exponentiation `9**9` is allowed but limited to smaller numbers to avoid DOS.
- Access to double underscore attributes like `__class__` is disallowed to avoid access to arbitrary classes like `Quitter`.
- Lambda expressions are disallowed for simplicity and to reduce the attack surface. Can be added back in if necessary.
- Access to builtin functions and globals are disallowed expect for `int`, `float`, `str`, `rand` and `randint`.
- Has support for accessing keys in dicts using the short syntax:`foo.bar` . Example dict: `{"foo": {"bar": 42}})`

### Related issue number

Fixes https://github.com/foundation-model-stack/fms-hf-tuning/issues/148

### How to verify the PR

Have added unit tests for the safe evaluator.

`make test`

### Was the PR tested

### DIFF
--- a/tuning/trainercontroller/callback.py
+++ b/tuning/trainercontroller/callback.py
@@ -17,12 +17,13 @@
 
 # Standard
 from importlib import resources as impresources
-from typing import List, Union
+from typing import Dict, List, Union
 import inspect
 import os
 import re
 
 # Third Party
+from simpleeval import EvalWithCompoundTypes, FeatureNotAvailable, NameNotDefined
 from transformers import (
     TrainerCallback,
     TrainerControl,
@@ -43,6 +44,7 @@ from tuning.trainercontroller.operations import Operation
 from tuning.trainercontroller.operations import (
     operation_handlers as default_operation_handlers,
 )
+from tuning.utils.evaluator import get_evaluator
 
 logger = logging.get_logger(__name__)
 
@@ -174,7 +176,7 @@ class TrainerControllerCallback(TrainerCallback):
         self.register_operation_handlers(default_operation_handlers)
 
         # controls
-        self.control_actions_on_event: dict[str, list[Control]] = {}
+        self.control_actions_on_event: Dict[str, list[Control]] = {}
 
         # List of fields produced by the metrics
         self.metrics = {}
@@ -208,23 +210,26 @@ class TrainerControllerCallback(TrainerCallback):
                 self.metrics[m.get_name()] = m.compute(event_name=event_name, **kwargs)
 
     def _take_control_actions(self, event_name: str, **kwargs):
-        """Invokes the act() method for all the operations registered for a given event. \
-            Note here that the eval() is invoked with `__builtins__` set to None. \
-            This is a precaution to restric the scope of eval(), to only the \
-            fields produced by the metrics.
+        """Invokes the act() method for all the operations registered for a given event.
 
         Args:
             event_name: str. Event name.
             kwargs: List of arguments (key, value)-pairs.
         """
         if event_name in self.control_actions_on_event:
+            evaluator = get_evaluator(metrics=self.metrics)
             for control_action in self.control_actions_on_event[event_name]:
                 rule_succeeded = False
                 try:
-                    # pylint: disable=eval-used
-                    rule_succeeded = eval(
-                        control_action.rule, {"__builtins__": None}, self.metrics
+                    rule_succeeded = evaluator.eval(
+                        expr=control_action.rule_str,
+                        previously_parsed=control_action.rule,
                     )
+                    if not isinstance(rule_succeeded, bool):
+                        raise TypeError(
+                            "expected the rule to evaluate to a boolean. actual type: %s"
+                            % (type(rule_succeeded))
+                        )
                 except TypeError as et:
                     raise TypeError("Rule failed due to incorrect type usage") from et
                 except ValueError as ev:
@@ -235,6 +240,14 @@ class TrainerControllerCallback(TrainerCallback):
                     raise NameError(
                         "Rule failed due to use of disallowed variables"
                     ) from en
+                except NameNotDefined as en1:
+                    raise NameError(
+                        "Rule failed because some of the variables are not defined"
+                    ) from en1
+                except FeatureNotAvailable as ef:
+                    raise NotImplementedError(
+                        "Rule failed because it uses some unsupported features"
+                    ) from ef
                 if rule_succeeded:
                     for operation_action in control_action.operation_actions:
                         logger.info(
@@ -374,9 +387,11 @@ class TrainerControllerCallback(TrainerCallback):
                             % (controller_name, event_name)
                         )
                     # Generates the byte-code for the rule from the trainer configuration
+                    curr_rule = controller[CONTROLLER_RULE_KEY]
                     control = Control(
-                        name=controller_name,
-                        rule=compile(controller_rule, "", "eval"),
+                        name=controller[CONTROLLER_NAME_KEY],
+                        rule_str=curr_rule,
+                        rule=EvalWithCompoundTypes.parse(expr=curr_rule),
                         operation_actions=[],
                     )
                     for control_operation_name in controller_ops:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,7 +36,8 @@ dependencies = [
 "trl",
 "peft>=0.8.0",
 "datasets>=2.15.0",
-"fire"
+"fire",
+"simpleeval",
 ]
 
 [project.optional-dependencies]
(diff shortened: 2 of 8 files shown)


architecture_records/001-trainer-controller-framework.md
```diff
@@ -58,18 +58,20 @@ We have implemented a trainer callback (see [here](https://huggingface.co/docs/t
 3. Control operation and action that needs to be performed if rule is evaluated to true.
```
alex-jw-brooks, line 58: Can you please separate the proposed changes into a different pull request, and add the following?

In the PR description for the ADR:
- A quick explanation for why this is safer than `eval`, e.g., malicious operations that can be added through `eval`, but not `simpleeval` to motivate why we should potentially add a new dependency to support this

And in the ADR:
- what things like "smaller numbers" mean to prevent DOS problems - I assume these limits are implemented in `simpleeval` and not specifically meant to be handled in this codebase, as this is a problem that is applicable to anything being executed by that library - is that correct?
- How "big numbers" etc that are not executed because of potential DOS attacks bubble up as errors for users executing this library, as it would be nice to keep error messages interpretable where possible

tuning/trainercontroller/callback.py
```diff
@@ -235,6 +261,14 @@ def _take_control_actions(self, event_name: str, **kwargs):
                     raise NameError(
                         "Rule failed due to use of disallowed variables"
                     ) from en
+                except NameNotDefined as en1:
+                    raise NameError(
+                        "Rule failed because some of the variables are not defined"
```
alex-jw-brooks, line 266: Are there any ways to make these exceptions more specific, e.g., what variables are not defined for this one, or what features are not supported in the one below this?

tuning/trainercontroller/callback.py
```diff
@@ -59,6 +67,21 @@
 CONTROLLER_OPERATIONS_KEY = OPERATIONS_KEY
 
 
+def get_evaluator(metrics: dict) -> EvalWithCompoundTypes:
```
alex-jw-brooks, line 70: Any thoughts on making this part of the class, e.g., as a staticmethod? It feels a bit strange to me to store it separately here like this

tuning/trainercontroller/callback.py
```diff
@@ -59,6 +67,21 @@
...
+    """Returns an evaluator that can be used to evaluate simple Python expressions."""
+    all_names = {
+        **metrics,
+        **DEFAULT_NAMES.copy(),
+    }
+    all_funcs = {
+        "abs": abs,
+        "len": len,
+        "sqrt": sqrt,
+        **DEFAULT_FUNCTIONS.copy(),
```
alex-jw-brooks, line 80: I see the default functions defined here: https://github.com/danthedeckie/simpleeval/blob/2a12b5856d6f70b78dc1ac38840c80c1be6c6c4e/simpleeval.py#L315

How was the total set of all_funcs created? For example, why were abs/len/sqrt added here, while things like `pow` were excluded from the total set of all funcs?

tuning/trainercontroller/callback.py
```diff
@@ -208,23 +231,26 @@ def _compute_metrics(self, event_name: str, **kwargs):
...
+            evaluator = get_evaluator(metrics=self.metrics)
             for control_action in self.control_actions_on_event[event_name]:
                 rule_succeeded = False
                 try:
-                    # pylint: disable=eval-used
-                    rule_succeeded = eval(
-                        control_action.rule, {"__builtins__": None}, self.metrics
+                    rule_succeeded = evaluator.eval(
+                        expr=control_action.rule_str,
+                        previously_parsed=control_action.rule,
```
alex-jw-brooks, line 247: Can you give an example of where the previously parsed contents would actually be used here? More specifically, if `get_evaluator()` is building a new instance of `EvalWithCompoundTypes` on line 241 - It seems like it is in case there are duplicated rule strings across different control actions for the same event. Is that correct, and are there scenarios where that would be expected?

tests/trainercontroller/test_simpleeval.py
```diff
@@ -0,0 +1,164 @@
...
+            actual_rule_is_true = evaluator.eval(
+                expr=rule,
+                previously_parsed=rule_parsed,
+            )
+            assert (
+                actual_rule_is_true == expected_rule_is_true
+            ), "failed to execute the rule"
+        else:
+            with pytest.raises(Exception) as exception_handler:
+                actual_rule_is_true = evaluator.eval(
```
alex-jw-brooks, line 160: Should the value of `actual_rule_is_true` be validated in any way, since this is only checked in the previous block?

alex-jw-brooks (changes requested):
Thanks, this is a really interesting contribution - a lot of questions and one request to split the PR up a little 😄

architecture_records/001-trainer-controller-framework.md
HarikrishnanBalagopal: @alex-jw-brooks Will create an ADR with more details but for a quick reference you can look at the unit tests that contain some of the malicious strings. It also has examples of the "big numbers".

On safety, mainly 2 points:
- At a high level `simpleeval` is safer than `eval` because it is a whitelist. It simply doesn't have to support the entirety of Python language featues like `eval` has to.
- `simpleeval` also adds runtime checks (for array multiplication, attribute access, big numbers, etc.) to mitigate the impact of malicious code that can cause DOS.

tuning/trainercontroller/callback.py
HarikrishnanBalagopal: Yes the Exception does contain the name of the variable that is not defined https://github.com/foundation-model-stack/fms-hf-tuning/pull/147/files#diff-73791d95640c88e27cc8aef08ee77265bd2939f7f7a0691640aabb3f6394d932R62

We can add more information if necessary.

tuning/trainercontroller/callback.py
HarikrishnanBalagopal: This is just a set of functions that we felt would be useful. Functions like `pow` can be added if necessary (with some additional checks to make sure the inputs to `pow` are "small numbers").

tuning/trainercontroller/callback.py
HarikrishnanBalagopal: Making it a static method would require overriding the class. Since this is just a utility function I will move it to a more appropriate location. We can also remove the function entirely if necessary.

tests/trainercontroller/test_simpleeval.py
HarikrishnanBalagopal: I don't think it makes sense to validate `actual_rule_is_true` in the case where the rule is invalid and is supposed to raise an exception of some kind. The `with pytest.raises(Exception) as exception_handler:` does make sure to fail the test if no exception is raised by the code inside the `with` block.

I can remove this variable if necessary.

tuning/trainercontroller/callback.py
HarikrishnanBalagopal: I am not sure I entirely understand the question.

When the `previously_parsed` argument is passed in, the `expr` argument is not used. https://github.com/danthedeckie/simpleeval/blob/2a12b5856d6f70b78dc1ac38840c80c1be6c6c4e/simpleeval.py#L429

We are only giving `expr=control_action.rule_str,` for nice error messages.

tuning/trainercontroller/callback.py
alex-jw-brooks: That sounds reasonable to me - I agree that minimizing the list of potential supported funcs in the beginning and potentially expanding it as needed is the better approach, as it would be more difficult to remove support for funcs later on (i.e., since it could break working rules, which has some implications for the type of version bump we would need to make for the library to follow good practices)

tuning/trainercontroller/callback.py
alex-jw-brooks: Thanks that helps - I was partially asking about deduplication, but this
> We are only giving expr=control_action.rule_str, for nice error messages. We can make it an empty string if necessary.

clarifies my confusion. I just didn't understand why both the expr / previously parsed were being passed here when the non-singleton evaluator is being created right before this loop is entered.

alex-jw-brooks (approved):
Looks good to me. LGTM, thanks!

PR #150, opened 2024-05-08
feat: Trainer state as a trainer controller metric

### LINKED ISSUE #134: feat: Expose the trainer state as a trainer controller metric

This feature is an add-on to the trainer controller framework. We add a new metric to this framework to expose the trainer state of the trainer loop.

This metric now enables rules like the below example, wherein we could check the epochs elapsed in order to stop training on a loss threshold.
```
controller-metrics:
  - name: state
    class: StateOfTrainer
  - name: loss
    class: Loss
controllers:
  - name: loss-controller
    triggers:
      - on_log
    rule: loss < 2.2 and state['epoch'] > 5
    operations:
      - hfcontrols.should_training_stop
```

We checked if the trainer state is a call-by-reference object which could be obtained once and referred many-times, but this is not how the trainer loop exposes the trainer state. The only way we could find the trainer state to be exposed is in through the arguments of the events of trainer callback.

## Additional context

NA

seshapad:
### Description of the change

This PR exposes the trainer state at every trainer loop event as a metric, which could be used as part of the trainer controller rules.

### Related issue number

Closes #134

### How to verify the PR

Apart from the test cases, the feature could be validated by running the trainer scripts using the example configuration:
```
python ./tuning/sft_trainer.py
...
--trainer_controller_config_file examples/trainercontroller_configs/loss.yaml
...
```

### Was the PR tested


alex-jw-brooks (approved):
This one looks good to me, thanks! While thinking about the `TrainerState`, I did have a question though - does special care need to be taken to handle distributed training, specifically on setting `should_training_stop` on the trainer state depending on whether or not the process executing is 0? I'm not sure off the top of my head if setting it from other processes would cause weird behavior or not

seshapad:
> This one looks good to me, thanks! While thinking about the `TrainerState`, I did have a question though - does special care need to be taken to handle distributed training, specifically on setting `should_training_stop` on the trainer state depending on whether or not the process executing is 0? I'm not sure off the top of my head if setting it from other processes would cause weird behavior or not

@alex-jw-brooks Firstly,  Thank you very much for the prompt response. 

Following are two points which attempt to answer your queries:

1. `TrainerState` does not have any control variables like `should_training_stop`. Therefore we will not be risking any control exposure.

1. Just to keep you in the loop, we will ne addressing the trainer controller behavior in multigpu setting in subsequent PRs. However, the current PR 150 is not related to this.

alex-jw-brooks:
That makes sense, thanks! To clarify, I didn't mean as a control variable, I was just asking more generally if it breaking out of the training loop with this callback would cause issues if it were executed from a process besides 0. let's leave that discussion for future then, I'll update and merge the changes from main in and merge this one 😄

Issue #148, opened 2024-05-03
bug: `eval` is still not safe even with checks for `__` and `"__builtins__": None`


HarikrishnanBalagopal:
https://github.com/foundation-model-stack/fms-hf-tuning/blob/dd29d491b57e7efd28c808da17380c54f6bbc0d0/tuning/trainercontroller/callback.py#L198-L200

Some examples of malicious strings that get past current checks are in the unit tests in the corresponding PR that fixes this issue:

https://github.com/foundation-model-stack/fms-hf-tuning/pull/147/files#diff-73791d95640c88e27cc8aef08ee77265bd2939f7f7a0691640aabb3f6394d932R66-R81

PR #151, opened 2024-05-09
docs: ADR for moving from `eval` to `simpleeval` for evaluating trainer controller rules

HarikrishnanBalagopal:
### Description of the change

ADR for documenting the change from `eval` to `simpleeval`.

## Safety

> A quick explanation for why this is safer than eval, e.g., malicious operations that can be added through eval, but not simpleeval to motivate why we should potentially add a new dependency to support this

On safety, at a high level mainly 2 points:
- `simpleeval` is safer than `eval` because it is a whitelist. It simply doesn't have to support the entirety of the Python language features that `eval` has to.
- `simpleeval` also adds runtime checks (for array multiplication, attribute access, big numbers, etc.) to catch malicious code that bypass the simple regex based checks https://github.com/foundation-model-stack/fms-hf-tuning/blob/74caf85140a112cd9289502b0777baac636adf1d/tuning/trainercontroller/callback.py#L267 and also to reduce the chance for DOS.

This whitelist approach follows the recommended alternative to `eval` https://youtu.be/7e-2nUqnNqE?t=574

## Convenience

In addition to safety, this also gives us the ability to add support for nice syntax sugar. Example for accessing deeply nested trees of dictionaries and lists.
- https://github.com/danthedeckie/simpleeval/tree/master?tab=readme-ov-file#compound-types
- https://github.com/foundation-model-stack/fms-hf-tuning/pull/147/files#diff-73791d95640c88e27cc8aef08ee77265bd2939f7f7a0691640aabb3f6394d932R52

## Examples

- https://github.com/foundation-model-stack/fms-hf-tuning/pull/147/files#diff-73791d95640c88e27cc8aef08ee77265bd2939f7f7a0691640aabb3f6394d932R100
This is one of many examples that bypasses the checks in the current implementation and hangs the runtime (DOS).

- Another example using "big numbers" https://github.com/foundation-model-stack/fms-hf-tuning/pull/147/files#diff-73791d95640c88e27cc8aef08ee77265bd2939f7f7a0691640aabb3f6394d932R117

- Many more examples are available in the test suite of the library https://github.com/danthedeckie/simpleeval/blob/2a12b5856d6f70b78dc1ac38840c80c1be6c6c4e/test_simpleeval.py#L379-L397

### Related issue number

Proposed fix for #148 

### How to verify the PR

NA

### Was the PR tested

### DIFF
--- a/architecture_records/001-trainer-controller-framework.md
+++ b/architecture_records/001-trainer-controller-framework.md
@@ -51,6 +51,15 @@ Could add to the overhead of the trainer loop as the callback is invoked at vari
 
 ## Detailed Design
 
+### High-level architecture
+Following is a high-level design diagram. Following are the touch-points to the framework through which user of this framework could interact with it:
+- **Registration**: The registration mechanism de-couples the metrics and operators from the trainer framework. A user could implement a custom metric or operator and register it through the registration methods mentioned previously. This makes the framework highly extensible.
+
+- **Configuration**: The trainer controller configuration supplies the definition for triggers, rule, operations and metrics to orchestrate the enactment of a particular control policy. These details are split up and passed off to the respective modules by the trainer controller as shown in the figure.
+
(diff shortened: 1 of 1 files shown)


alex-jw-brooks (approved):
Looks good to me, thanks for the writeup and reference links - I think there are enough compelling reasons to make this switch and add the new dependency, especially given that `simpleeval` doesn't have any dependencies itself

PR #127, opened 2024-04-17
OS update in dockerfile

jbusche:
### Description of the change

This Dockerfile change does two things:
1.  It does a `dnf update -y` so that we catch the latest fixes that might not yet be in the latest ubi build. For example in this case there are 2 known `libexpat` medium vulnerabilities that this update method closes in our image.
2. This removes a high `Private keys stored in image` issue with `Found: /usr/share/doc/perl-Net-SSLeay/examples/server_key.pem`
### Related issue number

Closes issue https://github.ibm.com/ai-foundation/watson-fm-stack-tracker/issues/758
### How to verify the PR
(Note, you need an environment running with Kueue and KFTO)  Otherwise, it might be possible with Kind.

1. Clone the branch to your podman environment
```
git clone https://github.com/jbusche/fms-hf-tuning.git -b jb-issue-758
cd fms-hf-tunin
```
2.  Build the image:
```
docker build -t fms-hf-tuning:issue758-1 . -f build/Dockerfile
```
3.  Push it to quay.io for testing:
```
podman tag localhost/fms-hf-tuning:issue758-1 quay.io/jbusche/fms-hf-tuning:issue758-1
podman push quay.io/jbusche/fms-hf-tuning:issue758-1
```
4.  Try Ted's testing method:
```
cat <<EOF | kubectl apply -f -
apiVersion: v1
kind: ConfigMap
metadata:
  name: my-config
data:
  config.json: |
    {
      "accelerate_launch_args": {
        "num_machines": 1,
        "num_processes": 2
      },
      "model_name_or_path": "bigscience/bloom-560m",
      "training_data_path": "/etc/config/twitter_complaints_small.json",
      "output_dir": "/tmp/out",
      "num_train_epochs": 1.0,
      "per_device_train_batch_size": 4,
      "per_device_eval_batch_size": 4,
      "gradient_accumulation_steps": 4,
      "evaluation_strategy": "no",
      "save_strategy": "epoch",
      "learning_rate": 1e-5,
      "weight_decay": 0.0,
      "lr_scheduler_type": "cosine",
      "logging_steps": 1.0,
      "packing": false,
      "include_tokens_per_second": true,
      "response_template": "\n### Label:",
      "dataset_text_field": "output",
      "use_flash_attn": false,
      "torch_dtype": "float32",
      "peft_method": "pt",
      "tokenizer_name_or_path": "bigscience/bloom"
    }
  twitter_complaints_small.json: |
[...]

PR #154, opened 2024-05-10
Switch to multistage dockerfile

tharapalanivel:
### Description of the change

Switch to multistage build 

### How to verify the PR

Build and run image locally similar to steps in the [docs](https://github.com/foundation-model-stack/fms-hf-tuning/tree/main/build#readme), check tuned model and logs exists in the output directory specified in job config. Additionally also load up the tuned model and run inference on it as an extra measure.

### Was the PR tested

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+<<<<<<< HEAD
 FROM registry.access.redhat.com/ubi9/python-311 as wheel
 
 ARG WHEEL_VERSION=""
@@ -36,13 +37,33 @@ ARG CUDA_VERSION=11.8.0
 ARG USER=tuning
 ARG USER_UID=1000
 ARG SET_NUM_PROCESSES_TO_NUM_GPUS=True
+=======
 
-USER root
+## Global Args #################################################################
+ARG BASE_UBI_IMAGE_TAG=9.3-1552
+ARG PROTOC_VERSION=25.2
+ARG PYTORCH_INDEX="https://download.pytorch.org/whl"
+>>>>>>> d876974 (Multi-stage build)
+
+# match PyTorch version that was used to compile flash-attention v2 pre-built wheels
+# e.g. flash-attn v2.5.2 => torch ['1.12.1', '1.13.1', '2.0.1', '2.1.2', '2.2.0', '2.3.0.dev20240126']
+# https://github.com/Dao-AILab/flash-attention/blob/v2.5.2/.github/workflows/publish.yml#L47
+# use nightly build index for torch .dev pre-release versions
+ARG PYTORCH_VERSION=2.2.1
+
+ARG PYTHON_VERSION=3.11
+
+
+## Base Layer ##################################################################
+FROM registry.access.redhat.com/ubi9/ubi:${BASE_UBI_IMAGE_TAG} as base
+WORKDIR /app
+
+ARG PYTHON_VERSION
 
 RUN dnf remove -y --disableplugin=subscription-manager \
         subscription-manager \
         # we install newer version of requests via pip
-        python3.11-requests \
+    python${PYTHON_VERSION}-requests \
     && dnf install -y make \
         # to help with debugging
         procps \
@@ -51,18 +72,23 @@ RUN dnf remove -y --disableplugin=subscription-manager \
 ENV LANG=C.UTF-8 \
     LC_ALL=C.UTF-8
 
-ENV CUDA_VERSION=$CUDA_VERSION \
-    NV_CUDA_LIB_VERSION=11.8.0-1 \
+
+## CUDA Base ###################################################################
+FROM base as cuda-base
+
+# Ref: https://docs.nvidia.com/cuda/archive/12.1.0/cuda-toolkit-release-notes/
+ENV CUDA_VERSION=12.1.0 \
+    NV_CUDA_LIB_VERSION=12.1.0-1 \
     NVIDIA_VISIBLE_DEVICES=all \
     NVIDIA_DRIVER_CAPABILITIES=compute,utility \
-    NV_CUDA_CUDART_VERSION=11.8.89-1 \
-    NV_CUDA_COMPAT_VERSION=520.61.05-1
+    NV_CUDA_CUDART_VERSION=12.1.55-1 \
+    NV_CUDA_COMPAT_VERSION=530.30.02-1
 
 RUN dnf config-manager \
        --add-repo https://developer.download.nvidia.com/compute/cuda/repos/rhel9/x86_64/cuda-rhel9.repo \
     && dnf install -y \
-        cuda-cudart-11-8-${NV_CUDA_CUDART_VERSION} \
-        cuda-compat-11-8-${NV_CUDA_COMPAT_VERSION} \
+        cuda-cudart-12-1-${NV_CUDA_CUDART_VERSION} \
+        cuda-compat-12-1-${NV_CUDA_COMPAT_VERSION} \
     && echo "/usr/local/nvidia/lib" >> /etc/ld.so.conf.d/nvidia.conf \
     && echo "/usr/local/nvidia/lib64" >> /etc/ld.so.conf.d/nvidia.conf \
     && dnf clean all
@@ -71,43 +97,48 @@ ENV CUDA_HOME="/usr/local/cuda" \
     PATH="/usr/local/nvidia/bin:${CUDA_HOME}/bin:${PATH}" \
     LD_LIBRARY_PATH="/usr/local/nvidia/lib:/usr/local/nvidia/lib64:$CUDA_HOME/lib64:$CUDA_HOME/extras/CUPTI/lib64:${LD_LIBRARY_PATH}"
 
+######## CUDA Runtime, needs to be verified if needed
+# ENV NV_NVTX_VERSION=11.8.86-1 \
+#     NV_LIBNPP_VERSION=11.8.0.86-1 \
+#     NV_LIBCUBLAS_VERSION=11.11.3.6-1 \
+#     NV_LIBNCCL_PACKAGE_VERSION=2.15.5-1+cuda11.8
 
-ENV NV_NVTX_VERSION=11.8.86-1 \
-    NV_LIBNPP_VERSION=11.8.0.86-1 \
-    NV_LIBCUBLAS_VERSION=11.11.3.6-1 \
-    NV_LIBNCCL_PACKAGE_VERSION=2.15.5-1+cuda11.8
+# RUN dnf config-manager \
+#        --add-repo https://developer.download.nvidia.com/compute/cuda/repos/rhel9/x86_64/cuda-rhel9.repo \
+#     && dnf install -y \
+#         cuda-libraries-11-8-${NV_CUDA_LIB_VERSION} \
+#         cuda-nvtx-11-8-${NV_NVTX_VERSION} \
+#         libnpp-11-8-${NV_LIBNPP_VERSION} \
+#         libcublas-11-8-${NV_LIBCUBLAS_VERSION} \
+#         libnccl-${NV_LIBNCCL_PACKAGE_VERSION} \
+#     && dnf clean all
+
+## CUDA Development ############################################################
+FROM cuda-base as cuda-devel
+
+# Ref: https://developer.nvidia.com/nccl/nccl-legacy-downloads
+ENV NV_CUDA_CUDART_DEV_VERSION=12.1.55-1 \
+    NV_NVML_DEV_VERSION=12.1.55-1 \
+    NV_LIBCUBLAS_DEV_VERSION=12.1.0.26-1 \
+    NV_LIBNPP_DEV_VERSION=12.0.2.50-1 \
+    NV_LIBNCCL_DEV_PACKAGE_VERSION=2.18.3-1+cuda12.1
 
 RUN dnf config-manager \
        --add-repo https://developer.download.nvidia.com/compute/cuda/repos/rhel9/x86_64/cuda-rhel9.repo \
     && dnf install -y \
-        cuda-libraries-11-8-${NV_CUDA_LIB_VERSION} \
-        cuda-nvtx-11-8-${NV_NVTX_VERSION} \
-        libnpp-11-8-${NV_LIBNPP_VERSION} \
-        libcublas-11-8-${NV_LIBCUBLAS_VERSION} \
-        libnccl-${NV_LIBNCCL_PACKAGE_VERSION} \
-    && dnf clean all
-
-ENV NV_CUDA_CUDART_DEV_VERSION=11.8.89-1 \
-    NV_NVML_DEV_VERSION=11.8.86-1 \
-    NV_LIBCUBLAS_DEV_VERSION=11.11.3.6-1 \
-    NV_LIBNPP_DEV_VERSION=11.8.0.86-1 \
-    NV_LIBNCCL_DEV_PACKAGE_VERSION=2.15.5-1+cuda11.8
-
-RUN dnf config-manager \
-       --add-repo https://developer.download.nvidia.com/compute/cuda/repos/rhel9/x86_64/cuda-rhel9.repo \
-    && dnf install -y \
-        cuda-command-line-tools-11-8-${NV_CUDA_LIB_VERSION} \
-        cuda-libraries-devel-11-8-${NV_CUDA_LIB_VERSION} \
-        cuda-minimal-build-11-8-${NV_CUDA_LIB_VERSION} \
-        cuda-cudart-devel-11-8-${NV_CUDA_CUDART_DEV_VERSION} \
-        cuda-nvml-devel-11-8-${NV_NVML_DEV_VERSION} \
-        libcublas-devel-11-8-${NV_LIBCUBLAS_DEV_VERSION} \
-        libnpp-devel-11-8-${NV_LIBNPP_DEV_VERSION} \
+        cuda-command-line-tools-12-1-${NV_CUDA_LIB_VERSION} \
+        cuda-libraries-devel-12-1-${NV_CUDA_LIB_VERSION} \
+        cuda-minimal-build-12-1-${NV_CUDA_LIB_VERSION} \
+        cuda-cudart-devel-12-1-${NV_CUDA_CUDART_DEV_VERSION} \
+        cuda-nvml-devel-12-1-${NV_NVML_DEV_VERSION} \
+        libcublas-devel-12-1-${NV_LIBCUBLAS_DEV_VERSION} \
+        libnpp-devel-12-1-${NV_LIBNPP_DEV_VERSION} \
         libnccl-devel-${NV_LIBNCCL_DEV_PACKAGE_VERSION} \
     && dnf clean all
 
 ENV LIBRARY_PATH="$CUDA_HOME/lib64/stubs"
 
+<<<<<<< HEAD
 RUN dnf install -y python3.11 git \
     && ln -s /usr/bin/python3.11 /bin/python \
     && python -m ensurepip --upgrade \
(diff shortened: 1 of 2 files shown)


jbusche:
I built the image and deployed it on both FIPS and a non-FIPS cluster, it worked well with no error messages on the FIPS cluster.  All the old python3.9 CVEs are cleaned up, though there's a new medium python3.11 CVE CVE-2024-34064 that's popped up... taking a closer look.

build/Dockerfile
```diff
@@ -12,38 +12,33 @@
...
-COPY .git .git
-COPY pyproject.toml pyproject.toml
-# build wheel if wheel version is empty else download the wheel from PyPi
-RUN if [[ -z "${WHEEL_VERSION}" ]]; \
-    then python -m build --wheel --outdir /tmp; \
-    else pip download fms-hf-tuning==${WHEEL_VERSION} --dest /tmp --only-binary=:all: --no-deps; \
-    fi && \
-    ls /tmp/*.whl >/tmp/bdist_name
+## Global Args #################################################################
+ARG BASE_UBI_IMAGE_TAG=9.3-1552
```
anhuong, line 17: Do we want to set the base image to such a defined level? If we just tag with `9.3` we can get patch updates. But they have already released 9.4 and currently we are taking latest. So this means we would have to actively update the base image versus picking up the new one automatically

build/Dockerfile
```diff
@@ -12,38 +12,33 @@
...
+ARG PROTOC_VERSION=25.2
```
anhuong, line 18: Are we using protoc?

build/Dockerfile
```diff
@@ -12,38 +12,33 @@
...
+ARG PYTORCH_INDEX="https://download.pytorch.org/whl"
 
+# match PyTorch version that was used to compile flash-attention v2 pre-built wheels
+# e.g. flash-attn v2.5.2 => torch ['1.12.1', '1.13.1', '2.0.1', '2.1.2', '2.2.0', '2.3.0.dev20240126']
+# https://github.com/Dao-AILab/flash-attention/blob/v2.5.2/.github/workflows/publish.yml#L47
+# use nightly build index for torch .dev pre-release versions
+ARG PYTORCH_VERSION=2.2.1
```
anhuong, line 25: will this cause issues if the version of flash-attn and torch that we install as dependent packages upgrades? Since we don't have them currently set to a specific version, but this is setting to a specific version?

build/Dockerfile
```diff
@@ -72,45 +72,132 @@ ENV CUDA_HOME="/usr/local/cuda" \
     PATH="/usr/local/nvidia/bin:${CUDA_HOME}/bin:${PATH}" \
     LD_LIBRARY_PATH="/usr/local/nvidia/lib:/usr/local/nvidia/lib64:$CUDA_HOME/lib64:$CUDA_HOME/extras/CUPTI/lib64:${LD_LIBRARY_PATH}"
 
+######## CUDA Runtime, needs to be verified if needed
```
anhuong, line 75: probably not needed?

build/Dockerfile
```diff
@@ -72,45 +72,132 @@ ENV CUDA_HOME="/usr/local/cuda" \
...
+ARG PYTORCH_VERSION
+ARG PYTHON_VERSION
+ARG MINIFORGE_VERSION=23.11.0-0
+
+# consistent arch support anywhere we compile CUDA code
+ENV TORCH_CUDA_ARCH_LIST="8.0;8.6+PTX;8.9"
+
+RUN dnf install -y unzip git ninja-build && dnf clean all
+
+RUN curl -fsSL -v -o ~/miniforge3.sh -O  "https://github.com/conda-forge/miniforge/releases/download/${MINIFORGE_VERSION}/Miniforge3-$(uname)-$(uname -m).sh" && \
```
anhuong, line 129: So this adds conda as a new dependency, I don't know if we need it since the environment is already isolated within the image but it doesn't hurt either to have a virtual env and additional package manager framework

build/Dockerfile
```diff
@@ -72,45 +72,132 @@ ENV CUDA_HOME="/usr/local/cuda" \
...
+
+# Install specific version of torch
+RUN pip install ninja==1.11.1.1 --no-cache-dir
+RUN pip install packaging --no-cache-dir
+RUN pip install torch==$PYTORCH_VERSION+cu121 --index-url "${PYTORCH_INDEX}/cu121" --no-cache-dir
+
+
+## Build flash attention v2 ####################################################
+FROM python-builder as flash-att-v2-builder
+ARG FLASH_ATT_VERSION=v2.5.8
```
anhuong, line 147: Can you explain how you are installing flash-attn versus how we are currently installing it and the differences? Currently we are running `pip install -e ".[flash-attn]"` which is running below, is that insufficient to install it?

build/Dockerfile
```diff
@@ -72,45 +72,132 @@ ENV CUDA_HOME="/usr/local/cuda" \
...
+# `pip` is installed in the venv here
+ENV PATH=/opt/tuning/bin:$PATH
+
+# Install flash attention v2 from the cache build
+RUN --mount=type=bind,from=flash-att-v2-cache,src=/usr/src/flash-attention-v2,target=/usr/src/flash-attention-v2 \
+    pip install /usr/src/flash-attention-v2/*.whl --no-cache-dir
+
+WORKDIR /tmp
+
+RUN dnf install -y python${PYTHON_VERSION} git \
```
anhuong, line 199: Why do we have to python, does it not already exist from `python-builder` layer from installing conda?

build/Dockerfile
```diff
@@ -130,10 +217,26 @@ RUN --mount=type=cache,target=/root/.cache/pip \
...
+
+# Copy in the full python environment
+COPY --from=python-installations /opt/tuning /opt/tuning
+
+ENV PATH=/opt/tuning/bin:$PATH
+
+# Print a list of all installed packages and versions
+RUN pip list -v --disable-pip-version-check --no-python-version-warning
+
+ENV HOME=/home/tuning
```
anhuong, line 235: Does this mean the working directory is `/home/tuning`? Should the scripts be copied into this directory instead of `/app`? And what are the implications, I know when I've run before I've hit into issue with creating the `.aim` directory and the output directory in `/tmp`. Since you removed the ownership of those dirs to not include the new user, are those directories still able to be created?

build/Dockerfile
```diff
@@ -12,38 +12,33 @@
...
+WORKDIR /app
+
+ARG PYTHON_VERSION
 
 RUN dnf update -y \
     && dnf remove -y --disableplugin=subscription-manager \
         subscription-manager \
         # we install newer version of requests via pip
-        python3.11-requests \
+    python${PYTHON_VERSION}-requests \
```
tedhtchang, line 41: dnf can not find the pthon3.11-requests package.
```
docker build --target base --progress=plain -t fms-hf-tuning:mytag . -f build/Dockerfile
...
No match for argument: python3.11-requests
```

tedhtchang: We could probably remove the installation of `make` and `procps` to further secure the final image since they are for development and debug.

tedhtchang (commented):
This looks promissing. I tested with a  local tuning job on KinD. The image size also reduced from 12G to 6G.  I think we can focus on removing the bits that don't apply to to this project.

tedhtchang:
To address the comments I made these changes:
1. Install python dependencies from repo's pyproject.toml for consistency 
2. Single python env installed by the OS (as oppose adding miniconda)
3. Install all python packages in user site and then copy over to the final stage

jbusche:
Hey @tedhtchang 
I took a look at it.
1.  I built it the image on my OC cluster:
```
git clone https://github.com/tharapalanivel/fms-hf-tuning.git -b multi_stage_docker
cd fms-hf-tuning/
docker build --progress=plain -t fms-hf-tuning:ted-0521-pr154 . -f build/Dockerfile
```
and it ends up (on OpenShift) looking at half the size of the main image:
```
podman images |grep localhost
localhost/fms-hf-tuning                                                                           ted-0521-pr154   191482d03e68  19 minutes ago  6.4 GB
localhost/fms-hf-tuning                                                                           jim-0521-nocves  80e609673e60  8 hours ago     12.7 GB
localhost/fms-hf-tuning                                                                           jim-main-0521    596f53a3df46  8 hours ago     12.5 GB
```
I tried running it with the quick twitter test and it looked good:
```
oc get pytorchjobs,pods
NAME                                   STATE       AGE
pytorchjob.kubeflow.org/ted-kfto-sft   Succeeded   12m

NAME                        READY   STATUS      RESTARTS   AGE
pod/ted-kfto-sft-master-0   0/1     Completed   0          12m
```
I did a twistlock scan and it solves all my python3.9 problems (for some reason) except for one:
```
cve | link | hasFix | status | packageType | packageName | packageVersion | packageLicense | packageBinaryPkgs | packagePath | description
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
CVE-2022-40897 | https://nvd.nist.gov/vuln/detail/CVE-2022-40897 | Y | fixed in 65.5.1 | python | setuptools | 53.0.0 | UNKNOWN |   | /usr/lib/python3.9/site-packages/setuptools-53.0.0.dist-info | Python Packaging Authority (PyPA) setuptools before 65.5.1 allows remote   attackers to cause a denial of service via HTML in a crafted package or   custom PackageIndex page. There is a Regular Expression Denial of Service   (ReDoS) in package_index.py.
```

I think if this PR works with the various GPU/LoRA testing, it might be a great improvement.

FYI, on Quay, the sizes don't match the size I see on OpenShift.  For example: https://quay.io/repository/jbusche/fms-hf-tuning?tab=tags
```
ted-0521-pr154 3.3 GiB
jim-0521-nocves 6.5 GiB
jim-main-0521 6.3 GiB
```

build/Dockerfile
```diff
@@ -148,12 +149,9 @@ RUN touch /.aim_profile && \
     mkdir /.cache && \
     chmod -R 777 /.cache
 
-# create tuning user and give ownership to dirs
-RUN useradd -u $USER_UID tuning -m -g 0 --system && \
```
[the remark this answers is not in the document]
tedhtchang: I moved this line to the [base layer](https://github.com/tharapalanivel/fms-hf-tuning/blob/ee87be95b303cb868af7052b4149c6266a7d41c8/build/Dockerfile#L42) because the [python-installation layer](https://github.com/tharapalanivel/fms-hf-tuning/blob/ee87be95b303cb868af7052b4149c6266a7d41c8/build/Dockerfile#L93-L94) also needs to install python packages in the user site.

build/Dockerfile
```diff
@@ -12,58 +12,51 @@
...
-FROM registry.access.redhat.com/ubi9/ubi AS release
-
-ARG CUDA_VERSION=11.8.0
+## Global Args #################################################################
+ARG BASE_UBI_IMAGE_TAG=latest
+ARG BASE_UBI=registry.access.redhat.com/ubi9/ubi
+ARG PYTORCH_INDEX="https://download.pytorch.org/whl"
 ARG USER=tuning
 ARG USER_UID=1000
-ARG SET_NUM_PROCESSES_TO_NUM_GPUS=True
```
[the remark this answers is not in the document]
tedhtchang: I think ARG should be assigned to an `ENV SET_NUM_PROCESSES_TO_NUM_GPUS=${SET_NUM_PROCESSES_TO_NUM_GPUS}`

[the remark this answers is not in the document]
anhuong: Yup I agree that the default in the image is already set to true so this isn't needed. It's a way of building the image with a different setting but since it's an env var that can easily be set at runtime, not needed at build time.

jbusche:
This is looking good to me @tedhtchang and @tharapalanivel.
I built the latest and pushed it to quay, testing on devstage and it's looking good.  For example, a single-gpu LoRA test:
```
oc get pytorchjob |grep jim
jim-test-image-1        Succeeded   3m4s
jamesbusche@jamess-mbp AHN % oc get pods |grep jim                
jim-test-image-1-master-0   0/1     Completed   0          3m8s
```
Size still looks good too...
```
podman images
REPOSITORY                                                                                     TAG               IMAGE ID      CREATED            SIZE
localhost/fms-hf-tuning                                                                        ted-0523-pr1543   127dbbc1a47b  About an hour ago  6.42 GB
```

build/Dockerfile
```diff
@@ -72,88 +64,91 @@ ENV CUDA_HOME="/usr/local/cuda" \
...
-    && python -m ensurepip --upgrade \
-    && dnf clean all
+FROM cuda-devel as python-installations
 
-# Removes the example private key to avoid high severity vulnerability warning
-RUN rm -f /usr/share/doc/perl-Net-SSLeay/examples/server_key.pem
+ARG WHEEL_VERSION
+ARG USER
 
+USER ${USER}
```
anhuong, line 97: Interesting that python installs are all happening as USER instead of as root. Previously, we've built everything as root and only at the final stage would we run the image as root. Would this pose any issues if someone runs the image as a different user? I wouldn't think so but want to check

pyproject.toml
```diff
@@ -38,10 +38,11 @@ dependencies = [
 "datasets>=2.15.0",
 "fire",
 "simpleeval",
+"packaging",
 ]
 
 [project.optional-dependencies]
-dev = ["wheel", "packaging", "ninja", "scikit-learn>=1.0, <2.0", "boto3"]
```
anhuong, line 44: Is packaging not only used in the building of the wheel and thus is needed as a required dependency?

anhuong (commented):
Thank you Ted for the changes! They look good, a few questions and please ensure that you or Jim do thorough testing with the new image running single/multi-GPU fine tuning and single/multi-GPU with LoRA. I see Jim already did single GPU LoRA on an Openshift cluster which is excellent!

build/Dockerfile
```diff
@@ -12,58 +12,50 @@
...
-        # to help with debugging
-        procps \
+    && dnf remove -y --disableplugin=subscription-manager subscription-manager \
+    && dnf install -y python${PYTHON_VERSION} git \
+    && ln -s /usr/bin/python${PYTHON_VERSION} /bin/python \
+    && python -m ensurepip --upgrade \
     && dnf clean all
-
+# Resolve Twistlock scan H severity: Private keys stored in image
+RUN rm -f /usr/share/doc/perl-Net-SSLeay/examples/server_key.pem
```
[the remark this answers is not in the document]
tedhtchang: I will check. The CVE might have been fixed in the upstream ubi9 image.
It came from a CVE fix in this [PR](https://github.com/foundation-model-stack/fms-hf-tuning/pull/127/files#diff-5aecd73bbcb9b7410938673269f0d4e676dd18509277b0ff2f2787c9d3e22604R118).

build/Dockerfile
tedhtchang: The tuning user is in the root group (-g 0), see [here](https://github.com/tharapalanivel/fms-hf-tuning/blob/e7a47f8e4598690ee663e7406baa7344e6900413/build/Dockerfile). Any user in the root group can access python packages installed by this user.
Unless someone specify non-zero `runAsGroup` in [scc](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-pod) or openshift runs pod as random user from root group. For example, when I log in the runing pod I ses:
```
bash-5.1$ whoami
1000690000
bash-5.1$ cat /etc/passwd
root:x:0:0:root:/root:/bin/bash
...
tuning:x:1000:0::/home/tuning:/bin/bash
1000690000:x:1000690000:0:1000690000 user:/:/sbin/nologin
```

pyproject.toml
tedhtchang: It's a dependency of the accelerate lib. It should be installed automatically with the lib. So I guess this change is not required. let me check...

build/Dockerfile
```diff
@@ -12,58 +12,50 @@
...
+ARG USER_UID
 
 RUN dnf update -y \
-    && dnf remove -y --disableplugin=subscription-manager \
-        subscription-manager \
-        # we install newer version of requests via pip
-        python3.11-requests \
-    && dnf install -y make \
-        # to help with debugging
-        procps \
```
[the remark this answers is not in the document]
tedhtchang: procps is the `ps -ef` command for listing current running processes. I didn't see `ps` or `make` used by anything. I don't problem adding them back because they are useful for development.

Alternatively, we can add additional stage to install any dev tools
```
...
FROM release as devleopment
dnf install -y make procps git 
```
For release image, `docker build --targt=release -t sft-trainer:release  .  -f build/Dockerfile`
For developmen image `docker build --targt=development -t sft-trainer:dev .  -f build/Dockerfile`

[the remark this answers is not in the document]
anhuong: I agree that a development image would be helpful. It should also contain the fms-hf-tuning library and all deps needed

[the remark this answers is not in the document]
anhuong: I wonder if `ps` is okay to leave in the final image if it is helpful for seeing multi-GPU processes running and such

build/Dockerfile
anhuong: So as long as the user that is running the image is in the root group we should be good, right? In which when the image is run on Openshift, it defaults to using the tuning user correct? Or does it default to a random user?

tedhtchang: > as long as the user that is running the image is in the root group we should be good, right?

correct as long as user is in root group we are good.

> In which when the image is run on Openshift, it defaults to using the tuning user correct? Or does it default to a random user?

User should be random But In our OpenShift, the user default to the id in the namespace CR. For example 
```bash
oc get ns fmaas-dev -oyaml
apiVersion: v1
kind: Namespace
metadata:
  name: fmaas-dev
  annotations:
    ...
    openshift.io/sa.scc.supplemental-groups: 1000690000/10000
    openshift.io/sa.scc.uid-range: 1000690000/10000
    ...
spec:
  finalizers:
  - kubernetes
status:
  phase: Active
```

build/Dockerfile
```diff
@@ -12,57 +12,58 @@
...
-ENV CUDA_VERSION=$CUDA_VERSION \
-    NV_CUDA_LIB_VERSION=11.8.0-1 \
+RUN useradd -u $USER_UID ${USER} -m -g 0 --system && \
+    chmod g+rx /home/${USER}
+
+## Used as base of the Release stage to removed unrelated the packages and CVEs
+FROM base as release-base
+
+# Removes the python3.9 code to eliminate possible CVEs.  Also removes dnf
+RUN rpm -e $(dnf repoquery python3-* -q --installed) dnf python3 yum crypto-policies-scripts
```
tedhtchang, line 48: @jbusche can you verify the change still works as intended ?

build/Dockerfile
```diff
@@ -12,57 +12,58 @@
...
         subscription-manager \
-        # we install newer version of requests via pip
-        python3.11-requests \
-    && dnf install -y make \
-        # to help with debugging
-        procps \
+    && dnf install -y python${PYTHON_VERSION} procps \
+    && ln -s /usr/bin/python${PYTHON_VERSION} /bin/python \
+    && python -m ensurepip --upgrade \
+    && dnf update -y \
```
tedhtchang, line 35: @jbusche can you verify the change still works as intended ?

build/Dockerfile
jbusche: Hey @tedhtchang, yes I'm not seeing any fixable Twistlock CVEs with the above.

build/Dockerfile
jbusche: @tedhtchang, looks to me that the change was to install `procps`?  But I am now seeing in the https://quay.io/repository/jbusche/fms-hf-tuning?tab=tags that the quay security scans are green (passed)

I've done a quick test of the image with this on my small OpenStack 4.14 cluster:
https://github.com/foundation-model-stack/fms-hf-tuning/wiki/Installing-and-Testing-OpenShift-fms%E2%80%90hf%E2%80%90tuning-Stack#6-testing
and it looks good.

On Devstage, I also did a `single GPU twitter fine-tuning` https://github.com/foundation-model-stack/fms-hf-tuning/wiki/SFT%E2%80%90Trainer%E2%80%90Image-Test-process#single-gpu-twitter-fine-tuning
which looked good.

I also did a `single GPU llama fine-tuning` https://github.com/foundation-model-stack/fms-hf-tuning/wiki/SFT%E2%80%90Trainer%E2%80%90Image-Test-process#single-gpu-llama-fine-tuning
Which looked good.

Lastly I tried a `single GPU llama LoRA tuning` https://github.com/foundation-model-stack/fms-hf-tuning/wiki/SFT%E2%80%90Trainer%E2%80%90Image-Test-process#single-gpu-llama-lora-tuning
and that looked good.

I can't get a job with 2 gpu to get out of pending though, so that's all the testing I can currently try.

tedhtchang: Thanks @jbusche This is great. Could you approve

anhuong:
Do we want to wait to merge this until after the 0.2.0 release?

build/Dockerfile
anhuong: Nice testing Jim! I think we will have to sort out the wiki though as we are trying to consolidate the JSON configs. So there should be a better place for those configs. I have some PRs in flight already across our repos and working to make the Lora vs FT vs PT configs more clear.

anhuong (approved):
Awesome work everyone! Thank you so much!

PR #156, opened 2024-05-14
feat: History based loss metric

seshapad:
### Description of the change
Following are four stopping criteria scenarios that supported in this metric:

    SC1: Stop when training loss does not decrease over time
    SC2: Stop when training loss is below a threshold or above a threshold
    SC3: Stop when training loss is below a threshold after certain number of epochs
    SC4: Stop when eval loss is below a threshold after certain number of epochs

### Related issue number

Closes #690

### How to verify the PR

Add trainer controller configurations for the above scenarios.

### Was the PR tested

image


tuning/trainercontroller/operations/patience_hfcontrols.py
```diff
@@ -0,0 +1,56 @@
...
+        """
+        self._patience_counter += 1
+        if self._patience_counter <= self._patience_threshold:
+            logger.info(
+                "Patience counter %d is within patience threshold of %d",
+                self._patience_counter,
+                self._patience_threshold,
+            )
+            return
+        logger.info("Patience counter exceeded threshold")
```
ashokponkumar, line 52: Lets make it more informative.

tuning/trainercontroller/operations/patience_hfcontrols.py
```diff
@@ -0,0 +1,56 @@
...
+from transformers import TrainerControl
+from transformers.utils import logging
+
+# Local
+from .operation import Operation
+
+logger = logging.get_logger(__name__)
+
+
+class HFControlsWithPatience(Operation):
```
ashokponkumar, line 16: Why not enhance HFControls to support patience too, instead of adding a new class?

tuning/trainercontroller/operations/patience_hfcontrols.py
```diff
@@ -0,0 +1,56 @@
...
+
+    def control_action(self, control: TrainerControl, **kwargs):
+        """This method peeks into the stack-frame of the caller to get the action the triggered
+        a call to it. Using the name of the action, the value of the control is set.
+
+        Args:
+            control: TrainerControl. Data class for controls.
+            kwargs: List of arguments (key, value)-pairs
+        """
+        self._patience_counter += 1
```
ashokponkumar, line 44: Where are we resetting the patience counter? Doesn't the patience increment need to be continuous? If it needs to be continuous how are we going to capture that since control action will be invoked only when the rule succeeds?

ashokponkumar (changes requested):
Thanks Nabha. Have added reviews for the patience score implementation. Will look at the history next.

tuning/trainercontroller/controllermetrics/history_based_metrics.py
```diff
@@ -0,0 +1,97 @@
...
+    _WINDOW_SIZE = "window-size"
+
+    def __init__(self, **kwargs):
+        """Initializes the metric handler, by registering the event \
+            list and arguments with base handler.
+
+        Args:
+            kwargs: List of arguments (key, value)-pairs
+        """
+        self._window_size = kwargs.get("window-size")
```
ashokponkumar, line 46: Why not make it as part of __init__ signature?

tuning/trainercontroller/controllermetrics/history_based_metrics.py
```diff
@@ -0,0 +1,97 @@
...
+        self._window = {
+            self._TRAINING_LOSS_KEY: deque(),
+            self._METRICS_KEY: deque(),
+            self._WINDOW_SIZE: self._window_size,
+        }
+        if self._window_size is None:
```
ashokponkumar, line 52: Use function default params to initialize

tuning/trainercontroller/controllermetrics/history_based_metrics.py
```diff
@@ -0,0 +1,97 @@
...
+            self._add_to_window(self._METRICS_KEY, kwargs[self._METRICS_KEY])
+        else:
+            size_of_log_history = len(state.log_history)
+            for i in range(size_of_log_history - 1, -1, -1):
+                log = state.log_history[i]
+                if self._TRAINING_LOSS_KEY in log:
+                    self._add_to_window(
+                        self._TRAINING_LOSS_KEY, float(log[self._TRAINING_LOSS_KEY])
+                    )
+                    break
```
ashokponkumar, line 96: Are these no other events other than loss and eval_loss that can occur in log history?

tuning/trainercontroller/controllermetrics/history_based_metrics.py
```diff
@@ -0,0 +1,97 @@
...
+
+        Args:
+            state: TrainerState object
+            kwargs: Remaining event arguments
+
+        Returns:
+            Any. The exposed variables are returned here.
+        """
+        if self._METRICS_KEY in kwargs:
+            self._add_to_window(self._METRICS_KEY, kwargs[self._METRICS_KEY])
```
ashokponkumar, line 87: Wont' this result in an always increasing array? There seems to be no implementation of sliding. For example, slide function is not called.

ashokponkumar (changes requested):
Added some comments on History metric as well.

tuning/trainercontroller/operations/patience_hfcontrols.py
seshapad: Done.

tuning/trainercontroller/operations/patience_hfcontrols.py
seshapad: Done.

tuning/trainercontroller/controllermetrics/history_based_metrics.py
```diff
@@ -0,0 +1,97 @@
...
+from transformers import TrainerState
+from transformers.utils import logging
+
+# Local
+from tuning.trainercontroller.controllermetrics.metricshandler import MetricHandler
+
+logger = logging.get_logger(__name__)
+
+
+class HistoryBasedMetric(MetricHandler):
```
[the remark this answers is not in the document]
seshapad: we dump the entire metrics dictionary from `on_evaluate`...this might contain metrics other than loss.

tuning/trainercontroller/operations/patience_hfcontrols.py
```diff
@@ -0,0 +1,56 @@
...
+        """Initializes the HuggingFace controls. In this init, the fields with `should_` of the
+        transformers.TrainerControl data class are extracted, and for each of those fields, the
+        control_action() method's pointer is set, and injected as a class member function.
+
+        Args:
+            kwargs: List of arguments (key, value)-pairs
+        """
+        self.kwargs = kwargs
+        self._patience_counter = 0
+        self._patience_threshold = kwargs["patience-threshold"]
```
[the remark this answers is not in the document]
seshapad: taken care [here](https://github.com/foundation-model-stack/fms-hf-tuning/pull/156#discussion_r1607920173).

tuning/trainercontroller/controllermetrics/history_based_metrics.py
seshapad: Done

tuning/trainercontroller/controllermetrics/history_based_metrics.py
seshapad: Addressed.

tuning/trainercontroller/controllermetrics/history_based_metrics.py
seshapad: Addressed.

tuning/trainercontroller/controllermetrics/history_based_metrics.py
seshapad: we could dump the training state as well...what are your thoughts?

tuning/trainercontroller/operations/patience_hfcontrols.py
seshapad: Addressed.

examples/trainercontroller_configs/epoch-level-eval-loss-patience.yaml
```diff
@@ -0,0 +1,19 @@
...
+operations:
+  - name: hfwithpatience
+    class: HFControls
+    arguments:
+      patience-threshold: 3
+controllers:
+  - name: epoch-level-eval-loss-patience
+    triggers:
+      - on_epoch_end
+    rule: len(eval_loss_window["metrics"]) > 0 and eval_loss_window["metrics"][-1]["eval_loss"] > 2.2 and trainer_state["epoch"] > 2
```
ashokponkumar, line 17: isn't patience an attribute at this level instead of at the control level? This allow us to not having to keep the logic of checking if a rule succeeded before invoking a control.

ashokponkumar (changes requested):
Some changes requested.

examples/trainercontroller_configs/epoch-level-eval-loss-patience.yaml
seshapad: Addressed.

tuning/trainercontroller/controllermetrics/history_based_metrics.py
```diff
@@ -0,0 +1,95 @@
...
+        """Initializes the metric handler, by registering the event \
+            list and arguments with base handler.
+
+        Args:
+            kwargs: List of arguments (key, value)-pairs
+        """
+        self._window = {
+            TRAINING_LOSS_KEY: deque(),
+            METRICS_KEY: deque(),
+            WINDOW_SIZE: window_size,
```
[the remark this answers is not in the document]
seshapad: Addressed.

tuning/trainercontroller/controllermetrics/history_based_metrics.py
```diff
@@ -0,0 +1,95 @@
...
+    def _add_to_window(self, data_type, data):
+        self._window[data_type].append(data)
+
+    def _slide_the_window(self, data_type):
+        if len(self._window[data_type]) < self._window[WINDOW_SIZE]:
+            return False
+        if len(self._window[data_type]) == self._window[WINDOW_SIZE]:
+            return True
+        self._window[data_type].popleft()
+        return True
```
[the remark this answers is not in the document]
seshapad: Addressed.

examples/trainercontroller_configs/epoch-level-eval-loss-patience.yaml
```diff
@@ -0,0 +1,14 @@
...
+    class: HistoryBasedMetric
+    arguments:
+      window-size: 2
+controllers:
+  - name: epoch-level-eval-loss-patience
+    triggers:
+      - on_epoch_end
+    rule: len(eval_loss_window["metrics"]) > 0 and eval_loss_window["metrics"][-1]["eval_loss"] > 2
+    patience:
+      patience-threshold: 2
```
ashokponkumar, line 12: Is `patience_threshold` the right one or is `patience-threshold` the right one?

tests/data/trainercontroller/epoch-level-eval-loss-patience.yaml
```diff
@@ -0,0 +1,14 @@
...
+    class: HistoryBasedMetric
+    arguments:
+      window-size: 2
+controllers:
+  - name: epoch-level-eval-loss-patience
+    triggers:
+      - on_epoch_end
+    rule: len(eval_loss_window["metrics"]) > 0 and eval_loss_window["metrics"][-1]["eval_loss"] > 2
+    patience:
+      patience_threshold: 2
```
ashokponkumar, line 12: Do we want to call it `patience_threshold` or `threshold` since patience is already part of the parent key

ashokponkumar (changes requested):

examples/trainercontroller_configs/epoch-level-eval-loss-patience.yaml
seshapad: addressed.

tests/data/trainercontroller/epoch-level-eval-loss-patience.yaml
seshapad: let it be `patience_threshold`. Although redundant, it will be less confusing to user to relate to the concept...isnt it?

ashokponkumar: Sure, But we should ensure we use either `-` or `_` consistently. We use `-` for `window-size` and `controller-metrics`, but we are using `_` here.

ashokponkumar (changes requested):

tuning/trainercontroller/patience.py
```diff
@@ -0,0 +1,54 @@
...
+    def __init__(self, patience_threshold=1, mode=MODE_RESET_ON_FAILURE, **kwargs):
+        self._patience_threshold = patience_threshold
+        self._patience_counter = 0
+        self._mode = mode
+
+    def should_tolerate(self, rule_outcome: bool) -> bool:
+        if rule_outcome:
+            self._patience_counter = self._patience_counter + 1
+        else:
+            if self._mode == MODE_RESET_ON_FAILURE:
```
ashokponkumar, line 39: Why not use `elif`?

tuning/trainercontroller/patience.py
```diff
@@ -0,0 +1,54 @@
...
+                self._patience_counter = 0
+        if self._patience_counter <= self._patience_threshold:
+            logger.info(
+                "Enforcing patience [patience_counter = {:.2f}, \
```
ashokponkumar, line 43: Shouldn't this be just a `debug` level message?

tuning/trainercontroller/patience.py
```diff
@@ -0,0 +1,54 @@
...
+                    self._patience_counter, self._patience_threshold
+                )
+            )
+            return True
+        logger.info(
+            "Exceeded patience [patience_counter = {:.2f}, patience_threshold = {:.2f}]".format(
+                self._patience_counter, self._patience_threshold
+            )
+        )
+        return False
```
ashokponkumar, line 54: Why is `self._patience_counter` not reset after it gets triggered?

tuning/trainercontroller/patience.py
```diff
@@ -0,0 +1,54 @@
...
+        if self._patience_counter <= self._patience_threshold:
+            logger.info(
+                "Enforcing patience [patience_counter = {:.2f}, \
+                    patience_threshold = {:.2f}]".format(
+                    self._patience_counter, self._patience_threshold
+                )
+            )
+            return True
+        logger.info(
+            "Exceeded patience [patience_counter = {:.2f}, patience_threshold = {:.2f}]".format(
```
ashokponkumar, line 50: We have to be more elaborate on the controller name and more, else this message will be too uninformative, isn't it?

tuning/trainercontroller/controllermetrics/history_based_metrics.py
```diff
@@ -0,0 +1,94 @@
...
+        """Initializes the metric handler, by registering the event \
+            list and arguments with base handler.
+
+        Args:
+            kwargs: List of arguments (key, value)-pairs
+        """
+        self._window = {
+            TRAINING_LOSS_KEY: deque(),
+            METRICS_KEY: deque(),
+            WINDOW_SIZE: window_size,
```
ashokponkumar, line 48: Why is window_size within _window? shouldn't it be just a base level property?

tuning/trainercontroller/controllermetrics/history_based_metrics.py
```diff
@@ -0,0 +1,94 @@
...
+        Args:
+            state: TrainerState object
+            kwargs: Remaining event arguments
+
+        Returns:
+            Any. The exposed variables are returned here.
+        """
+        if METRICS_KEY in kwargs:
+            self._add_and_slide(METRICS_KEY, kwargs[METRICS_KEY])
+        else:
```
ashokponkumar, line 85: Why is this in else? shouldn't this just check if state is present, and use the data irrespective of whether Metrics key is present or not?

tuning/trainercontroller/controllermetrics/history_based_metrics.py
```diff
@@ -0,0 +1,94 @@
...
+            size_of_log_history = len(state.log_history)
+            for i in range(size_of_log_history - 1, -1, -1):
+                log = state.log_history[i]
+                if TRAINING_LOSS_KEY in log:
```
ashokponkumar, line 89: Why are we not supporting say eval_loss?

tuning/trainercontroller/controllermetrics/history_based_metrics.py
```diff
@@ -0,0 +1,94 @@
...
+    """Implements the controller metric which evaluates loss-per-step"""
+
+    def __init__(self, window_size=1, **kwargs):
+        """Initializes the metric handler, by registering the event \
+            list and arguments with base handler.
+
+        Args:
+            kwargs: List of arguments (key, value)-pairs
+        """
+        self._window = {
```
ashokponkumar, line 45: What should be the behaviour of this metric before enough number of elements equal to the window_size is reached?

ashokponkumar (changes requested):

tuning/trainercontroller/controllermetrics/history_based_metrics.py
seshapad: `window_size` is a property of `_window`. That is why its stored inside.

tuning/trainercontroller/controllermetrics/history_based_metrics.py
seshapad: We are supporting `eval_loss` in the if section. Its part of the `metrics`.
```
        if METRICS_KEY in kwargs:
            self._add_and_slide(METRICS_KEY, kwargs[METRICS_KEY])
```

tuning/trainercontroller/controllermetrics/history_based_metrics.py
seshapad: The metric merely returns the current state of the window and moves the window. It does not do anything more. Its up to the rule to apply an particular operation on the window.

tuning/trainercontroller/controllermetrics/history_based_metrics.py
seshapad: The `if-else` here is to distinguish if the data is a `evaluation metrics` or `training loss`. Each have their own window into which they are added.

tuning/trainercontroller/patience.py
seshapad: Addressed.

tuning/trainercontroller/patience.py
seshapad: Addressed.

tests/data/trainercontroller/epoch-level-eval-loss-patience.yaml
seshapad: Addressed. Except for `controller` name, `-` should not used anywhere.

tuning/trainercontroller/patience.py
seshapad: Addressed.

tuning/trainercontroller/controllermetrics/history_based_metrics.py
ashokponkumar: If we dump the entire object instead of each metric as a list, won't operations like sum and others become hard in the rule?

tuning/trainercontroller/patience.py
```diff
@@ -0,0 +1,60 @@
...
+        if rule_outcome:
+            self._patience_counter = self._patience_counter + 1
+        elif self._mode == MODE_RESET_ON_FAILURE:
+            self._patience_counter = 0
+        if self._patience_counter <= self._patience_threshold:
+            logger.debug(
+                "Enforcing patience [patience_counter = {:.2f}, "
+                "patience_threshold = {:.2f}]".format(
+                    self._patience_counter, self._patience_threshold
+                )
```
ashokponkumar, line 47: This should also have control name and event name, isn't it?

tuning/trainercontroller/patience.py
```diff
@@ -0,0 +1,60 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# SPDX-License-Identifier: Apache-2.0
+# https://spdx.dev/learn/handling-license-info/
+
+# Third Party
+from transformers.utils import logging
+
+MODE_RESET_ON_FAILURE = "reset_on_failure"
```
ashokponkumar, line 21: Dont' we need another mode which does not reset on failure?

tuning/trainercontroller/patience.py
```diff
@@ -0,0 +1,60 @@
...
+        self._mode = mode
+
+    def should_tolerate(
+        self, rule_outcome: bool, event_name=None, control_name=None, **kwargs
+    ) -> bool:
+        if rule_outcome:
+            self._patience_counter = self._patience_counter + 1
+        elif self._mode == MODE_RESET_ON_FAILURE:
+            self._patience_counter = 0
+        if self._patience_counter <= self._patience_threshold:
```
ashokponkumar, line 42: if we do `<=`, then if it hits if the `patience_threshold` is 1, then it waits for the rule to succeed twice before it gets triggered, isn't it? Is that the intended behavior?

tuning/trainercontroller/controllermetrics/history_based_metrics.py
```diff
@@ -0,0 +1,94 @@
...
+        Returns:
+            Any. The exposed variables are returned here.
+        """
+        if METRICS_KEY in kwargs:
+            self._add_and_slide(METRICS_KEY, kwargs[METRICS_KEY])
+        else:
+            size_of_log_history = len(state.log_history)
+            for i in range(size_of_log_history - 1, -1, -1):
+                log = state.log_history[i]
+                if TRAINING_LOSS_KEY in log:
```
ashokponkumar, line 89: Don't we need to check which step this loss is for, and if it is repeating twice? or is it guaranteed to happen only once?

ashokponkumar (changes requested):

tuning/trainercontroller/patience.py
seshapad: Addressed.

tuning/trainercontroller/patience.py
seshapad: Do you have an usecase to support this requirement?

tuning/trainercontroller/patience.py
seshapad: Yes. Upto and including the threshold is to be tolerated. Beyond that is treated as `exceeding patience`. Is that acceptable?

tuning/trainercontroller/controllermetrics/history_based_metrics.py
seshapad: Addressed.

tests/data/trainercontroller/epoch-level-eval-loss-patience.yaml
seshapad: All `-` removed. Only using `_`.

kmehant (approved):
The PR adds support for controlling the training runtime based on the loss. Specific use case scenarios are listed in the PR description. A new patience mechanism is added, tests have been added, naming is fixed to use underscores as a standard, supporting tests and examples have been added.

PR #157, opened 2024-05-15
Install Acceleration Framework into Training Script

fabianlim:
### Description of the change

This PR installs the Acceleration Framework into the `sft_trainer.py`. This is a followup to #119 which proposes to have a lightweight integration; the implementation of Acceleration Framework is kept seperate in the repo [fms-acceleration](https://github.com/foundation-model-stack/fms-acceleration) under the same `foundation-model-stack` org.
- introduce a `AccelerationFrameworkArguments` that accepts an `--acceleration_framework_config_file` argument to configure the framework.
- update `pyproject.toml` with an *optional dependency* `[fms-accel]` that installs the [fms-acceleration framework](https://github.com/foundation-model-stack/fms-acceleration/plugins/framework).
- update `README.md` to include basic usage of `fms-acceleration`, using it to perform accelerated PEFT with a 4bit GPTQ-LoRA.
- ensure that the integration within `sft_trainer.py` is optional, the `.get_framework` call  below will **silently disable** framework if `fms-accel` dependency is not desired.
    ```
    framework = AccelerationFrameworkConfig.from_dataclasses(*dataclass_configs).get_framework()
    ```

- restrict to only *three integration points* within `sft_trainer.py` script:
  1. `framework.model_loader`: load model if `framework.requires_custom_loading == True` 
  2. `framework.augmentation`: load if `framework.requires_agumentation == True`
  3. `framework.get_callbacks_and_ready_for_train`: get callbacks and do final prep on `model` and `trainer.accelerator` if required.

**Update: Picture has been changed to reflect new dataclass arguments flow**

image

### Related issue number

Related to the merged PR #119 for the ADR of Training Acceleration.

### How to verify the PR

This PR can be verified in the following ways:
1. run the new folder `tests/acceleration`, see **Note-for-unit-tests** below.
2. **[Update: this bench needs to be reworked due to YAML arguments being disabled, right now works with some patching]** run the [provided benchmark utility](https://github.com/foundation-model-stack/fms-acceleration/blob/main/scripts/benchmarks/README.md), and checking with the results in the [PR](https://github.com/foundation-model-stack/fms-acceleration/pull/4) and the [a100_80gb](https://github.com/foundation-model-stack/fms-acceleration/blob/main/scripts/benchmarks/refs/a100_80gb.csv) 
   -  consult [this helm chart as a reference](https://gist.github.ibm.com/Aaron-Chew1/f56a53b14849d8c508f35592b05b98c5) to run the bench

**Note for unit tests**
For `tests/acceleration`, note that `testenv` will not install `.fms-accel` dependencies, so some of the newly added tests wil be skipped. So to run *all* the new tests:
```
pip install ".[fms-accel]"
python -m fms_acceleration.cli install peft
pytest tests/acceleration
```

### Was the PR tested


Ssukriti:
from DM - 

It would be great if there was a way to pass a Python dataclass instead of yaml file
`python sft_trainer.py \
    --acceleration_framework_config_file framework.yaml`
    
    
we can create a new GPTQ-LOra config here https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/tuning/config/peft_config.py#L21
so users of fms-tuning have exactly same experience as Lora


Inside fms-tuning, we know for GPTQ-Lora we need to call AccelerationFramework() -> can we convert to a Python dataclass object accepted by AccelerationFramework -AccelerationFramework(qLORAcONFIG ) 


You know what dataclass instance it is so you can parse params from that


we can implement this in steps though instead of requiring too much change to your framework-
Step 1 - users of fms-tuning continue to use dataclasses -> we will have to internally convert params passed to the yaml file you want (we wont expose yaml to users of tuning, but keep it inside tuning library ) -> call accelerationframework same way with yaml
step 2- > acceleration framework adds support for Python Dataclasses for each plugin as well besides yaml , so we dont have to create yaml but can just pass params needed for a certain plugin (edited) 

if we do step 1, we can merge code and get it working, while working on step 2 as enhancement


We are discussing if we need 1 dataclass or multiple, but we can start with step 1 with dataclass only in tuning and see how it goes

fixtures/accelerated-peft-autogptq-sample-configuration.yaml
```diff
@@ -0,0 +1,24 @@
...
+
+        # Kernel to be used for GPTQ linear laeyer
+        # NOTE: Not all kernels are suitable for PEFT training; need to use 
+        # kernels that support autograd forward / backward. The best 
+        # recommendation at the moment is "triton_v2".
+        kernel: triton_v2
+
+        # If true, then will already expect quantized checkpoint 
+        # passed into TrainingArguments.model_name_or_path
+        from_quantized: true
```
Ssukriti, line 24: will from_quantized: false, even work ? it is not supported currently right? so wondering if we should remove this as option in YAML and harcode code True, otherwise throw error?

tuning/sft_trainer.py
```diff
@@ -248,6 +277,11 @@ def train(
             "Validation dataset length is %s", len(formatted_validation_dataset)
         )
 
+    if framework is not None and framework.requires_agumentation:
+        model, (peft_config,) = framework.augmentation(
+            model, train_args, modifiable_args=(peft_config,)
```
Ssukriti, line 282: @fabianlim what will happen if `peft_config` is Ptconfig or None? only LoraConfig is supported by the framework currently?

fixtures/accelerated-peft-autogptq-sample-configuration.yaml
fabianlim: yes, this is a flag supported in the AutoGPTQ package, such that you can take a non-quantized model, and quantize it, then start training. 

But we disable this and we actually do have a [check in our plugin](https://github.com/foundation-model-stack/fms-acceleration/blob/79bc89b3ce15d804e826d7c98e11bde0182fa7a4/plugins/accelerated-peft/src/fms_acceleration_peft/framework_plugin_autogptq.py#L47), where this will raise an [AccelerationPluginConfigError](https://github.com/foundation-model-stack/fms-acceleration/blob/79bc89b3ce15d804e826d7c98e11bde0182fa7a4/plugins/framework/src/fms_acceleration/framework_plugin.py#L183)

But yes im tihking as we implement the User facing dataclass, I will hide this argument, and only have it in the `AccelerationFrameworkConfig` dataclass (but also hardcode True)

When we have the dataclass I will remove all the YAMLs.

tuning/sft_trainer.py
fabianlim: some of the `augmentation` requires some of the existing configurations to operate properly. For instance GPTQ LoRA, we need to know the lora configuration to instantiate the LoRA trainable model. 

So for instance inside our GPTQ plugin, we need to check if `peft_config` is None, if it is we will [raise an assertion](https://github.com/foundation-model-stack/fms-acceleration/blob/79bc89b3ce15d804e826d7c98e11bde0182fa7a4/plugins/accelerated-peft/src/fms_acceleration_peft/framework_plugin_autogptq.py#L205).

fabianlim: I have updated the PR starting from https://github.com/foundation-model-stack/fms-hf-tuning/pull/157#:~:text=%E2%80%A6-,9580b99,-fmt%20%2B%20lint to remove the yamls.

fixtures/accelerated-peft-autogptq-sample-configuration.yaml
fabianlim: as shown in https://github.com/foundation-model-stack/fms-hf-tuning/pull/157#:~:text=%E2%80%A6-,9580b99,-fmt%20%2B%20lint for all non supported cases we will have multi-levels of checking
- checking in the dataclass
- checking in the framework itself

I will update the tests to clarify these checks.

fabianlim:
@Ssukriti @alex-jw-brooks I have made the requested changes:
- now with dataclass parsing, it should be easy for the user to figure out what the required arguments are from inspecting the dataclasses. There is no more specifying of YAML.
- see for example [quantized_lora_config](https://github.com/fabianlim/fms-hf-tuning/blob/accel-pr/tuning/config/acceleration_configs/quantized_lora_config.py). 
- I have put unit tests in [tests](https://github.com/fabianlim/fms-hf-tuning/blob/accel-pr/tests/acceleration/test_acceleration_framework.py). It is almost complete but I might add one or two more.

Notes:
- it was not that straightforward to get `transformers.HfArgumentParser` work with nested dataclasses. I had to implement two additional utilities [here](https://github.com/fabianlim/fms-hf-tuning/blob/accel-pr/tuning/config/acceleration_configs/utils.py)
- The argument parsing works as expected, you can see the [tests](https://github.com/fabianlim/fms-hf-tuning/blob/accel-pr/tests/acceleration/test_acceleration_framework.py)

tuning/sft_trainer.py
```diff
@@ -83,6 +90,8 @@ def train(
                               or TrainerControllers. Callbacks associated with \
                               tracker with automatically be added.
         exp_metadata: Dict of key value pairs passed to train to be recoreded by the tracker.
+        quantized_lora_config: tuning.config.acceleration_configs.QuantizedLoraConfig \
```
Ssukriti, line 93: some questions I have:

since this requires augmentation and peft_config , is quantized_lora_config only expected to work with LoRA tuning and Loraconfig ,  or can one also apply it with fine tuning and prompt tuning?

I dont see any checks saying only if LorAconfig can apply this plugin

Ssukriti:
@fabianlim thank you for the redesign and the accelerationframework unit tests. Design looks good, so no major changes needed. Just few comments:

1. question above on when is GPTQLora applicable

2. The tests for acceleration framework check integration in detail, thank you. But I think if possible it would be beneficial to add tests to test_sft_trainer as well, just to ensure 
a. if quantization_lora config passed, tuning still succeeds and model after tuning can still be loaded and inferred on (like rest of tuning unit tests) . I understand you may need a quantized base model to do these unit tests? so let me know if its feasible to add or not . 
b. similarly with kernels as well, if passed, tuning still succeeds.

3. Documentation may need some more updates - but we can do that in subsequent PRs. Mainly if there is any limitation on what model types we can apply QLoRA to (needs a 4bit quantized model), that should be documented and highlighted in README

fabianlim:
@Ssukriti thank you for reviewing!

> since this requires augmentation and peft_config , is quantized_lora_config only expected to work with LoRA tuning and Loraconfig , or can one also apply it with fine tuning and prompt tuning?

The `AccelerationFramework` has logic inside its plugins to [check if the](https://github.com/foundation-model-stack/fms-acceleration/blob/40aad4683e01be3faebbcaae2670524c431cbf3e/plugins/accelerated-peft/src/fms_acceleration_peft/framework_plugin_autogptq.py#L239) `peft_config` is not properly set. The `peft_config` is passed to the framework via the [`augmentation` step](https://github.com/foundation-model-stack/fms-hf-tuning/blob/a07ae26296cad8d17a8ed4d504b6a0e1c7ba84a8/tuning/sft_trainer.py#L268-L269). Maybe I can have a unit test to demonstrate this.

> The tests for acceleration framework check integration in detail, thank you. But I think if possible it would be beneficial to add some top level tests to test_sft_trainer as well, just to ensure
a. if quantization_lora config passed, tuning still succeeds and model after tuning can still be loaded and inferred on (like rest of tuning unit tests) . I understand you may need a quantized base model to do these unit tests? so let me know if its feasible to add or not .
b. similarly with kernels as well, if passed, tuning still succeeds.
Either way might be good to add tests for failure case as well in test_sft_trainer. What if user passes GPTQLoraconfig to an unsupported model that is not quantized - what happens then? is error given and caught

Actually I think you are referring to this [kind of test](https://github.com/foundation-model-stack/fms-hf-tuning/blob/a07ae26296cad8d17a8ed4d504b6a0e1c7ba84a8/tests/acceleration/test_acceleration_framework.py#L239-L250). We do already have one. 
- I can add two much such tests, i) for fused ops and kernels, and ii) a negative test on an unsupported (non-quantized) model. 

 >Documentation may need some more updates - but we can do that in subsequent PRs. Mainly if there is any limitation on what model types we can apply QLoRA to (needs a 4bit quantized model), that should be documented and highlighted in README
ok np!

fabianlim:
@Ssukriti I have added a set of extra tests 

These tests involve calling `sft_trainer.train`, so I think these are the tests that you are requesting for. They can be considered integration tests, to ensure correct working of the framework integrated with the trainer. 

Therefore, I have added the following integration tests:
1. `test_framework_raises_due_to_invalid_arguments`: this test will demonstrate that the framework plugins will also check the arguments, and throw if invalid arguments are passed in. For example,. if we attempt to use `accelerated peft` and no peft_config is passed, then it throws
2. `test_framework_intialized_properly_peft`:  this was refactored from an older test, but now it also demonstrates `BNBQLora` properly loading. Previously it was only `GPTQ_LoRA` happy path
3. `test_framework_intialized_properly_foak`: this one demonstrates that fused ops and kernels are also integrated properly.

Ssukriti (commented):
will approve after conflicts are merged and all checks pass. Suggested minor edits to make it clear that GPTQLora needs a peft_config.LORAconfig passed. we have to make that clear in our documentation as users of tuning will not know that. 


Remaining work in subsequent PRs after this PR is merged:
1. we need to ensure that in CI/CD all the tests run regularly and they are not skipped. That means all dependencuies should be installed for our tests to run regularly . Purpose is to ensure with every release, all tests pass. 
2. Unit tests - Additional unit tests added are good, thank you. I did want to ensure model after tuning after GPTQLora is of correct format , and can be loaded and inferred correctly. We have had issues in past, when something would change and model format produced is no longer correct - we should have tests to capture that to have full confidence (will DM about this)

tests/acceleration/test_acceleration_framework.py
```diff
@@ -0,0 +1,455 @@
...
+        installation_path, (MockedPlugin, spy) = mock_and_spy
+
+        # 1. mock a plugin class
+        # 2. register the mocked plugin
+        # 3. call sft_trainer.train
+        with build_framework_and_maybe_instantiate(
+            [([installation_path], MockedPlugin)],
+            instantiate=False,
+        ):
+            sft_trainer.train(
```
Ssukriti, line 366: also will produce artifacts correct? to avoid artifacts getting generated and stored in local machines whenever we run tests locally, please use tempdir `    with tempfile.TemporaryDirectory() as tempdir:` whenever applicable. 

This will ensure directory is cleaned after test is complete. 
See test https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/tests/test_sft_trainer.py#L351C1-L351C51

README.md
```diff
@@ -326,6 +332,60 @@ tuning/sft_trainer.py  \
...
+
+```
+An acceleration feature is requested by specifying the '--auto_gptq' argument, but the this requires acceleration packages to be installed. Please do:
+- python -m fms_acceleration install fms_acceleration_peft
+```
+
+The list of configurations for various `fms_acceleration` plugins:
+- [quantized_lora_config](./tuning/config/acceleration_configs/quantized_lora_config.py): For quantized 4bit LoRA training
+  - `--auto_gptq`: 4bit GPTQ_LoRA with AutoGPTQ
+  - `--bnb_qlora`: 4bit QLoRA with bitsandbytes
```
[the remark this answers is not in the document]
fabianlim: Sorry I need to rewrite this abit, but I will copy out your suggestion

README.md
```diff
@@ -326,6 +332,60 @@ tuning/sft_trainer.py  \
...
+- [fused_ops_and_kernels](./tuning/config/acceleration_configs/fused_ops_and_kernels.py) (experimental):
+  - `--fused_lora`: fused lora for more efficient LoRA training.
```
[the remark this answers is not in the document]
fabianlim: yes i will add one comment

tuning/sft_trainer.py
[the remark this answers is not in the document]
fabianlim: addressed in https://github.com/foundation-model-stack/fms-hf-tuning/pull/157/commits/06c48722b874fd64f50417cc7e13d5f1a1bf821f

fabianlim:
@Ssukriti i have rebased the changes, and also created an issue here to track the remaining work items #205. Working on making the tests pass

Ssukriti (approved):

PR #159, opened 2024-05-15
add .complete file to output dir when done

kellyaa:
Add a .complete file to the output directory when finished training. This is a signal to users who are monitoring the output directory that we are finished writing them model to storage.


tedhtchang (approved):
/LGTM

anhuong (approved):
I agree with Ted, LGTM, just requires pulling in changes from main branch

PR #161, opened 2024-05-17
Add formatting function alpaca

Ssukriti:
### Description of the change

Allows passing a data formatter template to create single sequence of dataset_text_field internally with JSONL supplied . Eliminates need to do preprocessing and format data to alpaca style https://github.com/foundation-model-stack/fms-hf-tuning?tab=readme-ov-file#data-format

Adds a new data_formatter field to args similar to SFT Trainer formatting function https://huggingface.co/docs/trl/en/sft_trainer#trl.SFTTrainer
Similarly, users either need to pass a dataset_text_field in JSON with preformatted template or can pass a formatter to do formatting on the fly 

### Related issue number
https://github.ibm.com/ai-foundation/watson-fm-stack-tracker/issues/863

### How to verify the PR

1. Unit tests
2. README updated 

### Was the PR tested


tuning/utils/data_utils.py
```diff
@@ -0,0 +1,34 @@
...
+    formatted_dataset_field = "formatted_data_field"
+    template += eos_token
+    def formatter(element):
+        nonlocal template
+
+        def replace_text(match_obj):
+            captured_groups = match_obj.groups()
+            if len(captured_groups) != 1:
+                raise ValueError(
+                        "Unexpectedly captured multiple groups in verbalizer rendering"
```
alex-jw-brooks, line 23: Any thoughts on changing this error message? It looks like this might be a legacy message from caikit nlp or something like that, but verbalizer rendering might be confusing. Maybe `"Unexpectedly captured multiple groups in template formatting"` or something would be a bit more understandable?

tuning/utils/data_utils.py
```diff
@@ -0,0 +1,34 @@
...
+    Args:
+        dataset: the HF Dataset element loaded from a JSON or DatasetDict object.
+        template: Template to format data with. Features of Dataset 
+            should be referred to by {{key}}
+    Returns:
+        Formatted HF Dataset, dataset_field name that contains formatted data.
+    """
+
+    formatted_dataset_field = "formatted_data_field"
+    template += eos_token
```
alex-jw-brooks, line 15: Will this cause issues if the template provided already has a terminating EOS token?

tuning/utils/data_utils.py
```diff
@@ -0,0 +1,34 @@
...
+                        "Unexpectedly captured multiple groups in verbalizer rendering"
+                    )
+
+            index_object = captured_groups[0]
+            if index_object not in element:
+                raise KeyError("Requested template string is not a valid key in dict")
+                
+            return element[index_object]
+
+        return {formatted_dataset_field : re.sub(r"{{([_a-z\sA-Z0-9]+)}}", replace_text, template)}
```
alex-jw-brooks, line 32: How were the allowed characters in the matching regex chosen? Is this documented somewhere?

tuning/sft_trainer.py
```diff
@@ -238,12 +246,20 @@ def train(
     }
 
     json_dataset = datasets.load_dataset("json", data_files=data_files)
-    formatted_train_dataset = json_dataset["train"].map(format_dataset)
+    if data_args.data_formatter_template:
+        formatted_train_dataset, data_args.dataset_text_field = \
+            formatting_function(json_dataset["train"], data_args.data_formatter_template, tokenizer.eos_token)
```
alex-jw-brooks, line 251: IMO it's a little confusing to name the formatter for the custom templates `formatting_function`. I think the reason is that to me `formatting_function` sounds like something that you would map over the dataset, while `format_dataset` sounds like something that does the mapping, but in the `if/else` here, it's the opposite.

Any thoughts on renaming this `apply_custom_formatting_template` or something?

tuning/sft_trainer.py
```diff
@@ -218,15 +219,22 @@ def train(
...
-            raise ValueError("Dataset_text_field is None, needs to be set for training")
         data_collator = DataCollatorForCompletionOnlyLM(
             response_template_ids,
             tokenizer=tokenizer,
             ignore_index=configs.IGNORE_INDEX,
         )
         packing = False
 
+    # Currently we support formatted datasets with single sequence instances.
+    if (data_args.dataset_text_field is None) and (data_args.data_formatter_template is None):
```
alex-jw-brooks, line 230: I noticed that in this condition, it explicitly checks `None`, and the next it's checking for `None` or empty strings. Any thoughts on changing this to `if data_args.dataset_text_field and data_args.data_formatter_template :` so that `None` and empty string are treated consistently here, since I think that makes sense for these checks?

tests/utils/test_data_utils.py
```diff
@@ -0,0 +1,51 @@
...
+    formatted_dataset,  dataset_text_field = data_utils.formatting_function(json_dataset, template)
+    # a new dataset_text_field is created in Dataset
+    assert dataset_text_field in formatted_dataset['train'][0]
+    assert formatted_dataset['train'][0][dataset_text_field] == expected_response
+
+def test_formatting_function_adds_eos_token():
+    json_dataset = datasets.load_dataset("json", data_files=TWITTER_COMPLAINTS_DATA)
+    template = "### Input: {{Tweet text}} \n\n ### Response: {{text_label}}"
+    # First response from the data file that is read.
+    expected_response = "### Input: @HMRCcustomers No this is my first job \n\n ### Response: no complaintEOS"
```
alex-jw-brooks, line 40: Why is it assumed to be okay that appending a raw EOS to the string will not change the tokenization, since this is operating on raw strings and not appending a token? I.e., in this kind of contrived example, just tokenizing the final token as `complaintEOS` instead of two tokens, the latter of which is EOS

tests/test_sft_trainer.py
```diff
@@ -63,7 +63,6 @@
     "save_strategy": "epoch",
```
alex-jw-brooks, line 63: just a more general question - is there any scenario you can imagine where `{{foo}}` literal is actually expected to be part of the template, and not something to render, as this would currently not be allowed?

Ssukriti:
@alex-jw-brooks the PR is ready for review now. I will address your existing comments soon. Thank you

tuning/utils/data_utils.py
```diff
@@ -0,0 +1,34 @@
...
+        template: Template to format data with. Features of Dataset 
+            should be referred to by {{key}}
+    Returns:
+        Formatted HF Dataset, dataset_field name that contains formatted data.
+    """
+
+    formatted_dataset_field = "formatted_data_field"
+    template += eos_token
+    def formatter(element):
+        nonlocal template
```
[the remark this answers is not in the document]
Ssukriti: fixed

tuning/utils/data_utils.py
```diff
@@ -0,0 +1,34 @@
+import re
+from datasets import Dataset
+
+def formatting_function(dataset, template, eos_token=""):
```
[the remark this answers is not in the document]
Ssukriti: I left it because it is a utility function and can be called as preprocessing step by any users if they wish. so didnt want to impose EOS token

tuning/utils/data_utils.py
```diff
@@ -0,0 +1,34 @@
...
+    """Function to format datasets with Alpaca style / other templates.
+    Args:
+        dataset: the HF Dataset element loaded from a JSON or DatasetDict object.
+        template: Template to format data with. Features of Dataset 
+            should be referred to by {{key}}
```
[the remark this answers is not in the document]
Ssukriti: fixed.

tuning/utils/data_utils.py
Ssukriti: fixed

tuning/utils/data_utils.py
Ssukriti: good catch, sorry juts took it from caikit NLP and forgot to update. Updated to any char now, leaving it open for you to have a look

tuning/sft_trainer.py
Ssukriti: changed name as suggested

tuning/sft_trainer.py
Ssukriti: updated but first check is checking that at least one is set , second check is checking that only one is set. we could consolidate to one check in one if statement , but just making error messages more specific in 2 separate checks

tuning/utils/data_utils.py
Ssukriti: apparently not, I just duplicated functionality already in fms-tuning https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/tuning/sft_trainer.py#L235 where a formatter with EOS token is applied. I just didnt want us to loop over dataset twice hence moved that here . Quality has been tested with existing behavior.

tests/utils/test_data_utils.py
Ssukriti: i agree its not ideal, but I just duplicated existing behavior in train() https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/tuning/sft_trainer.py#L235 to this function to avoid looping and formatting dataset twice. Appended it same way as it was happening before - we can make update in future PRs to correct in both places

tests/test_sft_trainer.py
Ssukriti: ya we ll make it clear that its not allowed. Only keys of template should be in {{}} , if anyone has other requirements they can raise request

Ssukriti:
I have manually verified that formatted data matches our pre-processing step and no warnings occur while training . We will do another quality test to be safe before 0.2.0 release. Safe to merge PR though

README.md
```diff
@@ -24,7 +24,10 @@ pip install -e ".[aim]"
 ```
 
 ## Data format
-The data format expectation is a single column text. The trainer is configured to expect a response template as a string. For example, if one wants to prepare the `alpaca` format data to feed into this trainer, it is quite easy and can be done with the following code.
+We support two data formats:
```
alex-jw-brooks, line 27: this documentation is great! nice job

tuning/sft_trainer.py
```diff
@@ -218,15 +219,26 @@ def train(
...
+    if not (data_args.dataset_text_field or data_args.data_formatter_template):
+        raise ValueError(
+            "Dataset_text_field and data_formatter_template are None. \
+                            One of them needs to be set for training"
+        )
+    # Only one of dataset_text_field or data_formatter_template should be set.
+    if data_args.dataset_text_field and data_args.data_formatter_template:
+        raise ValueError(
+            "Dataset_text_field and data_formatter_template are set. \
+                            Only one of them needs to be set for training"
```
alex-jw-brooks, line 239: Can you change this to be more explicit that only one of them can be set, instead of needs to be set, and also change `Dataset_text_field` to lowercase in these two errors since it's the field with the underscores? 

E.g., `dataset_text_field and data_formatter_template are both set, but are mutually exclusive options`?

tuning/sft_trainer.py
```diff
@@ -238,12 +250,34 @@ def train(
...
+            tokenizer.eos_token,
+        )
+    else:
+        formatted_train_dataset = json_dataset["train"].map(format_dataset)
     logger.info("Training dataset length is %s", len(formatted_train_dataset))
 
     formatted_validation_dataset = None
     if data_args.validation_data_path:
-        formatted_validation_dataset = json_dataset["validation"].map(format_dataset)
+        if data_args.data_formatter_template:
```
alex-jw-brooks, line 268: there will need to be some consolidation with my PR, just FYI - yours will be merged first, but the calling of this is probably going to get shuffled into some utils to make this a bit easier to manage with different collators

tuning/utils/data_utils.py
```diff
@@ -0,0 +1,36 @@
...
+                    "Unexpectedly captured multiple groups in template formatting"
+                )
+
+            index_object = captured_groups[0]
+            if index_object not in element:
+                raise KeyError("Requested template string is not a valid key in dict")
+
+            return element[index_object]
+
+        return {formatted_dataset_field: re.sub(r"{{(.+)}}", replace_text, template)}
```
alex-jw-brooks, line 34: Just a thought, matching on `(.+)` is also going to match `{` / `}`, which could cause some slightly weird scenarios if people format templates badly - should we be more restrictive on the templates characters? E.g., `[ 0-9a-zA-Z_-]` or something?

I don't have a strong opinion, but might be a good idea to discuss before merging rather than after

tests/utils/test_data_utils.py
```diff
@@ -0,0 +1,66 @@
...
+# First Party
+from tests.data import TWITTER_COMPLAINTS_DATA
+
+# Local
+from tuning.utils import data_utils
+
+
+def test_apply_custom_formatting_template():
+    json_dataset = datasets.load_dataset("json", data_files=TWITTER_COMPLAINTS_DATA)
+    template = "### Input: {{Tweet text}} \n\n ### Response: {{text_label}}"
```
alex-jw-brooks, line 31: Some overlap with previous comment on whether or not we want to add more restrictions or at least recommended conventions for templates here, since these are a bit different 😄

alex-jw-brooks (changes requested):
looks great! some small questions

tuning/utils/data_utils.py
Ssukriti: the regex here is used to find fields of JSON , after field is found the text is put into the template.if people have / in JSON field - "{{input/text}}" , I think its ok to allow it?

Ssukriti: oh that wont work, hmm ya I can restrict and document. doing that now

tuning/sft_trainer.py
Ssukriti: updated

tuning/utils/data_utils.py
Ssukriti: tested various edge cases and it worked ok

Ssukriti: I pushed change anyway to be safe and updated README

Ssukriti:
@alex-jw-brooks the PR is ready for review

alex-jw-brooks (approved):
LGMT, thanks Sukriti!

PR #162, opened 2024-05-20
tests for fixing full fine tuning

Ssukriti:
### Description of the change

Unit tests for full fine tuning should not have any tuning config set . This PR fixes the unit tests by unsetting the PT config that was getting set for FT 

### How to verify the PR

1. Unit tests pass
2. No adapter_config is created which is how Full fine tuning should be 

### Was the PR tested

### DIFF
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -42,9 +42,4 @@ def causal_lm_train_kwargs(train_kwargs):
         tuning_config = lora_config
     elif train_kwargs.get("peft_method") == "pt":
         tuning_config = prompt_tuning_config
-    return (
-        model_args,
-        data_args,
-        training_args,
-        tuning_config
-    )
+    return (model_args, data_args, training_args, tuning_config)
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -37,11 +37,14 @@ def causal_lm_train_kwargs(train_kwargs):
         lora_config,
         prompt_tuning_config,
     ) = parser.parse_dict(train_kwargs, allow_extra_keys=True)
+    tuning_config = None
+    if train_kwargs.get("peft_method") == "lora":
+        tuning_config = lora_config
+    elif train_kwargs.get("peft_method") == "pt":
+        tuning_config = prompt_tuning_config
     return (
         model_args,
         data_args,
         training_args,
-        lora_config
-        if train_kwargs.get("peft_method") == "lora"
-        else prompt_tuning_config,
+        tuning_config
     )
--- a/tests/test_sft_trainer.py
+++ b/tests/test_sft_trainer.py
@@ -277,6 +277,9 @@ def test_run_causallm_ft_and_inference():
         model_args, data_args, training_args, tune_config = causal_lm_train_kwargs(
             BASE_FT_KWARGS
         )
+        # Just assuring no tuning config is passed for PT or LoRA
+        assert tune_config is None
+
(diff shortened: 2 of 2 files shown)


tests/test_sft_trainer.py
```diff
@@ -68,9 +68,9 @@
 BASE_LORA_KWARGS["peft_method"] = "lora"
 
 BASE_FT_KWARGS = copy.deepcopy(BASE_PEFT_KWARGS)
-BASE_FT_KWARGS["peft_method"] = ""
-BASE_FT_KWARGS["prompt_tuning_init"] = ""
-BASE_FT_KWARGS["prompt_tuning_init_text"] = ""
+BASE_FT_KWARGS["peft_method"] = None
+del BASE_FT_KWARGS["prompt_tuning_init"]
```
Ssukriti, line 72: Just FYI - we dont expect users to pass these keys with full fine tuning . But even if they are set, as long as peft_method is None, doesnt make a difference. Since this section is about setting correct args for testing , we should set what is expected

tests/test_sft_trainer.py
```diff
@@ -277,13 +277,14 @@ def test_run_causallm_ft_and_inference():
...
         )
+        # Just assuring no tuning config is passed for PT or LoRA
+        assert tune_config is None
+
         sft_trainer.train(model_args, data_args, training_args, tune_config)
 
         # validate ft tuning configs
         _validate_training(tempdir)
         checkpoint_path = _get_checkpoint_path(tempdir)
-        adapter_config = _get_adapter_config(checkpoint_path)
```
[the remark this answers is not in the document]
anhuong: Agreed!

tests/helpers.py
```diff
@@ -37,11 +37,9 @@ def causal_lm_train_kwargs(train_kwargs):
...
     ) = parser.parse_dict(train_kwargs, allow_extra_keys=True)
-    return (
-        model_args,
-        data_args,
-        training_args,
-        lora_config
-        if train_kwargs.get("peft_method") == "lora"
-        else prompt_tuning_config,
-    )
+    tuning_config = None
```
anhuong, line 40: hmm instead of this helper method, could we import build.utils and run `process_launch_training_args()` instead?

tests/test_sft_trainer.py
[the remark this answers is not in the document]
anhuong: Great catch on this fine tuning unit test Sukriti!

tests/helpers.py
Ssukriti: we could, but is that not going away too after the refactor when we ll have a common function? ideally we shouldnt use anything and just create config classes themselves as purpose here is to test functionality of other things given arguments are correct . let me see how hard it would be to just get of this entire need to pass kwargs

anhuong: I was thinking of mentioning that this would be going away anyways soon, but I thought changing to use the import would be easy. If not then no worries

anhuong (approved):
Potentially look into importing build.utils instead of having a test helper, but otherwise LGTM

Ssukriti:
@anhuong I have refactored tests in next PR https://github.com/foundation-model-stack/fms-hf-tuning/pull/163 to not use any argument parsing from a large dict, but rather have explicit parameter passing in each test. This will avoid errors in how args are created. Each test has more control over args passed. If you are in agreement, we can merge this and then just complete the refactor - I have not been able to finish it yet

PR #163, opened 2024-05-21
Refactor tests explicit params

Ssukriti:
### Description of the change

This removes the need for argument parser for tests and allows each test to have full control on what args are passed. This may be beneficial to avoid errors in how dataclasses are created from a large dict of args. Each test can just update the args the test is concerned with. Might be easier to maintain as we add new args and new dataclasses. 

### Was the PR tested


tests/test_sft_trainer.py
```diff
@@ -358,134 +302,110 @@ def test_tokenizer_has_no_eos_token():
...
     """Ensure that malformatted data explodes due to failure to generate the dataset."""
-    TRAIN_KWARGS = {
-        **BASE_PEFT_KWARGS,
-        **{"training_data_path": EMPTY_DATA, "output_dir": "foo/bar/baz"},
-    }
-    model_args, data_args, training_args, tune_config = causal_lm_train_kwargs(
-        TRAIN_KWARGS
-    )
+    data_args = copy.deepcopy(DATA_ARGS)
+    data_args.training_data_path = MALFORMATTED_DATA
```
anhuong, line 351: Should be EMPTY_DATA instead of MALFORMATTED DATA

tests/test_sft_trainer.py
```diff
@@ -193,34 +150,27 @@ def test_run_causallm_pt_init_text():
...
-        validation_peft["validation_data_path"] = TWITTER_COMPLAINTS_DATA
-        validation_peft["evaluation_strategy"] = "epoch"
-        model_args, data_args, training_args, tune_config = causal_lm_train_kwargs(
-            validation_peft
-        )
-
-        assert data_args.validation_data_path == TWITTER_COMPLAINTS_DATA
+        train_args = copy.deepcopy(TRAIN_ARGS)
+        train_args.output_dir = tempdir
+        train_args.evaluation_strategy = "epoch"
```
anhuong: Transformers just updated this flag deprecating `evaluation_strategy` in favor of `eval_strategy` https://github.com/huggingface/transformers/commit/60d5f8f9f04026cb801d0dc5158bf4531e250072

anhuong (commented):
Two small changes needed that I made, otherwise the changes look great and it's very nice to not have to worry about the parsing of the params and be more explicit about what is being passed in.

anhuong (approved):
LGTM, thanks Sukriti! Excellent refactor

PR #166, opened 2024-05-22
Add Support for Passing Pretokenized Datasets to TRL

alex-jw-brooks:
### Description of the change
- Adds support for pretokenized datasets.
- Refactors logic for formatting the train / eval dataset + picking a collator into preprocessing utils

### How to verify the PR

- Run the unit tests and see all preprocessing tests are passing 
- Run an end to end tuning
```
export DATA_PATH=/home/SSO/us2j7257/fms-hf-tuning/tests/data/twitter_complaints_input_output.json
export CUDA_VISIBLE_DEVICES=0
export MODEL_PATH=TinyLlama/TinyLlama-1.1B-step-50K-105b 
export OUTPUT_PATH=out


python tuning/sft_trainer.py  --model_name_or_path $MODEL_PATH  --training_data_path $DATA_PATH  --output_dir $OUTPUT_PATH  --num_train_epochs 20  --per_device_train_batch_size 4 --per_device_eval_batch_size 4  --gradient_accumulation_steps 1  --evaluation_strategy "no"  --save_strategy "epoch"  --learning_rate 0.03  --weight_decay 0.  --warmup_ratio 0.03  --lr_scheduler_type "cosine"  --logging_steps 1  --include_tokens_per_second  --packing False  --use_flash_attn False  --tokenizer_name_or_path $MODEL_PATH --torch_dtype "float32" --peft_method "pt" --num_virtual_tokens 1500 --prompt_tuning_init_text "Classify if the tweet is a complaint or not:"
```
- You can also inspect the output of the seq2seq collator from above ^ and verify that padding works correctly, and that the input ids / labels / attention mask are correctly manipulated.

### Was the PR tested

### DIFF
--- a/tuning/utils/preprocessing_utils.py
+++ b/tuning/utils/preprocessing_utils.py
@@ -1,0 +1,144 @@
+# Copyright The FMS HF Tuning Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import json
+import datasets
+from datasets import Dataset
+from transformers import AutoTokenizer, DataCollatorForSeq2Seq
+from trl import DataCollatorForCompletionOnlyLM
+from tuning.config import configs
+
+def get_data_trainer_kwargs(training_data_path: str, validation_data_path: str, packing: bool, response_template, max_sequence_length, tokenizer, dataset_text_field):
+    """Get trainer args related to data / processing. At the moment, this consists of:
+        - the training dataset
+        - the evaluation dataset
+        - the data collator
+    The result can be kwarg expanded into the trainer initialization.
+    """
+    data_collator = get_data_collator(packing, dataset_text_field, response_template, max_sequence_length, tokenizer)
+    eval_dataset = None
+    if isinstance(data_collator, DataCollatorForSeq2Seq):
+        train_dataset = get_preprocessed_dataset(
+            training_data_path,
+            tokenizer,
+            max_sequence_length,
+            input_field_name="input",
+            output_field_name="output",
+        )
+        if validation_data_path:
+            eval_dataset = get_preprocessed_dataset(
+                validation_data_path,
+                tokenizer,
+                max_sequence_length,
+                input_field_name="input",
+                output_field_name="output",
+            )
+    else:
+        # Collator is a DataCollatorForCompletionOnlyLM or None;
+        # Load it as JSON and apply our normal preprocessing logic
+        train_dataset = get_formatted_dataset(training_data_path, dataset_text_field, tokenizer)
+        if validation_data_path:
+            eval_dataset = get_formatted_dataset(validation_data_path, dataset_text_field, tokenizer)
+    return {
+        "data_collator": data_collator,
+        "train_dataset": train_dataset,
+        "eval_dataset": eval_dataset,
+    }
+
+def get_data_collator(packing, dataset_text_field, response_template, max_sequence_length, tokenizer):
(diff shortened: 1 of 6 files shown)


tuning/utils/preprocessing_utils.py
```diff
@@ -0,0 +1,144 @@
...
+            yield {
+                input_field_name: data[input_field_name],
+                output_field_name: data[output_field_name]
+            }
+    return Dataset.from_generator(
+        get_jsonl_object
+    )
+
+### Utils for custom masking / manipulating input / output strs, etc
+def combine_sequence(input_element: str, output_element: str):
```
Ssukriti, line 119: with my upcoming change to accept a template in API - which will be like verbalizer field. In future we ll need to apply template while combining sequence - which will be a minor addition, that way we can accept "input/output" + custom template. Need not worry about that now though

alex-jw-brooks: Cool sounds good - that is partially why I wrote things this way, even though the input/output are hardcoded in the thing calling this, so we can just pass everything through here as needed

tuning/utils/preprocessing_utils.py
```diff
@@ -0,0 +1,329 @@
...
+            Data related kwargs to be used by the SFT Trainer.
+    """
+    data_collator = get_data_collator(
+        packing, dataset_text_field, response_template, max_sequence_length, tokenizer
+    )
+    eval_dataset = None
+    data_kwargs = {}
+    if isinstance(data_collator, DataCollatorForSeq2Seq):
+        # HACK: This function is never called, but is needed to sidestep TRL's internal validation.
+        data_kwargs["formatting_func"] = lambda x: x
```
alex-jw-brooks, line 71: I was implementing a kind of goofy workaround for TRL's validation logic with wrapping the tokenizer to make `tokenize()` a loop when I realized that there is actually already logic for handling pretokenized data based on the columns of the passed dataset: https://github.com/huggingface/trl/blob/main/trl/trainer/sft_trainer.py#L477

So all we have to do to handle the tokenized data is pass a dummy formatting function, because `_prepare_dataset` will inspect the dataset and return immediately without ever calling the value passed here.

There is additionally an extra argument, `dataset_kwargs`, which is a dict that can pass `skip_prepare_dataset`, which is bool-valued. I think the correct behavior is to just check that value in TRL and skip the validation for the `packing=True` case, which is a small change that does not affect its API or change the supported data formats.

I have opened an issue / pull request in TRL to this effect: https://github.com/huggingface/trl/pull/1673

ashokponkumar:
@alex-jw-brooks Are we using https://github.com/huggingface/trl/pull/1520 ?

tuning/utils/preprocessing_utils.py
```diff
@@ -0,0 +1,329 @@
...
+            Sequence combined with whitespace.
+    """
+    if not input_element.endswith((" ", "\n", "\t")) and not output_element.startswith(
+        (" ", "\n", "\t")
+    ):
+        return input_element + " " + output_element
+    return input_element + output_element
+
+
+def preprocess_and_tokenize(
```
Ssukriti, line 278: I think here we ll need to accept template as argument and use template to combine sequence

tuning/utils/preprocessing_utils.py
```diff
@@ -0,0 +1,329 @@
...
+from datasets import Dataset
+from transformers import AutoTokenizer, DataCollatorForSeq2Seq
+from trl import DataCollatorForCompletionOnlyLM
+import datasets
+
+# Local
+from tuning.config import configs
+
+
+def get_data_trainer_kwargs(
```
Ssukriti, line 28: I am just wondering if this wrapper function is hiding too much from main code. Wondering if it would make sense to move this back to main code, esp once the hack is removed . as its basically 2 steps ->
 get collator, get_formatted_dataset -> internally data formatting might happen in different ways. I am still thinking about it, but on initial thoughts I feel it might be good to know high level steps in train()

tuning/utils/preprocessing_utils.py
fabianlim: are the templates going to be jinja style like this? https://huggingface.co/docs/transformers/en/chat_templating

alex-jw-brooks:
Closing this PR, as it was refactored and merged in https://github.com/foundation-model-stack/fms-hf-tuning/pull/260

PR #168, opened 2024-05-23
update eval_strategy flag used in transformers

anhuong:
### Description of the change

Transformers 4.41 updated the `evaluation_strategy` flag, deprecating this previously used one in favor of `eval_strategy`. This was released few days ago. https://github.com/huggingface/transformers/commit/60d5f8f9f04026cb801d0dc5158bf4531e250072

Updating docs to show this change.

### Was the PR tested


Ssukriti:
wiki pages may need update - https://github.com/foundation-model-stack/fms-hf-tuning/wiki/SFT%E2%80%90Trainer%E2%80%90Image-Test-process

anhuong:
For first pass, updating evaluation_strategy to eval_strategy in all locations. I updated the wiki as well already. 

Next, I will create a separate PR to clean up the docs and work to consolidate both the number of params used (only pass in params needed, rely on defaults) and the example config.

PR #173, opened 2024-06-10
fix: bloom model can't run with flash-attn

anhuong:
Fixes the KFTO example so it is a fully functional test that users can run with. Bloom model can't run with flash-attn set.


jbusche (approved):
LGTM, thank you @anhuong

Issue #201, opened 2024-06-19
Performance regression in fms-hf-tuning image


kpouget:
## Describe the bug

As part of our performance evaluation of the fms-hf-tuning fine-tuning stack, we observed a regression between the images we tested:

*    quay.io/modh/fms-hf-tuning:01b3824c9aba22d9d0695399681e6f0507840e7f --> fast (April 17)
```
{'train_runtime': 829.2865, 'train_samples_per_second': 62.707, 'train_steps_per_second': 3.919, 'train_tokens_per_second': 12141.512, 'train_loss': 2.287814903846154, 'epoch': 1.0}
```
*    quay.io/modh/fms-hf-tuning:bd8bf628cd739c7a201a976bc3c1096785353f1a --> slow (May 26)
```
{'train_runtime': 1391.6176, 'train_samples_per_second': 37.368, 'train_steps_per_second': 2.335, 'train_tokens_per_second': 7235.315, 'train_loss': 2.287858623798077, 'epoch': 1.0}
```
*    quay.io/modh/fms-hf-tuning:84b0337b7baee119e909d4e901b6dadfe34c1f9a --> slow (May 21, planned for delivery)
```
{'train_runtime': 1404.3311, 'train_samples_per_second': 37.03, 'train_steps_per_second': 2.314, 'train_tokens_per_second': 7169.813, 'train_loss': 2.2878201622596155, 'epoch': 1.0}
```
Additionally, we observed that with the bd8bf628cd739c7a201a976bc3c1096785353f1a image, the multi-model tuning performance (multiple independent fine-tuning jobs, running on a dedicated GPU) are degraded compared to the reference run (one job running alone in the cluster). This wasn't the case in our first tests with 84b0337b7baee119e909d4e901b6dadfe34c1f9a.

* fine tuning of gpt_bigcode-santacoder against the Alpaca dataset

## Platform

OpenShift AI

## Steps to reproduce

1. Create these 3 objects in a namespace,
    * [pytorchjob.yaml.log](https://github.com/user-attachments/files/15900894/pytorchjob.yaml.log)
    * [cm_entrypoint.yaml.log](https://github.com/user-attachments/files/15900895/cm_entrypoint.yaml.log)
    * [cm_config.yaml.log](https://github.com/user-attachments/files/15900896/cm_config.yaml.log)
2. See the Pod being executed with the `quay.io/modh/fms-hf-tuning:release` image, currently hitting the regression
[...]

PR #203, opened 2024-06-19
deps: pin transformers below v4.41

anhuong:
### Description of the change

- Remove deprecated `requirements.yaml` that is no longer used for install or building
- Pin transformers version due to performance issue found in #201 

### Was the PR tested


anhuong:
Note that we are closing this PR for now while the transformers degradation issue is under investigation. We will only merge this change to `release` branch but not to `main`.

Issue #60, opened 2024-02-27
Ensure Huggingface, Torch Packages Up-to-Date with Necessary Improvements/Fixes


fabianlim:
@hickeyma @Ssukriti @tedhtchang

The platform is build upon the `torch` and `huggingface` ecosystem whereby there are constant fixes that are pushed every release. 
- sometimes we also require to be on nightly pre-releases as there is a biweekly-monthly lag for each release.
- The fear of upgrading packages comes with breaking APIs, which will severely impact any CI/CD testing. On the other hand, fixes improve memory consumption and training speed, and will be a good idea to regress the codebase with the latest nightlies.

#### Desiderata
1. Have a ground source of truth for package versions. A good choice would be `pyproject.toml` that can be used to also address #56 , #57 
2. Have a system whereby nightlies (or new releases) can be constantly regressed with new code changes (a good choice would be to have one such regression for each new PR).
   - will be a good idea to also consider some means to disable regressions, for PRs that do not update source code (e.g., documentation updates). 
3. Have a means to _freeze package versions_ during CD/CI testing, so that if a test fails, we know exactly on what versions it has failed for. And vice-versa, if a commit is known to work, we know exactly for what versions it worked for.

#### NPM addresses these desiderata for `node`

See [here](https://support.deploybot.com/article/131-why-developers-should-use-npm-ci-instead-of-npm-install-and-its-benefits):
- `npm install` will occur from `package.json` with semvar versioning that tells what kind of upgrades (e.g., minor, patch) are allowed. 
   * during `npm install` the packages will be resolved to install the _latest_ packages that satisfy the constraints. 
   * the final versions of packages installed will be output in `package-lock.json`.
   * the `package-lock.json` is typically checked to have a record of what exact package versions are working.
 - during CI/CD, all tests are run only after `npm ci`; note that this differs from `npm install`, in that it installs from `package-lock.json`. __Thus there is no package resolution during testing__. This ensures if the tests pass, we know exactly what package versions they pass for.

#### Solutions for Python

There are at least two possibilities that we know of:
1. Poetry has `poetry.lock`, that works together with `pyproject.toml`. The [poetry flow as described here](https://python-poetry.org/docs/basic-usage/#installing-with-poetrylock) is completely analogue with the `npm` flow described above.
2. Using `pip-compile` to generate `requirements.txt` from `pyproject.toml` and checking that in as the lock file.

Ssukriti:
[...]

PR #209, opened 2024-06-21
build: use poetry for reproducible virtual environments

VassilisVassiliadis:
### Description of the change

Use poetry:

1. to build reproducible virtual environments
2. during CI/CD testing
3. to build the container image

### Related issue number

Contributes to #60 

### How to verify the PR

Use `poetry install` to install `fms-hf-tuning` from source. You will end up installing the python dependencies listed in `poetry.lock`

### Was the PR tested


pyproject.toml
```diff
@@ -25,35 +36,61 @@ classifiers=[
...
+
+[tool.poetry.group.aim.dependencies]
+aim=">=3.19.0,<4.0"
+
+
+[tool.poetry.group.fms-accel]
+optional=true
+
+[tool.poetry.group.fms-accel.dependencies]
+fms_acceleration = {git = "https://github.com/foundation-model-stack/fms-acceleration.git", subdirectory="plugins/framework", rev = "40aad4683e01be3faebbcaae2670524c431cbf3e"}
```
VassilisVassiliadis, line 77: Actually, this doesn't make sense, so I'll revert this change. Poetry already takes  care of "pinning" this dependency for the reproducible virtual environments it creates (i.e. the `poetry.lock` file).

We just have to run `poetry lock` (or manually update python dependencies) whenever fms_acceleration releases a new version and we'd like to update the dependencies of fms-hf-tuning to match those of fms_acceleration.

For example, at the moment fms_acceleration has a constraint [`transformers<4.40.0`](https://github.com/foundation-model-stack/fms-acceleration/blob/40aad4683e01be3faebbcaae2670524c431cbf3e/plugins/framework/pyproject.toml#L26) which propagates to fms-hf-tuning and sets transformers to 4.39.3.

README.md
```diff
@@ -29,6 +34,37 @@ pip install -e ".[fms-accel]"
...
+If you wish to use [aim](https://github.com/aimhubio/aim), then you need to install it:
+```
+poetry install --with aim
+```
+
+If you wish to use [fms-acceleration](https://github.com/foundation-model-stack/fms-acceleration), you need to install it. 
+```
+poetry install --with fms-accel
+```
+`fms-acceleration` is a collection of plugins that packages that accelerate fine-tuning / training of large models, as part of the `fms-hf-tuning` suite. For more details on see [this section below](#fms-acceleration).
```
tedhtchang, line 65: In order to make the the README.md cleaner, should we move this section to the [Development](https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/CONTRIBUTING.md#development) in CONTRIBUTING.md? Replacing the current Installation.

tedhtchang, line 65: In order to make the the README.md cleaner, should we move this section to the [Development](https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/CONTRIBUTING.md#development) in CONTRIBUTING.md? Replacing the current Installation.

README.md
```diff
@@ -29,6 +34,37 @@ pip install -e ".[fms-accel]"
...
 
+
+### Using versions of python dependencies that are known to be compatible with fms-hf-tuning
+
+This repository uses [`poetry`](https://python-poetry.org/) for reproducible builds. You can use poetry to create the same virtual environment that we use for our CI/CD tests like so:
+
+```
+python -m venv venv
+. ./venv/bin/activate
+pip install poetry
```
tedhtchang, line 45: Maybe remove the `pip install poetry`. Instead list poetry as a required packages along with the git, python, pip, in the [Set up your dev environment](https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/CONTRIBUTING.md#set-up-your-dev-environment) and just hpyerlink to the official installation doc maybe highlighting `Poetry should always be installed in a dedicated virtual environment to isolate it from the rest of your system`

README.md
VassilisVassiliadis: I moved this block of text in the development section. Also I noticed that my docs were actually teaching bad practices (i.e. install poetry in your dev venv as opposed to an isolated venv), so I fixed that too.

tuning/sft_trainer.py
```diff
@@ -515,7 +515,7 @@ def main(**kwargs):  # pylint: disable=unused-argument
             exp_metadata,
         )
     except Exception as e:  # pylint: disable=broad-except
-        logging.error(traceback.format_exc())
+        logging.get_logger("error").error(traceback.format_exc())
```
tedhtchang, line 518: Just curious what's this change for ?

VassilisVassiliadis: `logging.error` isn't part of `transformers.utils.logging` .


For example, when you run this:

```bash
 python -c 'from transformers.utils import logging; logging.error("foo")'
```

python raises the following exception:

```
  File "<string>", line 1, in <module>
AttributeError: module 'transformers.utils.logging' has no attribute 'error'
```

so I just created a logger object called "error" and had that print the traceback. I assume that this line of code was originally using the `logging` module for which `logging.error("foo")` works

VassilisVassiliadis: Unfortunately, I don't remember how I made sft_trainer.py execute that code.

VassilisVassiliadis:
I updated the description of the issue with a couple more changes that I had to do for the tests to pass.

build/Dockerfile
```diff
@@ -110,29 +110,31 @@ RUN dnf install -y git && \
...
+# Ensure that git directory is owned by current user, otherwise git raises
+# "fatal: detected dubious ownership" for `/tmp`
+WORKDIR /tmp/fms-hf-tuning
+
 RUN --mount=type=cache,target=/home/${USER}/.cache/pip,uid=${USER_UID} \
-    python -m pip install --user build
-COPY --chown=${USER}:root tuning tuning
-COPY .git .git
-COPY pyproject.toml pyproject.toml
+    python -m pip install --user wheel packaging poetry poetry-plugin-export
```
tedhtchang, line 118: I don't think we want the poetry and its plugins installed into the /home/tuning/.local/lib/python3.11/site-packages and copied to the final release stage because it's for development.


Maybe remove `build` packages because is not used anymore.
Also the `packaging` is already listed as dependencies the pyproject.toml did you have a reason to install here?

VassilisVassiliadis: Thanks, that's a good find!

While fixing this I noticed that I had made another mistake. I was using poetry `groups` instead of poetry `extras`. However, poetry groups are incompatible with pip i.e. `pip install fms-hf-tuning[flash-attn]` wouldn't install the packages from the `flash-attn` group (which in this example would resolve to just the `flash-attn` python package).

I fixed `pyproject.toml`, the related section in `CONTRIBUTING.md`, and the Dockerfile. I also tested building the dockerfile locally with and without using the `WHEEL_VERSION` build-arg and it seems to be working fine.

tox.ini
```diff
@@ -3,50 +3,59 @@ envlist = py, lint, fmt
 
 [testenv]
 description = run unit tests
-deps =
-    pytest>=7
+allowlist_externals=
+    pytest
+    poetry
+commands_pre =
+    poetry install --with dev --no-root
```
tedhtchang, line 10:
```suggestion
    poetry install --extras dev --no-root
```

tox.ini
```diff
@@ -3,50 +3,59 @@ envlist = py, lint, fmt
...
     pytest {posargs:tests}
 
 [testenv:fmt]
 description = format with pre-commit
-deps =
-    pre-commit
-commands = ./scripts/fmt.sh
-allowlist_externals = ./scripts/fmt.sh
+commands_pre =
+    poetry install --with dev --no-root
```
tedhtchang, line 18:
```suggestion
    poetry install --with extras --no-root
```

tox.ini
```diff
@@ -3,50 +3,59 @@ envlist = py, lint, fmt
...
+    ./scripts/fmt.sh
 
 [testenv:lint]
 description = lint with pylint
-deps =
-    pylint>=2.16.2,<=3.1.0
-    pytest
-    .[dev]
+commands_pre =
+    poetry install --with dev --no-root
```
tedhtchang, line 28:
```suggestion
    poetry install --extras dev --no-root
```

tox.ini
```diff
@@ -3,50 +3,59 @@ envlist = py, lint, fmt
...
 description = report unit test coverage
-deps =
+allowlist_externals=
+    poetry
     coverage
-    pytest
-    genbadge[coverage]
+    genbadge
+commands_pre =
+    poetry install --with dev
```
tedhtchang, line 56:
```suggestion
    poetry install --extras dev
```

tedhtchang:
>This change won't affect fms-hf-tuning packages users who do `pip install fms-hf-tuning`.  pip calculate the appropriate indirect dependencies versions.

For package developer, I verified poetry functionalities:
```
# Install dependencies from the poetry.lock file + fms-hf-tuning in editing mode
# If not already in a venv poetry will create and activate one automatically  
poetry install

# with dev group
poetry install --extras=dev

# without fms-hf-tuning in editing mode(dependencies only)
poetry install --no-root

# update packages version in lock file. `git diff poetry.lock`
poetry update --lock

# Update both lock file and venv
poetry update

# verified run all tox tasks
tox -e py,fmt,lint,build,twinecheck,coverage

# Verified install deps specified in the lock file
docker build . -t sft-trainer:dev -f build/Dockerfile
docker run -it sft-trainer:dev pip freeze

```

tedhtchang:
@VassilisVassiliadis some files have been updated could you rebase. thanks

tox.ini
VassilisVassiliadis: Thank you, I missed changing these changes in tox.ini file!

VassilisVassiliadis:
I rebased the PR and removed the instructions to install fms-accell using poetry. I replaced that text with a link to the relevant section in the README.md file.

tedhtchang (approved):
@VassilisVassiliadis Thanks this looks good. And I agree. We need a way to build/rebuild a release image with consistent dependencies for a commit/tag/sha.

@Ssukriti @anhuong Could you take a look.

jbusche:
Hi @tedhtchang, I tried it and it looked good...
I was able to `poetry install` and try the 
```
poetry update --lock
git diff poetry.lock
```
and see that it had an update:
```
 [[package]]
 name = "orjson"
-version = "3.10.5"
+version = "3.10.6"
```
I was then able to install aim with the original orjson and then update it:
```
poetry install --extras=aim
poetry shell
pip list |grep orj
orjson                   3.10.5
poetry update --lock
git diff poetry.lock
-version = "3.10.5"
+version = "3.10.6"
poetry install --extras=aim
Installing dependencies from lock file
Package operations: 0 installs, 1 update, 0 removals
 - Updating orjson (3.10.5 -> 3.10.6)
 ```

**Review by jbusche — DISMISSED** [review]
LGTM, tested on my Red Hat Server

**Review by Ssukriti — COMMENTED** [review]
I pushed some changes, please review.


Additionally in release process, @tedhtchang please add a post step to run `poetry update` and commit the new lock file. This has to be done post release by maintainers to keep it updated

File: .github/workflows/build-and-publish.yaml [hunk 10]
```diff
@@ -43,6 +43,8 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           python -m pip install tox
+          python -m pip install poetry --user
```
fabianlim, line 46: why is there a need to install this to `--user`? If you can install peotry the same way as tox, that this obliviates the need for the `PATH` setting below.

fabianlim: There quite a few instances of these in other worflows as well.

CONTRIBUTING.md
```diff
@@ -42,6 +42,8 @@ If additional new Python module dependencies are required, think about where to
 - If they're optional dependencies for additional functionality, then put them in the pyproject.toml file like were done for [flash-attn](https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/pyproject.toml#L44) or [aim](https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/pyproject.toml#L45).
 - If it's an additional dependency for development, then add it to the [dev](https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/pyproject.toml#L43) dependencies.
 
+if using `poetry` to install dependencies, you can optionally leverage [poetry add](https://python-poetry.org/docs/cli/#add) to add new dependencies to pyproject.
```
fabianlim, line 45: I feel this is abit confusing. If we have a peoetry flow, and there is a new dep, shouldnt we have some guidance on when it has to be added to the lockfile?

Also some guidance to say that for optional deps it should not be added.

fabianlim:
@VassilisVassiliadis there is a PR here to add back the `[.fms-accel]` dep. It was removed before with the intention of being added back https://github.com/foundation-model-stack/fms-hf-tuning/pull/223
- why will `poetry` constraint packages under optional dependencies? that makes no sense

CONTRIBUTING.md
```diff
@@ -82,12 +84,45 @@ The following tools are required:
...
+  - Just like `poetry` install `tox` in an isolated virtual environment
 
 Installation:
-``` 
-pip install -U datasets
-pip install -e .
+
+```bash 
+: Install poetry and tox in an isolated virtual environment
+python3 -m venv isolated
```
fabianlim, line 96: but why we care so much about `poetry` being updated? will there really be some random package to update peotry? If so then we should pay the same care to `tox` but when dont..

Can we add `poetry` as a dep in the `tox` environment, and have tox manage it?

README.md
```diff
@@ -27,7 +28,13 @@ If you wish to use [fms-acceleration](https://github.com/foundation-model-stack/
...
 pip install git+https://github.com/foundation-model-stack/fms-acceleration.git#subdirectory=plugins/framework
 ```
-`fms-acceleration` is a collection of plugins that packages that accelerate fine-tuning / training of large models, as part of the `fms-hf-tuning` suite. For more details on see [this section below](#fms-acceleration).
+`fms-acceleration` is a collection of plugins that packages that accelerate fine-tuning / training of large models, as part of the `fms-hf-tuning` suite. For more details see [this section below](#fms-acceleration).
+
+### 2. Build from source 
+
+We have committed a `poetry.lock` file to allow reproducible enviornments. If building from source, you can clone the repository and use poetry to install as mentioned in [development docs](/CONTRIBUTING.md#development) 
+
+If building in a dockerfile use `poetry export --format requirements.txt` can install same dependencies from the lock file. Maintainers regularly update the lock file.
```
fabianlim, line 37: this line is not very clear.. this will do not any installation correct? Its just an export of the locked deps to a requirements file?

CONTRIBUTING.md
tedhtchang: Poetry depends on `builds` package which conflicts with our local `builds` package.

tox's deps are also installed in same venv.

CONTRIBUTING.md
Ssukriti: true, maybe we can continue adding to pyproject.yml , like we do now and then we have to run `poetry update` ? 
I can change it to that instead

README.md
Ssukriti: ya its just to export the requirements and same lock file and then we have to install. hmm should be just link our dockerfile as example?

CONTRIBUTING.md
fabianlim: oic.. guess `builds` is too common of a package name.

.github/workflows/build-and-publish.yaml
tedhtchang: tox always creates it's own isolated venv.
Agree it would probably work without `--user`

README.md
VassilisVassiliadis: > @VassilisVassiliadis there is a PR here to add back the [.fms-accel] dep. It was removed before with the intention of being added back https://github.com/foundation-model-stack/fms-hf-tuning/pull/223

@fabianlim I'll rebase the PR once #223 makes its way into the `main` branch. 

> why will poetry constraint packages under optional dependencies? that makes no sense

I'm actually not sure. I assume that they do this so that `pip install` and `pip install -E fms-accel` both end up installing the same non-optional dependencies (in our case  transformers).

CONTRIBUTING.md
VassilisVassiliadis: There're 2 ways to add a new dependency. One is to use `poetry` i.e. `poetry add` the other is to manually edit the `pyproject.toml` file and then run `poetry update` which will then ensure `poetry.lock` is consistent with pyproject.toml (that's basically what `poetry add` does on your behalf).

If you opt for the 2nd route but forget to run `poetry update` then the next time you run `poetry install` (or the CI/CD runs it) you'll get an exception that the `poetry.lock` and `pyproject.toml` files are inconsistent and that you should run  the `poetry update` command.

.github/workflows/build-and-publish.yaml
VassilisVassiliadis: Good catch, `tox` and `poetry` should go in the same virtual-environment and that environment should not be the one that we install `fms-hf-tuning` in. Because `tox`  creates a new virtual environment for the `environments` it processes (e.g. `py`, `fmt, etc) poetry and tox will be outside the tested virtual-environment.

We don't need to run tox "inside" one of the virtual environments that tox creates but we do need to run `poetry`. Here's how I chose to do that.

First, I installed poetry under the user pip install directory (which by default for linux is `~/.local/`. Then I updated the file that [$GITHUB_ENV](https://docs.github.com/en/actions/learn-github-actions/variables#passing-values-between-steps-and-jobs-in-a-workflow) points to so that in subsequent steps the `$PATH` environment variable contains the path `~/.local/bin`. 

This ensures that the `poetry` commands running inside `tox` can use the `poetry` executable. Alternatively, we can create a new virtual environment under a directory of our choosing e.g. `/tmp/isolated` in there, we install just poetry and then update `$GITHUB_ENV` so that `$PATH` includes `/tmp/isolated/bin`.

PR #218, opened 2024-06-27
deps: cap transformers at 4.40.2

anhuong:
### Description of the change

Due to performance degredation found in transformers 4.41 (see issue #201 ), pinning transformers at 4.40

### Was the PR tested


Ssukriti (approved):
thank you

PR #224, opened 2024-06-30
minor refactor to allow modular functions

Ssukriti:
### Description of the change

allows minor refactoring to allow easier integration of pretokenized input and other forms of masking
Changes added:

    1. validate_data_args function - validates the JSON formats. Can be further extended to do  any validation for pretokenized datasets can also be done in this function

    2. get_data_collator - function to return appropriate data collator as per data format .

### Related issue number

1. Related to : https://github.ibm.com/ai-foundation/watson-fm-stack-tracker/issues/991 - gives modules on where to add pretokenized changes
2. first step to allow masking without completion LM

### Was the PR tested

### DIFF
--- a/tuning/sft_trainer.py
+++ b/tuning/sft_trainer.py
@@ -34,7 +34,7 @@ from transformers import (
     TrainerCallback,
 )
 from transformers.utils import is_accelerate_available, logging
-from trl import DataCollatorForCompletionOnlyLM, SFTConfig, SFTTrainer
+from trl import SFTConfig, SFTTrainer
 import datasets
 import fire
 import transformers
@@ -62,6 +62,7 @@ from tuning.utils.error_logging import (
     USER_ERROR_EXIT_CODE,
     write_termination_log,
 )
+from tuning.utils.preprocessing_utils import get_data_collator, validate_data_args
 
 
 def train(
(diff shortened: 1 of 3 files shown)


tuning/utils/preprocessing_utils.py
```diff
@@ -18,32 +18,149 @@
...
+        #     # if dataset_text_field is None and response_template is None:
+        #         # Use the seq2seq data collator; note that this automatically pads labels with -100
+        #         return DataCollatorForSeq2Seq(
+        #             tokenizer=tokenizer, padding=True, max_length=max_sequence_length
+        #         )
+        # 2. add anything needed for preprocessed input
+
+
+###################################################################################
+### The functions below are not yet used. Iterative development towards new features
```
Ssukriti, line 117: dont worry about reviewing anything below this line as it is not used yet

tuning/sft_trainer.py
```diff
@@ -244,31 +237,14 @@ def train(
...
-            # We should do this validation up front, then do the encoding, then handle the collator
-            raise ValueError("Response template is None, needs to be set for training")
-        data_collator = DataCollatorForCompletionOnlyLM(
-            response_template_ids,
-            tokenizer=tokenizer,
-            ignore_index=configs.IGNORE_INDEX,
-        )
         packing = False
 
-    # Currently we support formatted datasets with single sequence instances.
```
Ssukriti, line 258: moved to validate function

tuning/sft_trainer.py
```diff
@@ -195,14 +196,6 @@ def train(
             }
         )
 
-    # TODO: near term - how response template ids are parsed out needs to be cleaned.
```
Ssukriti, line 198: moved to get_data_collator function as we only need it for completionLM colator

kmehant (approved):

tuning/utils/preprocessing_utils.py
```diff
@@ -25,25 +25,141 @@
...
 
-def validate_data_args(
+def validate_data_args(data_args: configs.DataArguments, packing: bool):
+
+    assert isinstance(
+        data_args.training_data_path, str
+    ), "Training data path has to be set and str"
+
+    # Dataset containing single sequence needs a response template for masking
+    if data_args.response_template is None and data_args.dataset_text_field is not None:
```
praveenj83, line 35: Isn't this condition fully subsumed by the next error condition? The next error condition is when packing is false and response_template is empty. This one is the same condition AND dataset_text_field is non-empty. The error will be spotted even without this condition - is the intent to provide a specific error message?

praveenj83 (commented):
One minor comment on this PR

PR #226, opened 2024-07-01
wip: all the 20 plus PRs in one commit

HarikrishnanBalagopal:
### Was the PR tested

### DIFF
--- a/tuning/trainercontroller/callback.py
+++ b/tuning/trainercontroller/callback.py
@@ -273,6 +273,7 @@ class TrainerControllerCallback(TrainerCallback):
                         operation_action.instance.act(
                             action=operation_action.action,
                             event_name=event_name,
+                            tc_metrics=self.metrics,
                             **kwargs,
                         )
 
@@ -449,6 +450,20 @@ class TrainerControllerCallback(TrainerCallback):
         kwargs["state"] = state
         kwargs["control"] = control
         self._actions_on_event(event_name="on_step_end", **kwargs)
+
+    def on_save(
+        self,
+        args: TrainingArguments,
+        state: TrainerState,
+        control: TrainerControl,
+        **kwargs,
+    ):
+        # Training arguments, state and controls are folded into kwargs to be passed off to
+        # handlers
+        kwargs["args"] = args
+        kwargs["state"] = state
+        kwargs["control"] = control
+        self._actions_on_event(event_name="on_save", **kwargs)
 
     def on_epoch_begin(
         self,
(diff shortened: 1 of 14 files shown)


tuning/trainercontroller/callback.py
```diff
@@ -425,6 +432,20 @@ def on_step_end(
...
+        args: TrainingArguments,
+        state: TrainerState,
+        control: TrainerControl,
+        **kwargs,
+    ):
+        # Training arguments, state and controls are folded into kwargs to be passed off to
+        # handlers
+        kwargs["args"] = args
+        kwargs["state"] = state
+        kwargs["control"] = control
```
ashokponkumar, line 446: Do we need to construct a new kwargs for this? Isn't just passing them in _actions_on_event just enough?

tuning/trainercontroller/controllermetrics/per_process_state.py
```diff
@@ -0,0 +1,80 @@
...
+                "on_step_end",
+                "on_epoch_begin",
+                "on_epoch_end",
+                "on_prediction_step",
+                "on_predict",
+                "on_log",
+                "on_train_end",
+                "on_train_begin",
+                "on_evaluate",
+                "on_save",
```
ashokponkumar, line 52: isn't doing this only on `on_init_end` sufficient?

tuning/trainercontroller/operations/logcontrol.py
```diff
@@ -0,0 +1,65 @@
...
+from transformers import TrainingArguments
+from transformers.utils import logging
+
+# Local
+from .operation import Operation
+
+logger = logging.get_logger(__name__)
+logger.setLevel(level=logging.DEBUG)
+
+VALID_LOG_LEVELS = {
```
ashokponkumar, line 11: Is https://github.com/python/cpython/blob/b93386e1c4dbc871148e081ae147f7d1e8d21444/Lib/logging/__init__.py#L126 usable here?

tuning/trainercontroller/operations/logcontrol.py
```diff
@@ -0,0 +1,65 @@
...
+            **kwargs,
+        )
+        if self.log_level == logging.ERROR:
+            logger.error(log_msg)
+        elif self.log_level == logging.WARNING:
+            logger.warning(log_msg)
+        elif self.log_level == logging.INFO:
+            logger.info(log_msg)
+        elif self.log_level == logging.DEBUG:
+            logger.debug(log_msg)
```
ashokponkumar, line 65: Can't we use a function like https://github.com/python/cpython/blob/b93386e1c4dbc871148e081ae147f7d1e8d21444/Lib/logging/__init__.py#L1594 instead?

ashokponkumar (changes requested):
Added some comments. Please incorporate them when you split the PR.

PR #227, opened 2024-07-02
Support custom embedding layer resizing to the desired multiple

### LINKED ISSUE #193: feat: resizing the embedding layer to the power of 2 or multiple of 8
Should we allow for resizing the embedding layer to power of 2 or multiple of 8 which ever works best? up for discussion.

kmehant:
### Description of the change

Allow resizing the model's embedding layer to the given multiple after addition of the tokenizer's new tokens for better memory efficiency.

### Related issue number

Closes #193

### How to verify the PR

You can use the embedding multiple option readily by passing the argument. 

```
--embedding_size_multiple_of <multiple> # 8 or 64 etc
```

### Was the PR tested

Sample training working and completed uses the `--embedding_size_multiple_of 64` option

Screenshot 2024-07-02 at 2 04 15 PM

Memory usage while training

Screenshot 2024-07-02 at 1 39 52 PM

Difference in memory with a bad multiple `--embedding_size_multiple_of 13`

Screenshot 2024-07-02 at 3 24 12 PM

Saved checkpoints
Screenshot 2024-07-02 at 2 04 15 PM

### DIFF
--- a/scripts/run_inference.py
+++ b/scripts/run_inference.py
@@ -182,30 +182,30 @@ class TunedCausalLM:
             with AdapterConfigPatcher(checkpoint_path, overrides):
                 try:
                     if base_model_name_or_path is None:
-                        raise KeyError("base_model_name_or_path has to be passed")
-                    modelbase = AutoModelForCausalLM.from_pretrained(
+                        raise ValueError("base_model_name_or_path has to be passed")
+                    base_model = AutoModelForCausalLM.from_pretrained(
                         base_model_name_or_path,
                         attn_implementation="flash_attention_2"
                         if use_flash_attn
                         else None,
                         torch_dtype=torch.bfloat16 if use_flash_attn else None,
                     )
-                    # since the peft library does not handle cases where the model's layers
-                    # are modified in PEFTModelForCausalLM in our case the embedding layer
+                    # since the peft library (PEFTModelForCausalLM) does not handle cases
+                    # where the model's layers are modified, in our case the embedding layer
                     # is modified, so we resize the backbone model's embedding layer with our own
                     # utility before passing it along to load the PEFT model.
                     tokenizer_data_utils.tokenizer_and_embedding_resize(
-                        {}, tokenizer=tokenizer, model=modelbase
+                        {}, tokenizer=tokenizer, model=base_model
                     )
                     model = PeftModel.from_pretrained(
-                        modelbase,
+                        base_model,
                         checkpoint_path,
                         attn_implementation="flash_attention_2"
                         if use_flash_attn
                         else None,
                         torch_dtype=torch.bfloat16 if use_flash_attn else None,
                     )
-                except (OSError, KeyError) as e:
+                except (OSError, ValueError) as e:
                     print("Failed to initialize checkpoint model!")
                     raise e
         except FileNotFoundError:
--- a/scripts/run_inference.py
+++ b/scripts/run_inference.py
@@ -28,10 +28,13 @@ import json
 import os
 
 # Third Party
-from peft import AutoPeftModelForCausalLM
+from peft import PeftModel
 from tqdm import tqdm
 from transformers import AutoModelForCausalLM, AutoTokenizer
 import torch
+
+# Local
+from tuning.data import tokenizer_data_utils
 
 
 ### Utilities
@@ -178,14 +181,31 @@ class TunedCausalLM:
         try:
             with AdapterConfigPatcher(checkpoint_path, overrides):
                 try:
-                    model = AutoPeftModelForCausalLM.from_pretrained(
+                    if base_model_name_or_path is None:
+                        raise KeyError("base_model_name_or_path has to be passed")
+                    modelbase = AutoModelForCausalLM.from_pretrained(
+                        base_model_name_or_path,
+                        attn_implementation="flash_attention_2"
+                        if use_flash_attn
+                        else None,
+                        torch_dtype=torch.bfloat16 if use_flash_attn else None,
+                    )
+                    # since the peft library does not handle cases where the model's layers
+                    # are modified in PEFTModelForCausalLM in our case the embedding layer
+                    # is modified, so we resize the backbone model's embedding layer with our own
+                    # utility before passing it along to load the PEFT model.
+                    tokenizer_data_utils.tokenizer_and_embedding_resize(
+                        {}, tokenizer=tokenizer, model=modelbase
+                    )
+                    model = PeftModel.from_pretrained(
+                        modelbase,
                         checkpoint_path,
                         attn_implementation="flash_attention_2"
                         if use_flash_attn
                         else None,
                         torch_dtype=torch.bfloat16 if use_flash_attn else None,
                     )
-                except OSError as e:
+                except (OSError, KeyError) as e:
                     print("Failed to initialize checkpoint model!")
                     raise e
         except FileNotFoundError:
--- a/tuning/sft_trainer.py
+++ b/tuning/sft_trainer.py
@@ -228,6 +228,7 @@ def train(
         special_tokens_dict=special_tokens_dict,
         tokenizer=tokenizer,
         model=model,
+        multiple=model_args.embedding_size_multiple,
     )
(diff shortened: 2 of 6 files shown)


scripts/run_inference.py
```diff
@@ -178,14 +181,31 @@ def load(
         try:
             with AdapterConfigPatcher(checkpoint_path, overrides):
                 try:
-                    model = AutoPeftModelForCausalLM.from_pretrained(
+                    if base_model_name_or_path is None:
+                        raise KeyError("base_model_name_or_path has to be passed")
+                    modelbase = AutoModelForCausalLM.from_pretrained(
+                        base_model_name_or_path,
```
fabianlim, line 187: its not very clear why cant you load the PEFT model first, then do a resize of the embedding layer. 

Also is there a use-case to do this for inference? Since `fms-hf-trainer` does not aim to support inference, it may be better to only do this for training, since the main concern is training throughput.

kmehant: > its not very clear why cant you load the PEFT model first, then do a resize of the embedding layer.

I see there are couple of ways to load PEFT model, but let me answer the problem with the usage of `AutoPeftModelForCausalLM` which is what being replaced in this PR. 

The base model is being internally resized by the PEFT library (https://github.com/huggingface/peft/blob/018a1f49c4a6e500c243efac61d72341879ea9f6/src/peft/auto.py#L128) based on the tokenizer, however, in our case, the tokenizer would not represent the true size of the embedding layer since it does not account for additional increase that was made to match the model embedding layer size to the nearest given multiple. 

Also, the current approach in this PR has been also recommended in their docs - https://huggingface.co/docs/peft/v0.8.2/en/developer_guides/troubleshooting#extending-the-vocabulary 

from docs
```
For inference, load the base model first and resize it the same way you did before you trained the model. After you’ve resized the base model, you can load the PEFT checkpoint.
```

Please let me know if there is a better way to handle this, thanks.


> Also is there a use-case to do this for inference? Since fms-hf-trainer does not aim to support inference, it may be better to only do this for training, since the main concern is training throughput.

@fabianlim The motivation behind these changes was to pass the unit tests around PEFT.

scripts/run_inference.py
```diff
@@ -178,14 +181,31 @@ def load(
         try:
             with AdapterConfigPatcher(checkpoint_path, overrides):
                 try:
-                    model = AutoPeftModelForCausalLM.from_pretrained(
+                    if base_model_name_or_path is None:
+                        raise KeyError("base_model_name_or_path has to be passed")
```
fabianlim, line 185: i think we shouldnt use `KeyError` here, it should be `ValueError`

tuning/sft_trainer.py
```diff
@@ -228,6 +228,7 @@ def train(
         special_tokens_dict=special_tokens_dict,
         tokenizer=tokenizer,
         model=model,
+        multiple=model_args.embedding_size_multiple,
```
fabianlim, line 231: maybe `multiple_of` instead of `multiple` is more readible?

scripts/run_inference.py
kmehant: Done!

tuning/sft_trainer.py
kmehant: I agree, modified!

scripts/run_inference.py
Ssukriti: thanks, while this script is not used in any product, we do use it in platform to quickly test changes in tuning with HF libraries instead of vLLLM deployment each time. So good to keep the script functional. Thank you for the change

Ssukriti (dismissed):
seems fine to me. Fabian can merge after his approval, as he was the primary reviewer. 

I hope fine tuning quality was also evaluated after change, since now default value is 8. It should not have caused any quality degradation using the default value

scripts/run_inference.py
fabianlim: @kmehant could you just document here which test was the failing one?

ashokponkumar:
Thanks @Ssukriti @fabianlim . Once you address Fabian's doc comment, let's merge it.

fabianlim:
@Ssukriti this change should not affect the behavior at all, this is because adding more tokens to the embedding layer, will still produce the *exact same output* 

My suggestion @kmehant is to just introduce a simple unit test to test `tokenizer_data_utils.tokenizer_and_embedding_resize`. I think currently this function is untested, but you can have a unit test for it, and also test the `multiple_of` behavior
- the expectation is tthat with or without specifying `multiple_of`, if you pass inputs a a model, we should see *the exact same output*. This will address @Ssukriti 's concerns.
- you can have the test in the same place as this one `https://github.com/foundation-model-stack/fms-hf-tuning/blob/b655e1ae6f0e7cf7f93933c0cc2dd72023a8d0f3/tests/utils/test_data_utils.py`
- you can load a small model for the test, refer to this https://github.com/foundation-model-stack/fms-hf-tuning/blob/b655e1ae6f0e7cf7f93933c0cc2dd72023a8d0f3/tests/test_sft_trainer.py#L44

scripts/run_inference.py
kmehant: sure @fabianlim

kmehant: @fabianlim @Ssukriti 
Tests that were failing.

1. test_run_causallm_pt_and_inference
2. test_run_causallm_pt_and_inference_with_formatting_data
3. test_run_causallm_lora_and_inference[default]
4. test_run_causallm_lora_and_inference[custom_target_modules]
5. test_run_causallm_lora_and_inference[all_linear_target_modules]

kmehant:
https://github.com/foundation-model-stack/fms-hf-tuning/pull/227#issuecomment-2207889784 

@fabianlim Added! thanks

scripts/run_inference.py
fabianlim: @kmehant so we are clear on why the above tests are failing. Since resizing embedding table does not change model output, so why do they fail?

kmehant: @fabianlim 
The reasoning is mentioned here - https://github.com/foundation-model-stack/fms-hf-tuning/pull/227#discussion_r1662692069

Here is the summary
```
FAILED tests/test_sft_trainer.py::test_run_causallm_pt_and_inference - RuntimeError: Error(s) in loading state_dict for PeftModelForCausalLM:
FAILED tests/test_sft_trainer.py::test_run_causallm_pt_and_inference_with_formatting_data - RuntimeError: Error(s) in loading state_dict for PeftModelForCausalLM:
FAILED tests/test_sft_trainer.py::test_run_causallm_pt_and_inference_JSON_file_formatter - RuntimeError: Error(s) in loading state_dict for PeftModelForCausalLM:
FAILED tests/test_sft_trainer.py::test_run_causallm_lora_and_inference[default] - RuntimeError: Error(s) in loading state_dict for PeftModelForCausalLM:
FAILED tests/test_sft_trainer.py::test_run_causallm_lora_and_inference[custom_target_modules] - RuntimeError: Error(s) in loading state_dict for PeftModelForCausalLM:
FAILED tests/test_sft_trainer.py::test_run_causallm_lora_and_inference[all_linear_target_modules] - RuntimeError: Error(s) in loading state_dict for PeftModelForCausalLM:
```

One trace for your reference - 
```
E           RuntimeError: Error(s) in loading state_dict for PeftModelForCausalLM:
E           	size mismatch for base_model.model.model.embed_tokens.weight: copying a param with shape torch.Size([32008, 64]) from checkpoint, the shape in current model is torch.Size([32001, 64]).
E           	size mismatch for base_model.model.lm_head.weight: copying a param with shape torch.Size([32008, 64]) from checkpoint, the shape in current model is torch.Size([32001, 64]).

.tox/py/lib/python3.9/site-packages/torch/nn/modules/module.py:2189: RuntimeError
```

fabianlim (approved):
Verified there was an error related to resizing, and this PR fixes that error. Test has been added to ensure no behavioural changes in model

fabianlim:
@Ssukriti I can merged this, but should it be a regular merge or squash merge?

PR #229, opened 2024-07-03
feat: support custom tokenizer

### LINKED ISSUE #190: feat: support custom tokenizers
At this point the `tokenizer_name_or_path` argument is part of the peft config which should not be the case since it can be used irrespective of using peft config like in the case of providing a custom tokenizer. Given this, we should have a separate field to provide custom tokenizer path may be part of the model args dataclass will allow users to use custom tokenizer which can be passed while creating the tokenizer instance.

If not provided this can be initialized to `model_name_or_path` when not provided which can be handled in the `__post_init__` lifecycle of the dataclass.

kmehant:
## Description

Currently we presume the tokenizer is always available at `model_name_or_path`https://github.com/foundation-model-stack/fms-hf-tuning/blob/b655e1ae6f0e7cf7f93933c0cc2dd72023a8d0f3/tuning/sft_trainer.py#L174. So, the handy solution for this is to simply introduce new argument `tokenizer_name_or_path` in the model arguments https://github.com/foundation-model-stack/fms-hf-tuning/blob/b655e1ae6f0e7cf7f93933c0cc2dd72023a8d0f3/tuning/config/configs.py#L34. However, this isn't the case for us since that argument is already being used in PEFT PT config https://github.com/foundation-model-stack/fms-hf-tuning/blob/b655e1ae6f0e7cf7f93933c0cc2dd72023a8d0f3/tuning/config/peft_config.py#L82

### Approach 1
Given the premise, one approach would be to use the argument `tokenizer_name_or_path` from PEFT PT config dataclass as is to set the tokenizer if None we fall back to `model_name_or_path`. However, passing this info as part of the PEFT config for a different purpose (not related to PEFT) would raise confusion within the code needs access to the tuning_config object and as well falling back piece of code would be outside the dataclass post_init logic.

### Approach 2
Another approach would be to move the argument `tokenizer_name_or_path` from PEFT PT config dataclass to model arguments dataclass. However, for obvious reasons`tokenizer_name_or_path` is needed in the PEFT PT data class to use it while preparing PEFT config and adding this argument from model arguments dataclass to PEFT config class is not workable since dataclasses do not include those variables that are outside the class definition while using `asdict`
https://github.com/foundation-model-stack/fms-hf-tuning/blob/b655e1ae6f0e7cf7f93933c0cc2dd72023a8d0f3/tuning/utils/config_utils.py#L87-L88 

### Approach 3 (currently implemented in this PR)
Simply maintain two separate variables for both the purposes and point out the differences in the respective help docs.

### Usage

simply pass `--custom_tokenizer_name_or_path <path to tokenizer>` otherwise it defaults to `model_name_or_path`

### Testing

Command
```bash
torchrun --nnodes="${WORLD_SIZE}" --node_rank="${RANK}" --nproc_per_node="4" --rdzv_id="101" --rdzv_endpoint="${MASTER_ADDR}:${MASTER_PORT}" ./tuning/sft_trainer.py --fsdp="full_shard auto_wrap"  --fsdp_config="/app/test/data/config.json" --output_dir="/app/test/data/output" --model_name_or_path="TinyLlama/TinyLlama-1.1B-Chat-v1.0" --custom_tokenizer_name_or_path="TinyLlama/TinyLlama-1.1B-Chat-v1.0" --training_data_path="/app/test/data/data.jsonl" --num_train_epochs 4  --per_device_train_batch_size 1  --per_device_eval_batch_size 1  --gradient_accumulation_steps 1  --evaluation_strategy "no"  --save_strategy "epoch"  --learning_rate 1e-5  --warmup_ratio 0.03  --lr_scheduler_type "cosine"  --logging_steps 1  --include_tokens_per_second  --packing False  --response_template "\n### Response:"  --dataset_text_field "output" --use_flash_attn False --torch_dtype float16
```

logs
Screenshot 2024-07-03 at 11 15 08 PM

Output checkpoints

Screenshot 2024-07-03 at 11 16 14 PM

Closes #190

### DIFF
--- a/tests/test_sft_trainer.py
+++ b/tests/test_sft_trainer.py
@@ -70,7 +70,6 @@ PEFT_PT_ARGS = peft_config.PromptTuningConfig(
     prompt_tuning_init="RANDOM",
     num_virtual_tokens=8,
     prompt_tuning_init_text="hello",
-    tokenizer_name_or_path=MODEL_NAME,
 )
 
 PEFT_LORA_ARGS = peft_config.LoraConfig(r=8, lora_alpha=32, lora_dropout=0.05)
@@ -175,7 +174,7 @@ def test_run_causallm_pt_and_inference():
         _validate_training(tempdir)
         checkpoint_path = _get_checkpoint_path(tempdir)
         adapter_config = _get_adapter_config(checkpoint_path)
-        _validate_adapter_config(adapter_config, "PROMPT_TUNING", PEFT_PT_ARGS)
+        _validate_adapter_config(adapter_config, "PROMPT_TUNING")
 
         # Load the model
         loaded_model = TunedCausalLM.load(checkpoint_path, MODEL_NAME)
@@ -208,7 +207,7 @@ def test_run_causallm_pt_and_inference_with_formatting_data():
         _validate_training(tempdir)
         checkpoint_path = _get_checkpoint_path(tempdir)
         adapter_config = _get_adapter_config(checkpoint_path)
-        _validate_adapter_config(adapter_config, "PROMPT_TUNING", PEFT_PT_ARGS)
+        _validate_adapter_config(adapter_config, "PROMPT_TUNING")
 
         # Load the model
         loaded_model = TunedCausalLM.load(checkpoint_path, MODEL_NAME)
@@ -239,7 +238,7 @@ def test_run_causallm_pt_and_inference_JSON_file_formatter():
         _validate_training(tempdir)
         checkpoint_path = _get_checkpoint_path(tempdir)
         adapter_config = _get_adapter_config(checkpoint_path)
-        _validate_adapter_config(adapter_config, "PROMPT_TUNING", PEFT_PT_ARGS)
+        _validate_adapter_config(adapter_config, "PROMPT_TUNING")
 
         # Load the model
         loaded_model = TunedCausalLM.load(checkpoint_path, MODEL_NAME)
@@ -261,7 +260,6 @@ def test_run_causallm_pt_init_text():
         tuning_config = peft_config.PromptTuningConfig(
             prompt_tuning_init="TEXT",
             prompt_tuning_init_text="hello",
-            tokenizer_name_or_path=MODEL_NAME,
         )
 
         sft_trainer.train(MODEL_ARGS, DATA_ARGS, train_args, tuning_config)
@@ -270,7 +268,7 @@ def test_run_causallm_pt_init_text():
         _validate_training(tempdir)
         checkpoint_path = _get_checkpoint_path(tempdir)
         adapter_config = _get_adapter_config(checkpoint_path)
-        _validate_adapter_config(adapter_config, "PROMPT_TUNING", tuning_config)
+        _validate_adapter_config(adapter_config, "PROMPT_TUNING")
 
 
 invalid_params_map = [
@@ -364,7 +362,7 @@ def test_run_causallm_lora_and_inference(request, target_modules, expected):
         _validate_training(tempdir)
         checkpoint_path = _get_checkpoint_path(tempdir)
         adapter_config = _get_adapter_config(checkpoint_path)
-        _validate_adapter_config(adapter_config, "LORA", base_lora_args)
+        _validate_adapter_config(adapter_config, "LORA")
 
         for module in expected:
             assert module in adapter_config.get("target_modules")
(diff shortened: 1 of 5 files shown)


tuning/config/configs.py
```diff
@@ -38,6 +38,19 @@ class ModelArguments:
         metadata={"help": "Use Flash attention v2 from transformers, default is True"},
     )
     torch_dtype: Optional[Union[torch.dtype, str]] = torch.bfloat16
+    custom_tokenizer_name_or_path: Optional[str] = field(
```
ashokponkumar, line 41: Is it not possible to use `tokenizer_name_or_path`. Is it not possible to bypass the issue with PEFTConfig?

tuning/config/peft_config.py
```diff
@@ -79,4 +79,11 @@ class PromptTuningConfig:
     prompt_tuning_init: str = "TEXT"
     num_virtual_tokens: int = 8
     prompt_tuning_init_text: str = "Classify if the tweet is a complaint or not:"
-    tokenizer_name_or_path: str = "llama-7b-hf"
+    tokenizer_name_or_path: str = field(
```
ashokponkumar, line 82: can this value be set automatically based on the value from the other tokenizer path?

ashokponkumar (commented):
Few questions

tuning/config/configs.py
fabianlim: yes i agree, best is to not have 2 flags with does basically the same thing. This will involve a design change, so best to discuss with maintainers.

praveenj83:
@kmehant @ashokponkumar Can we move this discussion to the issue, so that others can chime in on the design? Since it affects the user args and the conflict with an existing config by the same name in PEFT config, best to discuss.

tuning/config/peft_config.py
praveenj83: Yes, I would prefer if we can copy the value explicitly from ModelArguments, if that is set

kmehant: Just keeping the threads updated here as well.

This isn't possible since, copying the arg `tokenizer_name_or_path` explicitly from ModelArguments to PromptTuningConfig would not work since the current code unfolds it into kwargs to prepare the actual config object using `asdict` 
https://github.com/foundation-model-stack/fms-hf-tuning/blob/b655e1ae6f0e7cf7f93933c0cc2dd72023a8d0f3/tuning/utils/config_utils.py#L88
which stricly captures the arguments defined in the dataclass definition and does not capture those that are added over the object. This has been explained in Approach 2 of the PR description.

Ssukriti:
@kmehant we should go with option 2 as you mentioned above, rather than duplicating same parameters. this is how design can look like 

1. Modelarguments dataclass can contain a new optional parameter tokenizer_name_or_path. If not set, it will default to model name or path. 
2. Parameter will be removed from our PromptTuningConfig dataclass. Parameter is only need in HF PEFT prompttuningConfig class . so in this function https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/tuning/utils/config_utils.py#L74 -> we can pass parameter tokenizer_name_or_path and while creating HF object you can do 
```
`hf_peft_config = PromptTuningConfig(
            task_type=task_type, **asdict(tuning_config), tokenizer_name_or_path=tokenizer_name_or_path
        )`
```

tuning/config/peft_config.py
Ssukriti: hf_peft_config = PromptTuningConfig(
            task_type=task_type, **asdict(tuning_config), tokenizer_name_or_path=tokenizer_name_or_path
        )

kmehant:
@Ssukriti Thanks for finalizing on the approach, I agree with the changes you have suggested and updated this PR. Please review! Thanks.

fabianlim:
LGTM

tests/test_sft_trainer.py
```diff
@@ -431,17 +429,9 @@ def _get_adapter_config(dir_path):
         return json.load(f)
 
 
-def _validate_adapter_config(adapter_config, peft_type, tuning_config):
+def _validate_adapter_config(adapter_config, peft_type):
     assert adapter_config.get("task_type") == "CAUSAL_LM"
     assert adapter_config.get("peft_type") == peft_type
-    assert (
-        (
```
Ssukriti, line 438: this test is checking if adapters produced after tuning has the correct fields. I think its worthwhile to check if `adapter_config.get("tokenizer_name_or_path") == MODEL_ARGS.tokenizer_name_or_path`

def _validate_adapter_config(adapter_config, peft_type, MODEL_ARGS.tokenizer_name_or_path)

Ssukriti (changes requested):
Minor requested change in tests, then will approve

kmehant:
@Ssukriti updated with your feedback thanks.

Ssukriti (approved):

PR #230, opened 2024-07-04
feat: support any dataset file format and directories

kmehant:
## Description

Currently `fms-hf-tuning` only supports single file datasets in json/jsonl format. This PR expands the scope to any data file format and to datasets present in nested directories leveraging inbuilt HF Datasets capabilities.

### DIFF
--- a/tuning/utils/data_load_utils.py
+++ b/tuning/utils/data_load_utils.py
@@ -1,0 +1,25 @@
+# Standard
+from typing import Union
+import os
+
+# Third Party
+import datasets
+
+
+def load_dataset(data_path: str) -> Union[datasets.Dataset, None]:
+    """loads datasets given as either a file or a directory
+
+    Args:
+        data_path (str): path to the dataset file or directory
+
+    Returns:
+        datasets.Dataset: loaded dataset
+    """
+    if not data_path:
+        return None
+
+    if os.path.isdir(data_path):
+        return datasets.load_dataset(path=data_path, split="train")
+    return datasets.load_dataset(
+        path=os.path.dirname(data_path), data_files=data_path, split="train"
+    )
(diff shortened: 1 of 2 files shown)


tuning/utils/data_load_utils.py
```diff
@@ -0,0 +1,25 @@
...
+    Returns:
+        datasets.Dataset: loaded dataset
+    """
+    if not data_path:
+        return None
+
+    if os.path.isdir(data_path):
+        return datasets.load_dataset(path=data_path, split="train")
+    return datasets.load_dataset(
+        path=os.path.dirname(data_path), data_files=data_path, split="train"
```
fabianlim, line 24: this looks abit strange to me. 
- what are about relative paths?
- why path is `dirname` but `data_files` can be the full path?

fabianlim: Im not sure why this function is needed, because I believe I can pass a file in with the old API and it will read it. you mean you want to support multiple files?

kmehant: @fabianlim 
> what are about relative paths?

1. Do we support relative paths currently? Also, should we even support relative paths instead error if its not an absolute path?
2. For data directory path case, relative paths should be fine, however, for file path case, it will fail.

I can add a validation part before this code to allow only for absolute paths

> why path is dirname but data_files can be the full path?

path kwarg is a strange one since
1. if it is pointed to a file, then it would consider it as a data loader script
2. if it is pointed to a folder, then it would consider it as data source directory
3. if it is pointed to a folder along with data_files then it would use that to prepare dataset name and data_files as source of data
4. if it is pointed to a word like "json" then it would use that as the data format, in the above 3 cases it would automatically infer the data format unlike point 4.

To alleviate confusion here, I can add such comments to wrapping this code. Let me know.

kmehant:
@dushyantbehl @ashokponkumar you can close this once superseded, thanks.

dushyantbehl:
This should be closed @kmehant

Issue #181, opened 2024-06-11
feat: [Trainer controller] Add trainer controller metric for per process state in multi-process training


seshapad:
## Is your feature request related to a problem? Please describe.

In multi-GPU training, multiple processes are created and each process has a state. Trainer controller does not have access to process specific state such as rank.

## Describe the solution you'd like

Create a trainer controller metric to expose the per-process state.

## Describe alternatives you've considered

NA

## Additional context

NA

seshapad:
Addressed in PR [239](https://github.com/foundation-model-stack/fms-hf-tuning/pull/239).

PR #239, opened 2024-07-13
feat: per process state metric

HarikrishnanBalagopal:
### Description of the change

Adds a metric that exposes per process state to the trainer controller.

### Related issue number

https://github.com/foundation-model-stack/fms-hf-tuning/issues/181

### How to verify the PR

`make test`

### Was the PR tested

### DIFF
--- a/tuning/trainercontroller/controllermetrics/per_process_state.py
+++ b/tuning/trainercontroller/controllermetrics/per_process_state.py
@@ -1,0 +1,77 @@
+# Copyright The IBM Tuning Team
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


ashokponkumar (approved):

ashokponkumar:
@kmehant please review

tuning/trainercontroller/controllermetrics/per_process_state.py
```diff
@@ -0,0 +1,80 @@
...
+            kwargs: Remaining event arguments
+
+        Returns:
+            dict. Trainer state as a dictionary
+        """
+        try:
+            rank = torch.distributed.get_rank()
+        except ValueError:
+            return {"rank": None}
+        return {"rank": rank}
```
kmehant, line 80:
```suggestion
        if torch.distributed.is_available() and torch.distributed.is_initialized():
            return {"rank": torch.distributed.get_rank()}
        return {"rank": None}
```

kmehant: optionally you can wrap this again in try and except but the value error should not occur with this check.

kmehant (changes requested):

kmehant (approved):
Thank you, process rank centric metric is supported.

kmehant (approved):
Thank you. Process rank centric state metric is supported.

PR #246, opened 2024-07-16
deps: Update transformers to latest and skip broken prompt tuning tests

Abhishek-TAMU:
### Description of the change

- set lower limit of transformers to 4.41.0 which uses TrainingArgument eval_strategy instead of evaluation_strategy (By @anhuong )
- set upper limit of transformers to 5.0, thus pulling in new versions of transformers (By @anhuong )
- Commenting of prompt tuning test cases.

### Related issue number

Previously we had seen degraded performance in v4.41 of transformers as described in issue https://github.com/foundation-model-stack/fms-hf-tuning/issues/201. After further evaluation described in the issue, the throughput degradation on llama3-8B was only 17%, only on 2 A100, and not at all on 4 A100, and that memory utilization was significantly improved by moving to 4.41.

### Was the PR tested


tests/test_sft_trainer.py
```diff
@@ -168,108 +168,108 @@ def test_parse_arguments_peft_method(job_config):
 ############################# Prompt Tuning Tests #############################
 
 
-def test_run_causallm_pt_and_inference():
```
anhuong, line 171: Instead of commenting out these tests, lets add a pytest skip instead like `@pytest.mark.skip(reason="currently crashes before validation is done")` as seen [here](https://github.com/foundation-model-stack/fms-hf-tuning/blob/844b61f51005a012dba2620974a9fc218ff99e78/tests/test_sft_trainer.py#L607) and then provide the reason and error

anhuong:
In addition, please address the [linter error](https://github.com/foundation-model-stack/fms-hf-tuning/actions/runs/9965550583/job/27569608245?pr=246) and run `tox -e lint` to ensure the error is addressed.

tests/test_sft_trainer.py
Abhishek-TAMU: Thank you for reviewing the PR Anh. I have added this in all 3 test cases.

tuning/sft_trainer.py
```diff
@@ -322,6 +322,7 @@ def train(
     # this validation, we just drop the things that aren't part of the SFT Config and build one
     # from our object directly. In the future, we should consider renaming this class and / or
     # not adding things that are not directly used by the trainer instance to it.
+    train_args.use_cpu=True
```
anhuong, line 325: Please leave this change out of the PR, although useful note that transformers 4.42 requires this change to run unit tests on macs!

tox.ini
```diff
@@ -5,6 +5,7 @@ envlist = py, lint, fmt
 description = run unit tests
 deps =
     pytest>=7
+    setuptools
```
anhuong, line 8: Is setuptools needed to run unittests?

anhuong (changes requested):

tuning/sft_trainer.py
Abhishek-TAMU: Sure thanks!

tox.ini
Abhishek-TAMU: On my pc it is required, but I am not sure if its required in the build. Hence removing it for now.

anhuong (changes requested):
LGTM thanks Abhishek, please ensure before committing to run linters to ensure they pass `tox -e fmt,lint`. The formatter one is failing

anhuong (approved):
LGTM, thanks Abhishek!

pyproject.toml
```diff
@@ -28,7 +28,7 @@ classifiers=[
 dependencies = [
 "numpy>=1.26.4,<2.0",
 "accelerate>=0.20.3,<0.40",
-"transformers>=4.34.1,<=4.40.2,!=4.38.2",
+"transformers>=4.41.0,<5.0,!=4.38.2",
```
Ssukriti, line 31: @anhuong @Abhishek-TAMU 

transformers 4.41 has a severe performance degbradation, hence we want to keep it > 4.41 and not >=

also `!=4.38.2",` is not needed now as it is lesser than 4.41 and does not satisfy condition

Abhishek-TAMU: This new [PR](https://github.com/foundation-model-stack/fms-hf-tuning/pull/250/) reflects this change.

PR #248, opened 2024-07-17
Replace shutil.copytree()

olson-ibm:
### Summary
Custom implementation of `shutil.copytree()` that does not error out in the event file metadata cannot be transferred.  

It was reported that if the `output_dir` parameter was set to the root of a NFS mount, `shutil.copytree()` would error out due to failure of the file metadata to transfer due to permissioning on the root volume. This error was successfully reproduced. Output files were successfully written to the root drive, but the copy failed due to file metadata not being written.

Code submitted in this PR should fix. There are a few issues with this implementation that will need to be addressed:

1. There did not seem to be another option in the shutil package that copied the contents of a directory. `copyfile` and `copy2` both copy individual files. `copy2` was used here to preserve the file metadata when possible. In the event the file metadata cannot be preserved, do not error out.

2.  Is there a better place to put this new function rather than in `accelerate_launch.py`?

### How to verify the PR

Execute a tuning session where `output_dir` is set to the root volume of a NFS drive. Tuning session should complete without error and all tuning artifacts should be created in the root directory.

### Was the PR tested

No unit tests were added. The following test plan was executed successfully:

1. Verify the original error by using a SFT tuning session with the output set to a root NFS mount. Verify the tuning session errors out as expected (**COMPLETED SUCCESSFULLY**)
2. Verifying the fix is in place, repeat the above, ensuring the tuning session does not error out and all expected tuning artifacts are created in the expected location. (**COMPLETED SUCCESSFULLY**)
3. Repeat, with the target directory not being the root NFS directory, but a subdirectory off root. Verify all expected tuning artifacts are created in the expected location. (**COMPLETED SUCCESSFULLY**)
4. Repeat with a COS bucket mount, rather than a NFS mount. Verify all expected tuning artifacts are created in the expected location. (**COMPLETED SUCCESSFULLY**)

Documentation of all of the above (deploy files, log files, and screenshots) can be provided upon request.

If we need to add unit tests for this, or better document the results of the above test cases, please advise.

### DIFF
--- a/build/accelerate_launch.py
+++ b/build/accelerate_launch.py
@@ -25,6 +25,7 @@ import sys
 import traceback
 import tempfile
 import shutil
+import stat
 from pathlib import Path
 
 # Third Party
@@ -46,6 +47,39 @@ from tuning.utils.error_logging import (
 
 ERROR_LOG = "/dev/termination-log"
 
+def copytree(source, destination, symlinks = False, ignore = None):
+  
+  if not os.path.exists(destination):
+    os.makedirs(destination)
+    shutil.copystat(source, destination)
+  file_list = os.listdir(source)
+  # If we want to ignore any files
+  if ignore:
+    exclude = ignore(source, file_list)
+    file_list = [a_file for a_file in file_list if a_file not in exclude]
+  # Have a list of directory objects, now iterate over them.
+  for item in file_list:
+    source_file = os.path.join(source, item)
+    destination_file = os.path.join(destination, item)
+    if symlinks and os.path.islink(source_file):
+      if os.path.lexists(destination_file):
+        os.remove(destination_file)
+      os.symlink(os.readlink(source_file), destination_file)
+      # Improvement over shutil.copytree: BAD CHMOD IS NOT FATAL.
+      try:
+        # get the permissions on the source file
+        st = os.lstat(source_file)
+        mode = stat.S_IMODE(st.st_mode)
+        # apply them to the destination file 
+        os.lchmod(destination_file, mode)
+      except:
+        pass # Not fatal.
+    elif os.path.isdir(source_file):
+      # recursive call for subdirectories
+      copytree(source_file, destination_file, symlinks, ignore)
+    else:
+      # straight copy.
+      shutil.copy2(source_file, destination_file)
 
 def main():
     LOGLEVEL = os.environ.get("LOG_LEVEL", "WARNING").upper()
@@ -124,7 +158,7 @@ def main():
                 pt_checkpoint_dir,
                 original_output_dir,
             )
-            shutil.copyfile(
+            copytree(
                 os.path.join(tempdir, pt_checkpoint_dir),
                 original_output_dir
             )
--- a/build/accelerate_launch.py
+++ b/build/accelerate_launch.py
@@ -124,10 +124,9 @@ def main():
                 pt_checkpoint_dir,
                 original_output_dir,
             )
-            shutil.copytree(
+            shutil.copyfile(
                 os.path.join(tempdir, pt_checkpoint_dir),
-                original_output_dir,
-                dirs_exist_ok=True,
+                original_output_dir
             )
         except Exception as e:  # pylint: disable=broad-except
             logging.error(traceback.format_exc())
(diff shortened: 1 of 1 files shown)


anhuong:
Thanks for the changes @olson-ibm ! Please run the formatter/linter to fix the failures noted. In addition, please sign your commits in order for the DCO error to be resolved. 

> Is there a better place to put this new function rather than in accelerate_launch.py?
You can put the method in [build/utils.py](https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/build/utils.py) which is where we put functions used by `accelerate_launch.py` and it is copied in to the final image. But I agree that it should not be in the `tuning` library and since the copying is only needed in the image, the method should live with the build scripts. It would be great if you can add unit tests [here](https://github.com/foundation-model-stack/fms-hf-tuning/tree/main/tests/build) for the method as well.

build/accelerate_launch.py
```diff
@@ -47,6 +48,40 @@
 ERROR_LOG = "/dev/termination-log"
 
 
+def copytree(source, destination, symlinks=False, ignore=None):
+    if not os.path.exists(destination):
+        os.makedirs(destination)
+        shutil.copystat(source, destination)
```
anhuong, line 54: This is what we are trying to avoid -- copying the directory permissions. We don't need to copy the directory permissions, only the files over. The issue with copytree is that it copies the permissions and times of **directories** are copied with [copystat()](https://docs.python.org/3/library/shutil.html#shutil.copystat), individual files are copied using [copy2()](https://docs.python.org/3/library/shutil.html#shutil.copy2).

build/accelerate_launch.py
```diff
@@ -47,6 +48,40 @@
...
+    file_list = os.listdir(source)
+    # If we want to ignore any files
+    if ignore:
+        exclude = ignore(source, file_list)
```
anhuong, line 58: I like the idea of allowing files to ignore although this is not a feature we are yet adding as this would require parsing the JSON the user passes in and adding a new flag for files to ignore. In addition, what is this `ignore()` method?

build/accelerate_launch.py
```diff
@@ -47,6 +48,40 @@
...
+    for item in file_list:
+        source_file = os.path.join(source, item)
+        destination_file = os.path.join(destination, item)
+        if symlinks and os.path.islink(source_file):
+            if os.path.lexists(destination_file):
+                os.remove(destination_file)
+            os.symlink(os.readlink(source_file), destination_file)
+            # Improvement over shutil.copytree: BAD CHMOD IS NOT FATAL.
+            try:
+                # get the permissions on the source file
```
anhuong, line 70: The issue we were seeing is with the directory permissions, not the file permissions. I don't think the file permissions are an issue at all and copytree already used `copy2()` to copy the files over. 

If this is symlinking the file, does that not already copy the permissions of the file over?

anhuong: Also I don't think you need to worry about symlinks because `shutil.copy()` and `shutil.copy2()` and `shutil.copyfile()` all allow the option to set `follow_symlinks=True` and covers this for you

build/accelerate_launch.py
```diff
@@ -47,6 +48,40 @@
...
+                # apply them to the destination file
+                os.lchmod(destination_file, mode)
+            except Exception as e:  # pylint: disable=broad-except
+                pass  # Not fatal.
+        elif os.path.isdir(source_file):
+            # recursive call for subdirectories
+            copytree(source_file, destination_file, symlinks, ignore)
+        else:
+            # straight copy.
+            shutil.copy2(source_file, destination_file)
```
anhuong, line 82: In the issue, we had noted that we didn't want to copy the file permissions over. I don't think the file permissions were the cause of the issue but that's why we were suggesting to use `copyfiles` instead which does not copy the file permissions.

> we should move away from copytree to using copyfiles instead that does not copy permissions
> https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/build/accelerate_launch.py#L127
> copyfiles https://docs.python.org/3/library/shutil.html#shutil.copyfile

anhuong (changes requested):
Thanks for the changes Joe! A few issues in regards to the root of the problem we were hitting into and whether we want to copy file permissions over.

build/accelerate_launch.py
anhuong: Ah is the main issue that you replicated is that if the destination directory already exists (such as if writing directly to the PVC path mounted and not an inner directory) then when copying over the final checkpoint the script failed? But then if the directory didn't already exists then the script would create the directory and that succeeds?

olson-ibm:
Closing. WIll re-open with properly signed commits, a simpler implementation and better test cases within the hour.

anhuong:
#251 contains new PR

Issue #62, opened 2024-02-27
bug: sft_trainer.py::train() raises exception due to change in transformers 4.38.0


VassilisVassiliadis:
**Issue**

The `__init__()` method of the transformers SFTTrainer class raises  an exception when using the latest commit of fms-hf-tuning (f4e8eb4db560fb2968c8c5e2860cf913d03c9676) and `transformers>=4.38.0`.

## Context

- Currently, the dependency to transformers in the `requirements.txt` file is `transformers>=4.34.1`
- The commit that makes fms-hf-tuning incompatible with transformers is https://github.com/huggingface/transformers/commit/5f06053dd821c91f7bd697309109abaa3396b605 
    - this got released at v 4.38.0

## Possible solutions

1. Update `fms-hf-tuning` source code to be compatible with `transformers>=4.38.0`
2. ~Update `requirements.txt` so that it uses `transformers<4.38.0`~

**Done when**
- [ ] sft_trainer.py works with the dependencies that get installed when installing fms-hf-tuning

## Example:

commit: f4e8eb4db560fb2968c8c5e2860cf913d03c9676

```bash
model="hf-tiny-model-private/tiny-random-BloomForCausalLM"
python tuning/sft_trainer.py --output_dir tuned --model_name_or_path ${model} --use_flash_attn=false --num_train_epochs 5 \
        --evaluation_strategy no --response_template "\n### Response" --dataset_text_field output --tokenizer_name_or_path ${model} \
        --torch_dtype float32 --data_path common_en_news_combined_512-preprocessed.jsonl \
         --target_modules all-linear --peft_method lora 
```

Exception:

```
Traceback (most recent call last):
  File "/projects/fms-hf-tuning/tuning/sft_trainer.py", line 308, in <module>
    fire.Fire(main)
  File "/projects/fms-hf-tuning/venv/lib/python3.10/site-packages/fire/core.py", line 141, in Fire
    component_trace = _Fire(component, args, parsed_flag_args, context, name)
  File "/projects/fms-hf-tuning/venv/lib/python3.10/site-packages/fire/core.py", line 475, in _Fire
    component, remaining_args = _CallAndUpdateTrace(
  File "/projects/fms-hf-tuning/venv/lib/python3.10/site-packages/fire/core.py", line 691, in _CallAndUpdateTrace
    component = fn(*varargs, **kwargs)
  File "/projects/fms-hf-tuning/tuning/sft_trainer.py", line 304, in main
[...]

PR #250, opened 2024-07-18
deps: Update transformers lower bound version

Abhishek-TAMU:
### Description of the change

set the transformers version greater than 4.41 as it may have severe performance degradation. (@Ssukriti) 
set upper limit of transformers to 5.0, and removing this version constraint (!=4.38.2)

### Related issue number

Fix of #246 

### Was the PR tested

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,7 +28,7 @@ classifiers=[
 dependencies = [
 "numpy>=1.26.4,<2.0",
 "accelerate>=0.20.3,<0.40",
-"transformers>=4.41.0,<5.0,!=4.38.2",
+"transformers>4.41,<5.0",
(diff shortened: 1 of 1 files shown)


anhuong:
@Abhishek-TAMU your commit history from this branch contains too many commits, this should only include the single commit with your change for the lower bound of teh transformers version

anhuong (approved):
Thanks for the quick update Abhishek! As Abhishek has noted, we've hit into some issues with some transformers versions. v4.41 has performance degradation [issue](https://github.com/foundation-model-stack/fms-hf-tuning/issues/201) and v4.38.2 had a [bug](https://github.com/foundation-model-stack/fms-hf-tuning/issues/62) with the training argument flags. This ensures we do not use those versions.

PR #251, opened 2024-07-18
Replace shutil.copytree() to fix permission error

olson-ibm:
### Description of the change

Intended to reproduce the functionality of `shutil.copytree()`, but uses `shutil.copy2()` to get around the problem  `shutil.copytree()` has of trying to modify the directory metadata, which fails on a root NFS directory mounted and owned by root in a SFT tuning pod.

### How to verify the PR

Run a SFT tuning session where the `output_dir` is set to a root owned, everyone writable NFS volume.

### Was the PR tested

Unit tests are not submitted as part of the PR, but can be found [here](https://github.com/olson-ibm/copytree/blob/main/test/test.py). All passed.

Integration tests also completed successfully, can provide the test cases and results upon request.

### DIFF
--- a/build/utils.py
+++ b/build/utils.py
@@ -24,18 +24,17 @@ from accelerate.commands.launch import launch_command_parser
 import shutil
 
 
-def copytree(source, destination):
+def copy_checkpoint(source, destination):
     if not os.path.exists(destination):
         os.makedirs(destination)
         shutil.copystat(source, destination)
-    file_list = os.listdir(source)
     # Have a list of directory objects, now iterate over them.
-    for item in file_list:
+    for item in os.listdir(source):
         source_file = os.path.join(source, item)
         destination_file = os.path.join(destination, item)
         if os.path.isdir(source_file):
             # recursive call for subdirectories
-            copytree(source_file, destination_file)
+            copy_checkpoint(source_file, destination_file)
         else:
             # straight copy.
             shutil.copy2(source_file, destination_file)
--- a/build/utils.py
+++ b/build/utils.py
@@ -21,6 +21,24 @@ import base64
 # Third Party
 import torch
 from accelerate.commands.launch import launch_command_parser
+import shutil
+
+
+def copytree(source, destination):
+    if not os.path.exists(destination):
+        os.makedirs(destination)
+        shutil.copystat(source, destination)
+    file_list = os.listdir(source)
+    # Have a list of directory objects, now iterate over them.
+    for item in file_list:
+        source_file = os.path.join(source, item)
+        destination_file = os.path.join(destination, item)
+        if os.path.isdir(source_file):
+            # recursive call for subdirectories
+            copytree(source_file, destination_file)
+        else:
+            # straight copy.
+            shutil.copy2(source_file, destination_file)
 
 
 def get_highest_checkpoint(dir_path):
--- a/tests/build/test_utils.py
+++ b/tests/build/test_utils.py
@@ -20,9 +20,12 @@ from unittest.mock import patch
 
 # Third Party
 import pytest
+import shutil
+import filecmp
 
 # Local
 from build.utils import process_accelerate_launch_args
+from build.utils import copy_checkpoint
 
 HAPPY_PATH_DUMMY_CONFIG_PATH = os.path.join(
     os.path.dirname(__file__), "dummy_job_config.json"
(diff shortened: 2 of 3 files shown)


build/utils.py
```diff
@@ -21,6 +21,24 @@
 # Third Party
 import torch
 from accelerate.commands.launch import launch_command_parser
+import shutil
+
+
+def copytree(source, destination):
```
anhuong, line 27: Can you include some of the unit tests you made for running this method in [test_utils](https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/tests/build/test_utils.py). A few useful test cases would be:
- Creating a tmp directory with foo and bar files and then running copytree on it and ensuring those two files exist in destination
-  Creating a tmp directory with foo directory and bar inside like /tmp/foo/bar and then running copytree on it and ensuring dir and file exists in destination
- Ensure it works even if destination dir already exists
- Ensure it works with symlinks as needed

build/utils.py
```diff
@@ -21,6 +21,24 @@
...
+    if not os.path.exists(destination):
+        os.makedirs(destination)
+        shutil.copystat(source, destination)
+    file_list = os.listdir(source)
+    # Have a list of directory objects, now iterate over them.
+    for item in file_list:
```
anhuong, line 33: nit refactor:
```suggestion
    for item in os.listdir(source):
```

build/utils.py
```diff
@@ -21,6 +21,24 @@
...
+        source_file = os.path.join(source, item)
+        destination_file = os.path.join(destination, item)
+        if os.path.isdir(source_file):
+            # recursive call for subdirectories
+            copytree(source_file, destination_file)
+        else:
+            # straight copy.
+            shutil.copy2(source_file, destination_file)
```
anhuong, line 41: What about symlinks? Do we want to set `follow_symlinks=True`? If a symlink (for some reason) is in the source directory, would this method fail or would the symlink just not get copied over?

anhuong (changes requested):
Thanks for the change, simplified method, and well documented testing Joe! Adding some simple unit tests would be helpful. Note that this new copytree method still copies the contents of the directory over with the same permissions as this was not the main issue with copytree. The main issue was that copytree tried to copy the top level directory permissions over.

build/utils.py
olson-ibm: Yes I have all of those already, [here](https://github.com/olson-ibm/copytree/blob/main/test/test.py).

I'll get them over ASAP.

build/utils.py
olson-ibm: fine

build/utils.py
olson-ibm: I pulled the symlink functionality. Do we need it added back?

anhuong: Symlinks are probably not needed as the checkpoints don't contain symlinks. But I'm just curious if setting `follow_symlinks=True` does copy over symlinks. In addition, if this method is just for copying the checkpoint, then I would rename the method to `copy_checkpoint()` instead of copytree to avoid confusion

build/utils.py
anhuong: - Success case if dest dir does not exist
- Success case if dest dir does already exist
- Failure case if dest dir not writeable
- Failure case if source dir does not exist

tests/build/test_utils.py
```diff
@@ -108,3 +111,110 @@ def test_process_accelerate_launch_custom_config_file(patch_path_exists):
...
+def test_copy_checkpoint_dest_dir_does_not_exist():
+
+    # Init source directory
+    config = CopyCheckpointTestConfig("/tmp")
+
+    target_dir_does_not_exist = os.path.join(config.test_root, "test_copytree_target")
+
+    # Execute the copy
+    copy_checkpoint(config.source_dir, target_dir_does_not_exist)
+    assert config.are_dir_trees_equal(config.source_dir, target_dir_does_not_exist)
```
anhuong, line 170: I was hopeful that this could be done with something like...
```py
assert list(pathlib.Path(source).rglob("*")) ==  list(pathlib.Path(dest).rglob("*"))
```
but alas, my code fails because the root directories are different and it doesn't check the contents of the files either. Unfortunate :/

tests/build/test_utils.py
```diff
@@ -108,3 +111,110 @@ def test_process_accelerate_launch_custom_config_file(patch_path_exists):
...
+        open(os.path.join(self.source_sub_dir, "tf5.txt"), "a").close()
+        open(os.path.join(self.source_sub_dir, "tf6.txt"), "a").close()
+
+    def are_dir_trees_equal(self, dir1, dir2):
+
+        dirs_cmp = filecmp.dircmp(dir1, dir2)
+        if (
+            len(dirs_cmp.left_only) > 0
+            or len(dirs_cmp.right_only) > 0
+            or len(dirs_cmp.funny_files) > 0
```
anhuong, line 145: Can we also check that `len(dirs_cmp.common_files) > 0`, when I test out `filecmp.dircmp(dir1, dir2)` it seems to only list the first level of files and not subdirs

tests/build/test_utils.py
```diff
@@ -108,3 +111,110 @@ def test_process_accelerate_launch_custom_config_file(patch_path_exists):
...
+
+
+def test_copy_checkpoint_dest_dir_does_exist():
+
+    # Init source directory
+    config = CopyCheckpointTestConfig("/tmp")
+
+    # Init target directory
+    target_dir_does_exist = os.path.join(config.test_root, "test_copytree_target2")
+    os.mkdir(target_dir_does_exist)
```
anhuong, line 184: Could be good to add a file into target dir to show that the copying overwrites the existing directory

tests/build/test_utils.py
```diff
@@ -108,3 +111,110 @@ def test_process_accelerate_launch_custom_config_file(patch_path_exists):
...
+
+    # Init target directory
+    target_dir_not_writeable = os.path.join(
+        config.test_root, "test_copytree_notwriteable"
+    )
+
+    os.makedirs(target_dir_not_writeable, mode=0o446)
+
+    # Execute the copy. Should FAIL
+    with pytest.raises(Exception) as e:
```
anhuong, line 208: Better to be more specific since we are expecting exception to be raised is `PermissionError`

tests/build/test_utils.py
```diff
@@ -108,3 +111,110 @@ def test_process_accelerate_launch_custom_config_file(patch_path_exists):
...
+        copy_checkpoint(config.source_dir, target_dir_not_writeable)
+    assert "Permission denied:" in str(e.value)
+
+    # Cleanup
+    shutil.rmtree(config.source_dir)
+    shutil.rmtree(target_dir_not_writeable)
+
+
+def test_copy_checkpoint_source_dir_does_not_exist():
+    with pytest.raises(Exception) as e:
```
anhuong, line 218: Better to be more specific since we expect this to raise `FileNotFoundError`

tests/build/test_utils.py
anhuong: The only alternative checking I was thinking of is...
```py
for root,dir,files in os.walk(source_dir):
     new_root = root.replace(source_dir, dest_dir)
     for file in files:
             assert os.path.exists(os.path.join(new_root, file))
```
but is not perfect either, so I think if yours works, good to go with it!

anhuong (changes requested):
Thanks for adding tests Joe! A few comments..

tests/build/test_utils.py
olson-ibm: Hmm.... `dirs_cmp.left_only` and `dirs_cmp.right_only` should include both files and subdirectories, and this gets called recursively throughout the file tree. 

In the event any level (including the topmost) are not equal, this bombs out at that level with a False:

Output from a two level check where everything is equal:
```
validating /tmp/test_copytree_source against /tmp/test_copytree_target
	dirs_cmp.right_only: []
	dirs_cmp.left_only: []
validating /tmp/test_copytree_source/subdir1 against /tmp/test_copytree_target/subdir1
	dirs_cmp.right_only: []
	dirs_cmp.left_only: []
True
```

Output from a two level check where the top level fails:
```
validating /tmp/test_copytree_source against /tmp/test_copytree_target
	dirs_cmp.right_only: []
	dirs_cmp.left_only: ['subdir1', 'tf1.txt', 'tf2.txt', 'tf3.txt']
False
```

Code with the debug prints:
```
def are_dir_trees_equal(dir1, dir2):
    print(f'validating {dir1} against {dir2}').       #OUTPUT
    dirs_cmp = filecmp.dircmp(dir1, dir2)
    print(f'\tdirs_cmp.right_only: {dirs_cmp.right_only}').  #OUTPUT
    print(f'\tdirs_cmp.left_only: {dirs_cmp.left_only}').      #OUTPUT

    if (
        len(dirs_cmp.left_only) > 0
        or len(dirs_cmp.right_only) > 0
        or len(dirs_cmp.funny_files) > 0
    ):
        return False
    (_, mismatch, errors) = filecmp.cmpfiles(
        dir1, dir2, dirs_cmp.common_files, shallow=False
    )
    if len(mismatch) > 0 or len(errors) > 0:
        return False
    for common_dir in dirs_cmp.common_dirs:
        new_dir1 = os.path.join(dir1, common_dir)
        new_dir2 = os.path.join(dir2, common_dir)
        if not are_dir_trees_equal(new_dir1, new_dir2):
            return False
    return True
```

Moreover, `common_files` should return what files are common between the two directories, once copied, and thus, should return all files at the level. This should certainly be > 0 in a successfully copy, so we can't put that in the 'return False' case.

anhuong: Ah I misunderstood this function's purpose, I thought it was to validate that the two directories are the same, not that it was determining whether they were the same or not. And I see the recursion that goes into subdirectories. Thanks for the clarification!

tests/build/test_utils.py
```diff
@@ -127,15 +128,11 @@ def __init__(self, test_root):
         #         tf5.txt
         #         tf6.txt
 
-        assert os.path.isdir(test_root)
-
-        self.test_root = test_root
+        self.test_root = temp_root
+        print(f"ROOT: {self.test_root}")
```
anhuong, line 132: Please remove print statement

tests/build/test_utils.py
```diff
@@ -108,3 +112,134 @@ def test_process_accelerate_launch_custom_config_file(patch_path_exists):
...
+        # Add a file to the target. This file does not exist in source.
+        with open(
+            os.path.join(target_dir_does_exist, "tf9.txt"),
+            "a",
+            encoding="utf-8",
+        ) as f:
+            f.close()
+        # Execute the copy
+        copy_checkpoint(config.source_dir, target_dir_does_exist)
+        # Validate the file that did not exist in source is still there.
```
anhuong, line 216: This comment seems to be in conflict with the comment on line 200. So for clarity, if the destination dir already exists and has a file in it, will that file still exist after the copying of source to dest? I think the answer is yes unless it has the same name as a file from source, in which it would be overwritten?

anhuong: @olson-ibm I see you deleted this comment but that doesn't address the issue or my question

anhuong: Ah I once again misunderstood and in this test Joe is doing two things...he is seeing that a file that already exists in target_dir and has same name as a file getting copied over gets overwritten and he is seeing that a file that  already exists in target_dir and has diff name does not get overwritten. So this comment does make sense but is a little hard to follow with the filenames

anhuong (approved):
LGTM, thank you so much Joe!

PR #258, opened 2024-07-22
fix: remove lm_head for granite with llama arch models

Ssukriti:
### Description of the change

- Before copying the last checkpoint created to the users's output dir, we load the checkpoint, test to see if it has llama arch and that the lm_head weight matches the embed_weight, and if so delete the lm_head.weight.
   - Loads model with `AutoModelForCausalLM.from_pretrained()` if is a fine tuned model, otherwise loads base model and adapter with `PeftModel.from_pretrained()`.
   - After deleting lm_head.weight, saves the model to output_dir with ` model.save_pretrained()`

### Related issue number

We found that for granite models with llama architecture (such as granite-3b-code) has lm_head weights tied to embed weights and when tuning with accelerate and FSDP for multi-GPU, the lm_head weight was created as a duplicate of the embed weight. Due to this unexpected weight being created, vLLM is unable to load the model. To fix this, we are deleting the lm_head weight.

### How to verify the PR

Ran updated `accelerate_launch.py` script in cluster with below configs:

#### granite-3b-code-base, fine tuning, multi-GPU

Successfully fine tuned model and removed lm_head. The tuned model was able to be loaded on vLLM (where verified previous lm_head error). Verified that lm_head is removed from checkpoint's `model.safetensors.index.json`.

tail tuning logs:
```sh
{'train_runtime': 585.5663, 'train_samples_per_second': 7.395, 'train_steps_per_second': 0.939, 'train_loss': 0.4583773977106268, 'epoch': 10.0}
100%|████████████████████████████████████████████████████████████████████████████████████| 550/550 [09:45<00:00,  1.06s/it]
You are attempting to use Flash Attention 2.0 with a model not initialized on GPU. Make sure to move the model to GPU after initializing it on CPU with `model.to('cuda')`.
Loading checkpoint shards: 100%|█████████████████████████████████████████████████████████████| 2/2 [00:16<00:00,  8.17s/it]
INFO:root:Removing lm_head from checkpoint
```

#### granite-3b-code, LoRA tuning, multi-GPU

Successfully LoRA tuned model and removed lm_head. The tuned model was able to be loaded on vLLM (where verified previous lm_head error).

tail tuning logs:
```sh
{'train_runtime': 281.5458, 'train_samples_per_second': 15.379, 'train_steps_per_second': 1.954, 'train_loss': 0.9379253248734908, 'epoch': 10.0}
100%|████████████████████████████████████████████████████████████████████████████████████| 550/550 [04:41<00:00,  1.95it/s]
DEBUG:urllib3.connectionpool:Starting new HTTPS connection (1): huggingface.co:443
DEBUG:urllib3.connectionpool:https://huggingface.co:443 "HEAD /ibm-granite/granite-3b-code-base/resolve/main/config.json HTTP/11" 200 0
You are attempting to use Flash Attention 2.0 with a model not initialized on GPU. Make sure to move the model to GPU after initializing it on CPU with `model.to('cuda')`.
Loading checkpoint shards: 100%|█████████████████████████████████████████████████████████████| 2/2 [00:12<00:00,  6.02s/it]
DEBUG:urllib3.connectionpool:https://huggingface.co:443 "HEAD /ibm-granite/granite-3b-code-base/resolve/main/generation_config.json HTTP/11" 200 0
INFO:root:Removing lm_head from checkpoint
```

#### granite-3b-code-base, prompt tuning, single-GPU

Successfully prompt tuned and as we know single GPU does not result in the lm_head issue and so it does not delete lm_head tuning runs.

#### llama-13b-base, fine tuned, multi-GPU

Successfully fine tuned and as expected, does not result in lm_head being deleted. Tuning logs show no message that lm_head is deleted.

### Was the PR tested


build/accelerate_launch.py
```diff
@@ -117,18 +118,48 @@ def main():
...
+            params_dict = dict(model.named_parameters())
+            if (
+                model_arch == "LlamaForCausalLM"
+                and hasattr(model, "lm_head")
+                and hasattr(model.model, "embed_tokens")
+                and model.lm_head.weight.untyped_storage().data_ptr()
+                == params_dict["model.embed_tokens.weight"].untyped_storage().data_ptr()
+            ):
+                logging.info("Removing lm_head from checkpoint")
+                del model.lm_head
```
Ssukriti, line 140: maybe just deleting the `lm_head.weight` should suffice?

build/accelerate_launch.py
```diff
@@ -117,18 +128,91 @@ def main():
...
+                    logging.info("Removing lm_head from checkpoint")
+                    del model.lm_head.weight
+
+                    if hasattr(model, "lm_head.weight"):
+                        logging.warning("Failed to delete lm_head.weight from model")
+
+                    logging.info("Saving checkpoint to %s", original_output_dir)
+                    model.save_pretrained(original_output_dir)
+                    # save tokenizer with model
+                    tokenizer.save_pretrained(original_output_dir)
```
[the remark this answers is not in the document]
anhuong: Finally, when I load the fine tuned model up in python, you do see lm_head.weight still even though it's deleted in `model.safetensors.index.json`. 

```py
>>> use_flash_attn = os.getenv("use_flash_attn", True)
>>> ft_model = "/data/anhuong/tuning_output/granite-3b-code-ft-test-remove-lm-head"
>>> ft_model = AutoModelForCausalLM.from_pretrained(ft_model, attn_implementation="flash_attention_2" if use_flash_attn else None, torch_dtype=bfloat16 if use_flash_attn else None)
Loading checkpoint shards: 100%|█████████████████████████████████████████████████████████████| 2/2 [00:33<00:00, 16.94s/it]
>>> ft_model.lm_head
Linear(in_features=2560, out_features=49152, bias=False)
>>> ft_model.lm_head.weight
Parameter containing:
tensor([[-0.0022,  0.0032,  0.0302,  ..., -0.0132, -0.0620,  0.0070],
        [ 0.0152,  0.0287, -0.0002,  ..., -0.0028,  0.0100,  0.0159],
        [ 0.0132, -0.0197,  0.0447,  ...,  0.0247,  0.0732, -0.0236],
        ...,
        [ 0.0067, -0.0435, -0.0234,  ..., -0.0294,  0.0168,  0.0226],
        [ 0.0129, -0.0044,  0.0471,  ...,  0.0024,  0.0066, -0.0032],
        [ 0.0057, -0.0280, -0.0093,  ...,  0.0039,  0.0347,  0.0199]],
       dtype=torch.bfloat16, requires_grad=True)
>>> ft_model.lm_head.weight.untyped_storage().data_ptr() == ft_model.model.embed_tokens.weight.untyped_storage().data_ptr()True
>>> ft_model.lm_head.weight.untyped_storage().data_ptr() == base_model.lm_head.weight.untyped_storage().data_ptr()
False
>>> ft_model.lm_head.weight.untyped_storage().data_ptr() == base_model.model.embed_tokens.weight.untyped_storage().data_ptr()
False
```

build/accelerate_launch.py
```diff
@@ -117,18 +128,91 @@ def main():
...
+                    torch_dtype=bfloat16 if use_flash_attn else None,
+                )
+
+            model_arch = model.config.model_type
+            # check that it is a granite model with llama architecture with tied weights
+            # ie. lm_head is duplicate of embeddings
+
+            # a fine tuned model will have params_dict.get("model.embed_tokens.weight")
+            # a prompt adapter has params_dict.get("base_model.model.embed_tokens.weight")
+            # a lora adapter has params_dict.get("base_model.model.model.embed_tokens.weight")
```
anhuong, line 174: I had assumed that `model.model.model.embed_tokens.weight.untyped_storage().data_ptr()` was the same as referencing 
```py
params_dict = dict(model.named_parameters())
params_dict.get("base_model.model.model.embed_tokens.weight")
``` 

Let me know if this assumption is incorrect. I moved away from params_dict in order to not have to differentiate between base_model and model

build/accelerate_launch.py
```diff
@@ -43,10 +47,17 @@
     USER_ERROR_EXIT_CODE,
     INTERNAL_ERROR_EXIT_CODE,
 )
+from tuning.data import tokenizer_data_utils
 
 ERROR_LOG = "/dev/termination-log"
 
 
+def get_base_model_from_adapter_config(adapter_config):
+    with open(adapter_config, "r", encoding="utf-8") as config_file:
```
alex-jw-brooks, line 56: Can you add a docstring to this function, just explaining that the adapter config is for peft models since this might be confusing to people that have only used transformers models?

build/accelerate_launch.py
```diff
@@ -117,18 +128,91 @@ def main():
             sys.exit(INTERNAL_ERROR_EXIT_CODE)
 
         try:
+            last_checkpoint_dir = get_highest_checkpoint(tempdir)
+            last_checkpoint_path = os.path.join(tempdir, last_checkpoint_dir)
+
+            use_flash_attn = job_config.get("use_flash_attn", True)
+            adapter_config_path = f"{last_checkpoint_path}/adapter_config.json"
```
alex-jw-brooks, line 135: It would be better to use `os.path.join` here since we use it in most places and it's easier to port to different platforms :)
```suggestion
            adapter_config_path = os.path.join(last_checkpoint_path, "adapter_config.json")
```

build/accelerate_launch.py
```diff
@@ -117,18 +128,91 @@ def main():
...
+                    base_model_path,
+                    attn_implementation="flash_attention_2" if use_flash_attn else None,
+                    torch_dtype=bfloat16 if use_flash_attn else None,
+                )
+
+                # since the peft library (PEFTModelForCausalLM) does not handle cases
+                # where the model's layers are modified, in our case the embedding layer
+                # is modified, so we resize the backbone model's embedding layer with our own
+                # utility before passing it along to load the PEFT model.
+                tokenizer_data_utils.tokenizer_and_embedding_resize(
```
alex-jw-brooks, line 152: is there a reason that this needs to happen again? I think this is called by the launched sft trainer command, so shouldn't the resizing already be handled?

anhuong: Unit tests for LoRA and PT were failing when trying to load the checkpoint with error:
```
RuntimeError: Error(s) in loading state_dict for LlamaForCausalLM:
	size mismatch for model.embed_tokens.weight: copying a param with shape torch.Size([32008, 64]) from checkpoint, the shape in current model is torch.Size([32000, 64]).
```
so then I had to resize the tokens to be able to load the checkpoint.

alex-jw-brooks: Iiiiinteresting. I guess the resize information must not be preserved in the config and reloading with the base class resets it to the base size or something 🤔

anhuong: I also verified that I get this same error when manually running tuning on tiny llama model with same params as unit test and then trying to load the checkpoint.

Ssukriti: huh? is this happening on current main due to some other changes in dependencies or something? 

why is this change related to the lm_head deletion here?

Ssukriti: if something changed, we should not just patch it by doing such things for final checkpoint only. Though product uses final checkpoint only for now , rest of research can use any checkpoint . so our load and inference scripts should work with any checkpoint

anhuong: I didn't think this was a patch, I thought this was always needed to load a model we tuned such as what we run in [run_inference](https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/scripts/run_inference.py#L193-L198). Since we update the tokenizer in our sft_trainer then when we load the tuned checkpoint again we need to update the tokenizer as well

anhuong: I didn't realize this was a new change added in a few weeks ago - https://github.com/foundation-model-stack/fms-hf-tuning/pull/227/files -- that resizes the embeddings to be a multiple of 8 with new tokens instead of the size of the tokenizer which required this addition....hmmm

alex-jw-brooks (approved):

PR #259, opened 2024-07-23
refactor code to preprocess datasets

Ssukriti:
### Description of the change
Adds utilities to format datasets to single sequence. Retains current functionality in main. Moves around code for modularity and expansion of new dataset formats.

Utility functions:
1. get_data_collator - to return data collator as needed
2. format_dataset - > to format dataset to single sequence using template

### How to verify the PR
Unit tests passed

### Was the PR tested


tuning/utils/data_utils.py
```diff
@@ -2,21 +2,25 @@
...
             Defaults to empty
     Returns:
-        Formatted HF Dataset, dataset_field name that contains formatted data.
+        Formatted HF Dataset
     """
 
-    formatted_dataset_field = "formatted_data_field"
     template += eos_token
 
+    assert formatted_dataset_field is not None
```
alex-jw-brooks, line 22: Can this assert be removed or changed to a `ValueError("Unable to apply custom formatting because the formatted_dataset_field was not provided")` or something?

tests/utils/test_preprocessing_utils.py
```diff
@@ -207,3 +209,57 @@ def test_get_trainer_kwargs_with_custom_masking(use_validation_data):
...
+                training_data_path=TWITTER_COMPLAINTS_DATA_INPUT_OUTPUT,
+                validation_data_path=TWITTER_COMPLAINTS_DATA_INPUT_OUTPUT,
+                dataset_text_field="formatted_field",
+                data_formatter_template="### Text:{{input}} \n\n### Label: {{output}}",
+            )
+        ),
+    ],
+)
+def test_format_dataset(data_args):
+    tokenizer = AutoTokenizer.from_pretrained("Maykeye/TinyLLama-v0")
```
alex-jw-brooks, line 260: Can we maybe make the model that we use in these tests / set trainer tests pull from the same place, similar to the `TWITTER_COMPLAINTS_DATA`, so that if we need to change it in the future we don't miss this one?

alex-jw-brooks (changes requested):

tests/utils/test_preprocessing_utils.py
Ssukriti: yes good catch :)

tuning/utils/data_utils.py
Ssukriti: done :)

alex-jw-brooks (approved):

PR #260, opened 2024-07-23
Data custom collator

Ssukriti:
### Description of the change
Addition of support for JSON datasets containing input/output fields . Masking on input to leave output unmasked. Create single sequence by concatenation.
Avoids datacollatorforcompletionLM that requires response template. This is because response template is very error prone and subject to it being present in text. 

Refactored https://github.com/foundation-model-stack/fms-hf-tuning/pull/166/files a bit to allow extension to support pretokenized datasets which are masked.

### How to verify the PR

1. Train using JSON input/output fields and leave dataset_text_field and response_template blank

### Was the PR tested

1. Quality test below
2. Unit tests 


tests/test_sft_trainer.py
```diff
@@ -724,3 +725,58 @@ def test_run_with_good_experimental_metadata():
...
+    with tempfile.TemporaryDirectory() as tempdir:
+        train_args = copy.deepcopy(TRAIN_ARGS)
+        train_args.output_dir = tempdir
+
+        data_args = copy.deepcopy(DATA_ARGS)
+        data_args.dataset_text_field = dataset_text_field
+        data_args.response_template = response_template
+        data_args.training_data_path = TWITTER_COMPLAINTS_DATA_INPUT_OUTPUT
+        # We should raise an error since we should not have a dataset text
+        # field or a response template if we're pretokenized data
```
alex-jw-brooks, line 762:
```suggestion
        # field or a response template if we have pretokenized data
```

tests/test_sft_trainer.py
```diff
@@ -724,3 +725,58 @@ def test_run_with_good_experimental_metadata():
...
+        train_args = copy.deepcopy(TRAIN_ARGS)
+        train_args.output_dir = tempdir
+
+        data_args = copy.deepcopy(DATA_ARGS)
+        data_args.dataset_text_field = None
+        data_args.response_template = None
+        data_args.training_data_path = TWITTER_COMPLAINTS_DATA
+
+        # It would be best to handle this in a way that is more understandable; we might
+        # need to add directly validation prior to the dataset generation since datasets
```
alex-jw-brooks, line 779:
```suggestion
        # need to add validation prior to the dataset generation since datasets
```

tuning/utils/preprocessing_utils.py
```diff
@@ -65,16 +59,32 @@ def validate_data_args(data_args: configs.DataArguments, packing: bool):
...
-    # TODO(s) In future seupport two more formats:
-    # 1. Allow no response template, and JSON with input/output fields and mask input
 
-    # 2. Allow pretokenized Dataset besides JSON.
+    # If not single sequence, JSON should contain input/output fields
+    if not (data_args.dataset_text_field or data_args.data_formatter_template):
+        json_dataset = datasets.load_dataset(
+            "json", data_files=data_args.training_data_path
+        )
+        if "input" not in json_dataset["train"].column_names:
```
alex-jw-brooks, line 68: Can you add two const vars 
```python
INPUT_KEY = "input"
OUTPUT_KEY = "output"
```
at the top of this file and check against the variables instead of hardcoding the value here?

Then it can also be used in the calls to `get_preprocessed_dataset` and the error messages below and only be hardcoded in one place

tests/utils/test_preprocessing_utils.py
```diff
@@ -108,80 +108,52 @@ def test_get_preprocessed_dataset(max_sequence_length):
...
+            ),
+            1024,
+            DataCollatorForSeq2Seq,
+        ),
     ],
 )
-def test_get_trainer_kwargs_with_response_template_and_text_field(
-    use_validation_data, collator_type, packing
+def test_get_data_collator(
+    packing,
```
alex-jw-brooks, line 143: Can you add a docstring to this test, e.g.,
```Ensure that the correct collator type is fetched based on the data args```
or something? I think it's obvious when you actually look at the test, but the parameterized stuff might be a little overwhelming for people unfamiliar with the collator logic

File: tests/utils/test_preprocessing_utils.py [hunk 5]
```diff
@@ -255,12 +242,26 @@ def test_get_formatted_dataset_with_single_sequence(
...
+                training_data_path=TWITTER_COMPLAINTS_DATA_INPUT_OUTPUT,
+                validation_data_path=TWITTER_COMPLAINTS_DATA_INPUT_OUTPUT,
+            )
+        ),
     ],
 )
 def test_format_dataset(data_args):
-    tokenizer = AutoTokenizer.from_pretrained(MODEL_NAME)
-    train_set, eval_set, dataset_text_field = format_dataset(data_args, tokenizer)
+    tokenizer = AutoTokenizer.from_pretrained("Maykeye/TinyLLama-v0")
```
alex-jw-brooks, line 255: Would also be nice to add a docstring like ```Ensure that the train/eval data are properly formatted based on the data args / text field```

alex-jw-brooks (changes requested):

Ssukriti:
find some quality results here https://github.ibm.com/ai-foundation/watson-fm-stack-tracker/issues/621#issuecomment-86629029 using both collators

Ssukriti:
@alex-jw-brooks review requested is addressed, also added EOS token 

Attached some quality results above.

**Remaining TODOs:**
1. Update README 
2. I think JSON file format is broken (JSONL works ) - to check
3. Run more quality tests if have time to have further record

Ssukriti:
The PR is ready to be reviewed @alex-jw-brooks @anhuong . I have documented only JSONL format for new input/output and will extend it to JSON in following PR . 

Also current functionality is retained and we are just adding support for a new data format . I think it is safe to merge as one quality test looked good. we can continue to do quality testing before announcing release

alex-jw-brooks (approved):

fabianlim:
@Ssukriti i think this PR is not accurately named, the changes should be to allow for custom processing by means of passing in the Jinja templates. Collation is usually referred to the step where the examples come out of the sampler, and they are formed into a single tensor to be passed out of the dataloader.

PR #265, opened 2024-07-25
docs: fix the instructions for running with LORA

HarikrishnanBalagopal:
### Description of the change

The command in the README is wrong.

### How to verify the PR

Run the command for LORA and see the output log messages.

### Was the PR tested


kmehant (approved):
+1 on the fix.

example is fixed.

PR #270, opened 2024-07-30
Fix: Removal of transformers logger and addition of python native logger

Abhishek-TAMU:
### Description of the change

Aim was to move away from using logger modules from transformers and use python native [logger](https://docs.python.org/3/library/logging.html). Now, log level is either set by CLI argument of `--log_level` OR environment variable `LOG_LEVEL`. The setting of log level applies to both python native logger and log level of transformers logging.

### Related issue number

#[994](https://github.ibm.com/ai-foundation/watson-fm-stack-tracker/issues/994)

### How to verify the PR

- Set environment variable `LOG_LEVEL` as desired log level and check if logs are correct according to passed LOG_LEVEL, by running tuning job for a small model like `Maykeye/TinyLLama-v0`. 

Command:
```
--model_name_or_path Maykeye/TinyLLama-v0  \
--training_data_path tests/data/twitter_complaints_small.json  \
--output_dir outputs/full-tuning  \
--num_train_epochs 5  \
--per_device_train_batch_size 2  \
--gradient_accumulation_steps 1  \
--learning_rate 1e-5  \
--use_flash_attn false \
--torch_dtype "float32"
```

- Pass `log_level` via cli argument, and check if logs are correct according to passed LOG_LEVEL, by running tuning job for a small model like `Maykeye/TinyLLama-v0`.

Command:
```
--model_name_or_path Maykeye/TinyLLama-v0  \
--training_data_path tests/data/twitter_complaints_small.json  \
--output_dir outputs/full-tuning  \
--num_train_epochs 5  \
--per_device_train_batch_size 2  \
--gradient_accumulation_steps 1  \
--learning_rate 1e-5  \
--use_flash_attn false \
--torch_dtype "float32" \
--log_level "info"
```

### Was the PR tested


tuning/config/configs.py
```diff
@@ -136,6 +136,13 @@ class TrainingArguments(transformers.TrainingArguments):
             + "Requires additional configs, see tuning.configs/tracker_configs.py"
         },
     )
+    log_level: str = field(
+        default="passive",
+        metadata={
+            "help": "The log level to adopt during training. \
+            Possible values are 'debug', 'info', 'warning', 'error' and 'critical'"
```
anhuong, line 143: Add note on what "passive" value means
> plus a ‘passive’ level which doesn’t set anything and keeps the current log level for the Transformers library (which will be "warning" by default)

tuning/sft_trainer.py
```diff
@@ -461,12 +461,46 @@ def parse_arguments(parser, json_config=None):
...
+
+    # Configure Python native logger log level
+    # If CLI arg is passed, assign same log level to python native logger
+    if parsed_training_args.log_level != "passive":
+        logging.basicConfig(level=parsed_training_args.log_level.upper())
+    else:
+        # Assign value of either env var LOG_LEVEL or warning
+        LOGLEVEL = os.environ.get("LOG_LEVEL", "WARNING").upper()
+        logging.basicConfig(level=LOGLEVEL)
+        train_logger = logging.getLogger()
```
anhuong, line 484: Since you call this on line 497, you don't need to call it within this block as well

tuning/sft_trainer.py
```diff
@@ -481,7 +515,11 @@ def main(**kwargs):  # pylint: disable=unused-argument
             fusedops_kernels_config,
             exp_metadata,
         ) = parse_arguments(parser, job_config)
-        logger.debug(
+
+        # Function to set log level for python native logger and transformers training logger
+        training_args, _ = set_log_level(training_args)
```
anhuong, line 520: Why are you not using the `logging.getLogger()` that you returned? Since you are using the root logger, running
```py
logging.warning("Warning")
# above is the same as below
root_logger = logging.getLogger()
root_logger.warning("Warning")
```
Running the above is the same. So if you are not using the getLogger, you shouldn't return it/use it.


But it looks like it is best practice to use getLogger from [python docs](https://docs.python.org/3/library/logging.html#logging.getLogger) and [python logging howto](https://docs.python.org/3/howto/logging.html).
So then instead of running `logging.warning()` you'd run `logger.warning()`. In which in the train() you would also call `logging.getLogger()`. It looks like its best practice to create module level loggers. So you could do `logging.getLogger("sft_trainer_main")` and logging.getLogger("sft_trainer_train")` and then the log messages should show like:

```sh
# named logger if you do getLogger("sft_trainer_train")
WARNING:sft_trainer_train:My warning

# root logger if you do getLogger()
WARNING:root:My warning
```

tuning/trackers/filelogging_tracker.py
```diff
@@ -80,7 +80,10 @@ def __init__(self, tracker_config: FileLoggingTrackerConfig):
                 which contains the location of file where logs are recorded.
         """
         super().__init__(name="file_logger", tracker_config=tracker_config)
-        self.logger = logging.get_logger("file_logging_tracker")
+        # Configure log level
+        LOGLEVEL = os.environ.get("LOG_LEVEL", "WARNING").upper()
+        logging.basicConfig(level=LOGLEVEL)
```
anhuong, line 85: `logging.basicConfig()` sets the log level for the root logger. But you can set individual loggers level by running `self.logger.setLevel(level)`

anhuong: There are a few of these to fix, not going to comment in each file

tuning/sft_trainer.py
```diff
@@ -461,12 +461,46 @@ def parse_arguments(parser, json_config=None):
...
+        logging.basicConfig(level=parsed_training_args.log_level.upper())
+    else:
+        # Assign value of either env var LOG_LEVEL or warning
+        LOGLEVEL = os.environ.get("LOG_LEVEL", "WARNING").upper()
+        logging.basicConfig(level=LOGLEVEL)
+        train_logger = logging.getLogger()
+
+    # Check if env var TRANSFORMERS_VERBOSITY is not set.
+    # Else if env var is already set then, log level of transformers is automatically set.
+    if os.environ.get("TRANSFORMERS_VERBOSITY") is None:
```
anhuong, line 488: nit: can do 
```suggestion
   if not os.environ.get("TRANSFORMERS_VERBOSITY"):
```

anhuong: Although do we even need the code block below? What use case is it covering? Do we have to check that TRANFORMERS_VERBOSITY at all?

is this trying to do
```py
if os.getenv("TRANFORMERS_VERBOSITY"):
   parsed_training_args.log_level = os.getenv("TRANFORMERS_VERBOSITY").lower()
```

tuning/sft_trainer.py
anhuong, line 484: Could this work to attain the logic we're looking for?
Goal for logic:
- TRANSFORMERS_VERBOSITY ---> should only set transformers logging, this is different than setting --log_level
- `--log_level` --> sets both python and transformers logging
- For consistency with the flag, LOG_LEVEL also sets python and transformers logging

```py
    if parsed_training_args.log_level != "passive":
        logging.basicConfig(level=parsed_training_args.log_level.upper())
    elif os.environ.get("LOG_LEVEL"):
        # Assign value of either env var LOG_LEVEL or warning
        LOGLEVEL = os.environ.get("LOG_LEVEL")
        logging.basicConfig(level=LOGLEVEL)
       # Set log_level in TrainingArguments
       parsed_training_args.log_level = LOGLEVEL.lower()

  train_logger = logging.getLogger()
```

anhuong (changes requested):
Thanks Abhishek for the logging update, I think this will help a lot. A few questions on how the logging works and the desired logic

tuning/trackers/filelogging_tracker.py
aluu317, line 85: It'd be helpful also if in each file, you could do:
```
logger = logging.getLogger(__name__)
```
This will specify in the logs which module the log came from, example:
```
INFO:__main__:Started
INFO:mylib:Doing something
INFO:__main__:Finished
```

aluu317: I don't think you even have to set log level for every file, you only have to set it once in the sft_trainer.py after parsing user's input right?

In each file, you just need to get a logger by doing:
```
import logging
logger = logging.getLogger(__name__)
```
and every time we log something, we just use that `logger`:

```
logger.info("Log this info")
```

aluu317 (changes requested):

tuning/sft_trainer.py
Abhishek-TAMU: In returned logger on line number 497, just to get logger in unit test cases functions. returning logger won't make sense in same file as we can set log level in same py file with returning logger.

But ya that's a good idea to use logging.getLogger. I will try and test it

tuning/sft_trainer.py
Abhishek-TAMU: This is done!

build/accelerate_launch.py
```diff
@@ -80,6 +77,14 @@ def main():
             or 'SFT_TRAINER_CONFIG_JSON_ENV_VAR'."
             )
 
+        # Configure log_level of python native logger.
+        LOGLEVEL = None
```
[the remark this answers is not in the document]
Abhishek-TAMU: Thank you for the PR review. This is done!

build/accelerate_launch.py
```diff
@@ -80,6 +77,14 @@ def main():
...
+        if "log_level" in job_config and job_config["log_level"]:
+            LOGLEVEL = job_config["log_level"].upper()
+            logging.basicConfig(level=LOGLEVEL)
+        else:
+            LOGLEVEL = os.environ.get("LOG_LEVEL", "WARNING").upper()
+            logging.basicConfig(level=LOGLEVEL)
```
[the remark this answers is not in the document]
Abhishek-TAMU: This is done!

tuning/config/configs.py
Abhishek-TAMU: Added note in metadata!

tuning/sft_trainer.py
Abhishek-TAMU: I don't think we are required to check "TRANFORMERS_VERBOSITY". Hence I haven't added this code in final commit.

tuning/sft_trainer.py
```diff
@@ -461,12 +461,46 @@ def parse_arguments(parser, json_config=None):
...
+    for handler in logging.root.handlers[:]:
+        logging.root.removeHandler(handler)
+
+    # Configure Python native logger log level
+    # If CLI arg is passed, assign same log level to python native logger
+    if parsed_training_args.log_level != "passive":
+        logging.basicConfig(level=parsed_training_args.log_level.upper())
+    else:
+        # Assign value of either env var LOG_LEVEL or warning
+        LOGLEVEL = os.environ.get("LOG_LEVEL", "WARNING").upper()
```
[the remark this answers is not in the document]
Abhishek-TAMU: Here if LOG_LEVEL is not set, it will default to logging.WARNING, which is an integer constant from the logging module (default is 30). The use of .upper() here will cause an error because .upper() is a string method and cannot be called on an integer.

tuning/trackers/filelogging_tracker.py
Abhishek-TAMU: Yea that's a good idea. I will do that.

tuning/sft_trainer.py
Abhishek-TAMU: I didn't add this line `parsed_training_args.log_level = LOGLEVEL.lower()` in the same if-else because if env variable 'TRANSFORMERS_VERBOSITY' is set, then this would get overwritten. Hence wrote this line in separate log below checking `if not os.environ.get("TRANSFORMERS_VERBOSITY"):`

But yea the code that you write here is optimized. Thank you!

Abhishek-TAMU: And yes it does achieve the logic we aim for. Tested locally and on image too!

tuning/trackers/filelogging_tracker.py
Abhishek-TAMU: @aluu317 Yes I think it makes sense.
I made changes and used `logger = logging.getLogger("sft_trainer_main")` everywhere.

anhuong: I agree that having each file set their own logging level helps to ensure that the logging level is set if a file is called outside of sft_trainer.py. I do agree that using `logging.getLogger(__name__)` is best practice and it is unlikely these files will be called outside of sft_trainer.py so I think going with Angel's suggestion works

tuning/sft_trainer.py
anhuong: Let's chat about this logic to clarify

build/accelerate_launch.py
```diff
@@ -80,6 +77,21 @@ def main():
...
+        if (
+            not log_level
+        ):  # if log level not set by job_config aka by JSON, set it via env var or set default
+            log_level = os.environ.get("LOG_LEVEL", "WARNING")
+        log_level = log_level.upper()
+        logging.basicConfig(level=log_level)
+
+        # Configure for Image to get log level as the code after
+        # launch_command only takes log level from env var LOG_LEVEL and not CLI
+        os.environ["LOG_LEVEL"] = log_level
```
anhuong, line 93: technically you shouldn't set the env LOG_LEVEL if the user already set it. so something more like:
```py
log_level = job_config.get("log_level")
if not log_level:
   log_level = os.environ.get("LOG_LEVEL", "WARNING")
else:
   os.environ["LOG_LEVEL"] = log_level.upper()
logging.basicConfig(level=log_level.upper())
```

tests/test_sft_trainer.py
```diff
@@ -808,3 +809,62 @@ def test_pretokenized_dataset_wrong_format():
...
+    """
+    Ensure that the correct log level is being set
+    for python native logger and transformers logger
+    """
+
+    # Set env var TRANSFORMERS_VERBOSITY as None and test
+    os.unsetenv("TRANSFORMERS_VERBOSITY")
+    os.unsetenv("LOG_LEVEL")
+    os.unsetenv("SFT_TRAINER_CONFIG_JSON_ENV_VAR")
+    os.unsetenv("SFT_TRAINER_CONFIG_JSON_PATH")
```
anhuong, line 824: Another way you can do this is by mocking the env:
```py
# decorator can be used on whole function
@mock.patch.dict(os.environ, {}, clear=True)

# or can create context scope
with mock.patch.dict(os.environ, {}, clear=True):
```

tests/test_sft_trainer.py
```diff
@@ -808,3 +809,62 @@ def test_pretokenized_dataset_wrong_format():
...
+    train_args = copy.deepcopy(TRAIN_ARGS)
+
+    # TEST IF NO ENV VAR ARE SET AND NO CLI ARGUMENT IS PASSED
+    training_args, logger = sft_trainer.set_log_level(train_args)
+    assert logger.getEffectiveLevel() == logging.WARNING
+    assert training_args.log_level in ["passive", "warning"]
```
anhuong, line 830: Since no CLI flag is passed, `training_args.log_level == "passive"` correct? Since this is the default value set

tests/test_sft_trainer.py
```diff
@@ -808,3 +809,62 @@ def test_pretokenized_dataset_wrong_format():
...
+    """
+
+    # Set env var TRANSFORMERS_VERBOSITY as None and test
+    os.unsetenv("TRANSFORMERS_VERBOSITY")
+    os.unsetenv("LOG_LEVEL")
+    os.unsetenv("SFT_TRAINER_CONFIG_JSON_ENV_VAR")
+    os.unsetenv("SFT_TRAINER_CONFIG_JSON_PATH")
+    train_args = copy.deepcopy(TRAIN_ARGS)
+
+    # TEST IF NO ENV VAR ARE SET AND NO CLI ARGUMENT IS PASSED
```
anhuong, line 827: don't need this comment since you should have described it both in the test name and in the test docstring. thus you should include a little more in the docstring to say "Ensure that the correct log level is being set for python native logger and transformers logger when no env var or CLI flag is passed"

tests/test_sft_trainer.py
```diff
@@ -808,3 +809,62 @@ def test_pretokenized_dataset_wrong_format():
...
+    training_args, logger = sft_trainer.set_log_level(train_args)
+    assert logger.getEffectiveLevel() == logging.WARNING
+    assert training_args.log_level in ["passive", "warning"]
+
+
+def test_set_log_level_for_logger_with_env_var():
+    """
+    Ensure that the correct log level is being set
+    for python native logger and transformers logger
```
anhuong, line 836: These should be more descriptive per test case and explain how the tests differentiate from each other. For example, this one should say "Ensure that the correct log level is being set for python native logger and transformers logger when LOG_LEVEL env var used"

tests/test_sft_trainer.py
```diff
@@ -808,3 +809,62 @@ def test_pretokenized_dataset_wrong_format():
...
+    """
+
+    os.unsetenv("TRANSFORMERS_VERBOSITY")
+    os.unsetenv("LOG_LEVEL")
+    os.unsetenv("SFT_TRAINER_CONFIG_JSON_ENV_VAR")
+    os.unsetenv("SFT_TRAINER_CONFIG_JSON_PATH")
+    train_args_env = copy.deepcopy(TRAIN_ARGS)
+
+    # TEST IF LOG_LEVEL ENV VAR IS SET AND NO CLI ARGUMENT IS PASSED
+    os.environ["LOG_LEVEL"] = "info"
```
anhuong, line 846: Similarly, an easier way of doing this would be to set:
```py
# decorator can be used on whole function
@mock.patch.dict(os.environ, {"LOG_LEVEL": "info"}, clear=True)

# or can create context scope
with mock.patch.dict(os.environ, {"LOG_LEVEL": "info"}, clear=True):
```

tests/test_sft_trainer.py
```diff
@@ -808,3 +809,62 @@ def test_pretokenized_dataset_wrong_format():
...
+    train_args_env.log_level = "passive"  # Default
```
anhuong, line 847: If this is the default value, then you don't need to set it right? You could do an assertion to ensure this default value is set though

tests/test_sft_trainer.py
```diff
@@ -808,3 +809,62 @@ def test_pretokenized_dataset_wrong_format():
...
+def test_set_log_level_for_logger_with_env_var_and_cli():
+    """
+    Ensure that the correct log level is being set
+    for python native logger and transformers logger
+    """
+
+    os.unsetenv("TRANSFORMERS_VERBOSITY")
+    os.unsetenv("LOG_LEVEL")
+    os.unsetenv("SFT_TRAINER_CONFIG_JSON_ENV_VAR")
+    os.unsetenv("SFT_TRAINER_CONFIG_JSON_PATH")
```
anhuong, line 862: update per previous comments

tests/test_sft_trainer.py
```diff
@@ -808,3 +809,62 @@ def test_pretokenized_dataset_wrong_format():
...
+
+    # TEST IF LOG_LEVEL ENV VAR IS SET AND NO CLI ARGUMENT IS PASSED
+    os.environ["LOG_LEVEL"] = "info"
+    train_args_env.log_level = "passive"  # Default
+    training_args, logger = sft_trainer.set_log_level(train_args_env)
+    assert logger.getEffectiveLevel() == logging.INFO
+    assert training_args.log_level == "info"
+
+
+def test_set_log_level_for_logger_with_env_var_and_cli():
```
anhuong, line 853: Please add another test case where:
- setting flag sets the python log level
- separate one for setting TRANSFORMERS_VERBOSITY doesn't affect the flags

tuning/config/configs.py
```diff
@@ -136,6 +136,15 @@ class TrainingArguments(transformers.TrainingArguments):
             + "Requires additional configs, see tuning.configs/tracker_configs.py"
         },
     )
+    log_level: str = field(
+        default="passive",
+        metadata={
+            "help": "The log level to adopt during training. \
+            'passive' level which doesn't set anything and keeps the \
```
anhuong, line 143:
```suggestion
            By default, 'passive' level is set which keeps the \
```

tuning/trackers/aimstack_tracker.py
```diff
@@ -97,7 +97,8 @@ def __init__(self, tracker_config: AimConfig):
             information about the repo or the server and port where aim db is present.
         """
         super().__init__(name="aim", tracker_config=tracker_config)
-        self.logger = logging.get_logger("aimstack_tracker")
+        # Configure log level
+        self.logger = logging.getLogger(__name__)
```
anhuong, line 101: Please provide an example of running tuning and the log messages printed per module. THanks!

anhuong (changes requested):
Logic looks great! Thanks for the updates Abhishek! A few changes on the unit tests and some details on the testing would be helpful. Otherwise, looks good

build/accelerate_launch.py
Abhishek-TAMU: Actually, I wrote the same code what you mentioned here, but not sure why, a weird thing was happening last week when running this log level logic in Image:

Whenever `sfttrainer.py` is run with accelerate, with CLI argument `--log_level` it doesn't take in this cli argument when executing `launch_command(args)` and hence  `log_level` CLI argument doesn't reach `sfttrainer.py`. But if I set log level by env variable, it consider that and runs fine as per logic. 

Hence as a workaround I check CLI value in `accelerate_launch.py` along with env variable and accordingly as per logic assigns the log_level to "LOG_LEVEL" env var so that `launch_command(args)` takes this env var and runs with correct log level.

tests/test_sft_trainer.py
Abhishek-TAMU: Thank you!

tests/test_sft_trainer.py
Abhishek-TAMU: Yea, that should be the correct one. But, `train_args = copy.deepcopy(TRAIN_ARGS)`, `train_args` takes in default value of log_level as "warning", which is quite strange and couldn't figure out why.

tests/test_sft_trainer.py
Abhishek-TAMU: Done. Thanks!

tests/test_sft_trainer.py
Abhishek-TAMU: Done!

build/accelerate_launch.py
anhuong: > Whenever sfttrainer.py is run with accelerate, with CLI argument --log_level it doesn't take in this cli argument when executing launch_command(args) 

The accelerate launch script only parses JSON to and not the flags. So passing `--log_level` would never work for accelerate launch but passing it in via JSON `"log_level": "info"` should. Was this also not working?

> log_level CLI argument doesn't reach sfttrainer.py

Please describe, if you pass in log_level via JSON it should be interpreted both by accelerate_launch and sft_trainer scripts.

And if the user sets `export LOG_LEVEL=info` this should also be interpreted both by accelerate_launch and sft_trainer scripts. 

The logic I have above is not too different frmo yours, it just doesn't reset the env var LOG_LEVEL if the user already sets this value.

Abhishek-TAMU: > The accelerate launch script only parses JSON to and not the flags. So passing --log_level would never work for accelerate launch but passing it in via JSON "log_level": "info" should. Was this also not working?

Yes, it was not taking the config json value of `--log_level`. I will share the example with you shortly.

tests/test_sft_trainer.py
anhuong: That is strange, doesn't look like its coming from transformers side and we set log_level to passive as well...
```py
from transformers import TrainingArguments
training_args = TrainingArguments(learning_rate=2e-5,per_device_train_batch_size=4, num_train_epochs=2, save_strategy="epoch", output_dir="output_dir")
print(training_args.log_level)
# 'passive'

```

Abhishek-TAMU: Yes only in unit test case, it basically takes default value as "warning". Attaching, the test fail logs. I will dig more into it.

image

Abhishek-TAMU: With a thought in mind that above test cases must be tampering with TRAIN_ARGS and hence moved all logging unit tests at the start in the same file. And now this issue is fixed. Quite strange!!

tests/test_sft_trainer.py
Abhishek-TAMU: Yea, related to the issue I was facing, which I mentioned above related to default value of TRAIN_ARGS as "warning". Now as the issue is fixed, I removed `train_args_env.log_level = "passive"`

tests/test_sft_trainer.py
Abhishek-TAMU: I am not sure if I understand the case completely on what we want to test here. Hence to save iteration, sharing the test case here, for your input and comments, Thanks! 

```
@mock.patch.dict(os.environ, {}, clear=True)
def test_set_log_level_for_logger_with_set_verbosity_and_cli():
    """
    Ensure that the correct log level is being set for python native logger and 
    log_level of transformers logger is unchanged when env var TRANSFORMERS_VERBOSITY is used 
    and CLI flag is passed
    """

    train_args = copy.deepcopy(TRAIN_ARGS)
    os.environ["TRANSFORMERS_VERBOSITY"] = "info"
    train_args.log_level = "error"
    training_args, logger = sft_trainer.set_log_level(train_args)
    assert logger.getEffectiveLevel() == logging.ERROR
    assert training_args.log_level == "error"
```

tuning/trackers/aimstack_tracker.py
Abhishek-TAMU: Do you mean all the modules where we changed logging right, I should log statements and check if changes in every file works...right ?

build/accelerate_launch.py
Abhishek-TAMU: Okay, I check again and it's working. Not sure why it was not taking in leg level from json file last week. 
Thanks though!

tuning/sft_trainer.py
```diff
@@ -477,12 +479,43 @@ def parse_arguments(parser, json_config=None):
     )
 
 
-def main(**kwargs):  # pylint: disable=unused-argument
-    logger = logging.get_logger("__main__")
+def set_log_level(parsed_training_args, logger_name=__name__):
```
dushyantbehl, line 482: Just a suggestion: imo it will look better if this were to be moved to a separate file `utils/logging.py`

dushyantbehl: Also since `parsed_training_args` looks similar to `train_args` it might be simple to rename this argument.

Abhishek-TAMU: Sure!

tests/test_sft_trainer.py
anhuong: Yes I think it's because you are editing the training_args env var when you run set_log_level() so the tests are conflicting with each other. Because if I parse the flags standalone, the default is correctly set to "passive"
```py
from tuning.config import configs
TRAIN_ARGS = configs.TrainingArguments(
    num_train_epochs=5,
    per_device_train_batch_size=4,
    per_device_eval_batch_size=4,
    gradient_accumulation_steps=4,
    learning_rate=0.00001,
    weight_decay=0,
    warmup_ratio=0.03,
    lr_scheduler_type="cosine",
    logging_steps=1,
    include_tokens_per_second=True,
    packing=False,
    max_seq_length=4096,
    save_strategy="epoch",
    output_dir="tmp",
)
print(TRAIN_ARGS.log_level)
# 'passive'
```

anhuong: But if you are using a new TRAIN_ARGS....then they shouldn't conflict with each other. But perhaps the env vars were being leftover set. When I run the unit test with:
```
with mock.patch.dict(os.environ, {}, clear=True):
        print(TRAIN_ARGS.log_level)
        training_args, logger = sft_trainer.set_log_level(TRAIN_ARGS)
        assert logger.getEffectiveLevel() == logging.WARNING
        assert training_args.log_level == "passive"
```
it passes

Abhishek-TAMU: Got it!

tuning/trackers/aimstack_tracker.py
Abhishek-TAMU: @anhuong 
Sharing the example of usage of root log level in different file other than `sfttrainer.py`.
In `data_type_utils.py`, I did `import logging` and did not set logger. For testing purpose wrote warning and error logs in one of the function `get_torch_dtype`

In `sfttrainer.py` main(), I set root log level (Line 502) and called the function `get_torch_dtype`. Required behaviour should be that the error message gets logged and it's happening.

Hence, removing the line `logger = logging.getLogger(__name__)` from all the files where this is globally defined, and just using `logging.error/warning/info` in those files would automatically works according to the root log-level, set in `main()` function of `sfttrainer.py`

image

Abhishek-TAMU: @anhuong New changes in PR are pushed. Feel free to review it whenever possible.

tuning/trainercontroller/controllermetrics/history_based_metrics.py
```diff
@@ -21,12 +21,11 @@
 
 # Third Party
 from transformers import TrainerState
-from transformers.utils import logging
 
 # Local
 from tuning.trainercontroller.controllermetrics.metricshandler import MetricHandler
 
-logger = logging.get_logger(__name__)
+# Configure log level
```
anhuong, line 28: remove comment, no longer relevant

anhuong (changes requested):

tuning/trackers/aimstack_tracker.py
Abhishek-TAMU: @dushyantbehl Wanted to mention something.

From the file (eval_metrices.py, tracker_factory.py etc), the global declaration of `logger = logging.getLogger(__name__)`, has been removed. This is because it runs before our code of parsing cli-argument and setting root log-level in `sfttrainer.py`. Now after removal of this, all the `logging` statements in the file works as per root log-level. This sounds good to you right.

tuning/trainercontroller/controllermetrics/history_based_metrics.py
Abhishek-TAMU: @anhuong  Thanks for the quick review. I have made all changes mentioned above and pushed it again. Thank you!
As discussed, I can also wait for Dushyant's comment!

tests/utils/test_logging.py
```diff
@@ -0,0 +1,98 @@
...
+from unittest import mock
+import copy
+import logging
+import os
+
+# Local
+from tuning.config import configs
+from tuning.utils.logging import set_log_level
+
+TRAIN_ARGS = configs.TrainingArguments(
```
dushyantbehl, line 28: Since this is same as the `TRAIN_ARGS` in sft_trainer, you can import this directly from the main test file instead of redeclaration, please see - 

https://github.com/foundation-model-stack/fms-hf-tuning/blob/d224be6a7d8aaf98872b5e14630750739db2f7a9/tests/trackers/test_aim_tracker.py#L29

tuning/trackers/aimstack_tracker.py
```diff
@@ -97,7 +97,8 @@ def __init__(self, tracker_config: AimConfig):
             information about the repo or the server and port where aim db is present.
         """
         super().__init__(name="aim", tracker_config=tracker_config)
-        self.logger = logging.get_logger("aimstack_tracker")
+        # Get logger with root log level
+        self.logger = logging.getLogger()
```
dushyantbehl, line 101: @Abhishek-TAMU any particular reason to remove module name from the `logger` I had kept it to segregate logs from various subcomponents.

dushyantbehl: Same comment for `file_logging_tracker`

tuning/utils/logging.py
```diff
@@ -0,0 +1,46 @@
+# Standard
```
dushyantbehl, line 1: Needs a code ownership declaration similar to this - 

https://github.com/foundation-model-stack/fms-hf-tuning/blob/d224be6a7d8aaf98872b5e14630750739db2f7a9/tuning/sft_trainer.py#L1

tuning/trackers/aimstack_tracker.py
dushyantbehl: Thanks @Abhishek-TAMU I missed this message please ignore my comments below.

tuning/trackers/aimstack_tracker.py
dushyantbehl: Explained in the comment above..no need for futher explanation.

tests/utils/test_logging.py
Abhishek-TAMU: Thank you for this input. This is done!

tuning/trackers/aimstack_tracker.py
Abhishek-TAMU: Cool. Thank you!

tuning/utils/logging.py
Abhishek-TAMU: This is done!

Abhishek-TAMU:
@anhuong Thank you again for all your review on this PR. I have pushed the final changes. Feel free to have look whenever possible. Thanks!

anhuong (dismissed):
Awesome work Abhishek! This will help a lot, many thanks!

tuning/trackers/aimstack_tracker.py
aluu317: I still dont quite understand this. You're setting log_level at runtime to restrict the amount of logs that a user wants to see (`debug` vs `info` etc.). But if you don't do `logging.get_logger("aimstack_tracker")` and so on in each file, all the logs will say they come from `root`, not module specific, helpful but not as helpful if the module is also included.
Maybe it'd be clearer if you could share an example run of what the logs look like if user sets `debug` or `error` for example and we can see what the logs look like.

aluu317 (changes requested):
@Abhishek-TAMU Some comments inline. Thank you!

Abhishek-TAMU:
@aluu317 The reason behind removing `logging.get_logger("aimstack_tracker")` was that it doesn't instantiate logger with log level equivalent to root log level and creates a new logger with default log level, whereas `logging.getLogger()` gets the logger which is equivalent to root log level.

build/accelerate_launch.py
```diff
@@ -80,6 +77,17 @@ def main():
             or 'SFT_TRAINER_CONFIG_JSON_ENV_VAR'."
             )
 
+        # Configure log_level of python native logger.
```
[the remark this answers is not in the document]
Abhishek-TAMU: Thank you! This is done

tests/utils/test_logging.py
```diff
@@ -0,0 +1,83 @@
...
+    training_args, logger = set_log_level(train_args)
+    assert logger.getEffectiveLevel() == logging.ERROR
+    assert training_args.log_level == "error"
+
+
+@mock.patch.dict(os.environ, {"LOG_LEVEL": "info"}, clear=True)
+def test_set_log_level_for_logger_with_env_var_and_cli():
+    """
+    Ensure that the correct log level is being set for python native logger and
+    transformers logger when env var LOG_LEVEL is used and CLI flag is passed
```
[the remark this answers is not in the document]
Abhishek-TAMU: This is done!

tuning/utils/logging.py
```diff
@@ -0,0 +1,60 @@
...
+
+# Standard
+import logging
+import os
+
+
+def set_log_level(train_args, logger_name=None):
+    """Set log level of python native logger and TF logger via argument from CLI or env variable.
+
+    Args:
```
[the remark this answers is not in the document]
Abhishek-TAMU: This is done!

Abhishek-TAMU:
We have used just `logging.getLogger()` so that logger automatically gets imported with root log level in all files but that means logs will look like this, with all logs mentioning `root`:

image

Hence, I have added an argument now in python native root logger, which logs the module name/function name for each logging statement, without us having to manually define it. Hence now the same above logs look like:

image

Thank you for the input!

aluu317 (approved):

PR #272, opened 2024-07-31
feat: Support pretokenized

kmehant:


tuning/utils/preprocessing_utils.py
```diff
@@ -29,8 +29,48 @@
 logger = logging.get_logger("sft_trainer_preprocessing")
 
 
+def _is_pretokenized_dataset(dataset: Dataset):
```
fabianlim, line 32: i suggest move this to inside `validate_data_args` and put a comment that his is how SFT checks for tokenized internally.

tuning/utils/preprocessing_utils.py
```diff
@@ -29,8 +29,48 @@
...
+    return ("input_ids" in dataset.column_names) and ("labels" in dataset.column_names)
+
+
 def validate_data_args(data_args: configs.DataArguments, packing: bool):
 
+    # validation for pretokenized datasets case
```
fabianlim, line 38: this kid of checks can be hard to parse, i suggest to document in plain english above.

tuning/utils/preprocessing_utils.py
kmehant: @fabianlim done

tuning/utils/preprocessing_utils.py
kmehant: @fabianlim documented them whereever possible thanks

tuning/utils/preprocessing_utils.py
```diff
@@ -86,11 +149,22 @@ def get_data_collator(
...
+            training data arguments
 
     Returns:
         Callable
             Callable collator to be leveraged by the trainer.
     """
+    if data_args.is_pretokenized:
+        return DataCollatorForSeq2Seq(
+            tokenizer=tokenizer,
+            padding="max_length",
```
fabianlim, line 164: @kmehant can we handle the datacollator kwargs more carefully, because this collator calls `pad()` underneath, and there are various options 

https://github.com/huggingface/transformers/blob/ef177a5e1cdf0ca53e24e6d76e813198f7300dc4/src/transformers/tokenization_utils_base.py#L3418-L3429

If we decide for the user what options, then we need to chose the best one, for example we probably do not want to chose `padding='max_length'` aor `padding=None`

fabianlim: And if there are defaults, then we can see below https://github.com/foundation-model-stack/fms-hf-tuning/pull/272/files#diff-0949af2da3611757d2552841654a34f1fe0a349ee114ce6342b9e0ade3eb13a6R180 how they are handled in `configs`

kmehant: Thanks, updated!

kmehant: > On the fence on "pt" or configs.PADDING_RETURN_TENSORS


Avoided this since collator should mostly return tensors rather python lists.

tuning/utils/preprocessing_utils.py
```diff
@@ -29,8 +29,55 @@
 logger = logging.get_logger("sft_trainer_preprocessing")
 
 
+# check if the provided dataset is pretokenized or not
+# the check is taken from trl
+# https://github.com/huggingface/trl/blob/ddf4c8dc3ecf6d9ee2b24f94c62182ffd682c808/trl/trainer/sft_trainer.py#L498-L509
+def is_pretokenized_dataset(data_path: str):
+    if data_path:
+        # load one sample from the dataset in order to inspect columns
+        dataset = datasets.load_dataset("json", data_files=data_path, split="train[:1]")
```
alex-jw-brooks, line 38: Can you add graceful handling for if the dataset is empty?

tuning/utils/preprocessing_utils.py
```diff
@@ -86,11 +133,24 @@ def get_data_collator(
...
             Callable collator to be leveraged by the trainer.
     """
+    is_train_data_pretokenized = is_pretokenized_dataset(data_path=training_data_path)
+
+    if is_train_data_pretokenized:
+        return DataCollatorForSeq2Seq(
+            tokenizer=tokenizer,
+            padding=configs.PADDING_STRATEGY_LONGEST,
+            max_length=max_sequence_length,
+            return_tensors="pt",
```
alex-jw-brooks, line 152: nit - I'd suggest not setting `return_tensors="pt"` directly here, since it is the default, and we don't set it when initializing other collators

tuning/utils/preprocessing_utils.py
```diff
@@ -29,8 +29,55 @@
...
+
+    # if the provided train dataset is pretokenized
+    # however user provides formatting flags, error out
+    if is_train_data_pretokenized:
+        if (
+            data_args.response_template
+            or data_args.data_formatter_template
+            or data_args.dataset_text_field
+        ):
+            raise ValueError(
```
alex-jw-brooks, line 62: Can you change this to a warning instead of an error? IMO this and packing=True should be handled the same way

alex-jw-brooks (changes requested):

tuning/utils/preprocessing_utils.py
```diff
@@ -68,13 +115,13 @@ def validate_data_args(data_args: configs.DataArguments, packing: bool):
     # TODO(s) In future seupport two more formats:
     # 1. Allow no response template, and JSON with input/output fields and mask input
 
-    # 2. Allow pretokenized Dataset besides JSON.
-
 
 def get_data_collator(
```
Ssukriti, line 119: unit tests need to be added for updates. see https://github.com/foundation-model-stack/fms-hf-tuning/blob/6e836d14522de9092457cf63386ae6b20eaa2a2d/tests/utils/test_preprocessing_utils.py#L129 for reference. 
Probably after rebasing, we can re-use same unit test

tuning/utils/preprocessing_utils.py
```diff
@@ -29,8 +29,55 @@
...
+        return ("input_ids" in dataset.column_names) and (
+            "labels" in dataset.column_names
+        )
+    return False
+
+
 def validate_data_args(data_args: configs.DataArguments, packing: bool):
 
+    is_train_data_pretokenized = is_pretokenized_dataset(
+        data_path=data_args.training_data_path
```
Ssukriti, line 48: unit tests need to be added as well

tuning/utils/preprocessing_utils.py
```diff
@@ -129,6 +189,21 @@ def format_dataset(data_args: configs.DataArguments, tokenizer: AutoTokenizer):
             tuple containing train_dataset, eval_dataset and dataset_text_field
     """
     eval_dataset = None
+    is_train_data_pretokenized = is_pretokenized_dataset(
+        data_path=data_args.training_data_path
```
Ssukriti, line 193: unit tests

kmehant: Added, thanks

tuning/utils/preprocessing_utils.py
kmehant: using the same unit test, thanks. 

But if you see some sort of duplicated flow in this function let me know, we will work it out, thanks

tuning/utils/preprocessing_utils.py
kmehant: modified thanks @alex-jw-brooks

tuning/utils/preprocessing_utils.py
kmehant: ah, right! updated

tuning/utils/preprocessing_utils.py
kmehant: nesting it from the same error class `DatasetGenerationError` since there are tests on malformed data already that assert on DatasetGenerationError Thanks. let me know if the message has to be refined. Thanks

tuning/utils/preprocessing_utils.py
```diff
@@ -125,10 +173,7 @@ def get_data_collator(
             )
         # Note that this automatically pads labels with -100
         # TODO check if this is sufficient for preprocessed
-        if (
-            "attention_mask" in formatted_train_dataset.column_names
-            and "labels" in formatted_train_dataset.column_names
-        ):
+        if is_train_data_pretokenized:
```
Ssukriti, line 176: packing is anyway not supported for pretokenized as well, so I just moved it here @kmehant like it was before. The check was doing same thing as your function is_pretokenized. 

padding=True is same as padding = longest . see https://huggingface.co/transformers/v4.9.2/main_classes/data_collator.html

tuning/utils/preprocessing_utils.py
```diff
@@ -33,8 +34,53 @@
...
+
+    # if the provided train dataset is pretokenized
+    # however user provides formatting flags, error out
+    if is_train_data_pretokenized:
+        if (
+            data_args.response_template
+            or data_args.data_formatter_template
+            or data_args.dataset_text_field
+        ):
+            logger.warning(
```
Ssukriti, line 65: should be a valueerror . If response template is provided, we will return completionLMcollator. There is no need to allow a response template, it is an incorrect argument.
I am committing this change

tuning/utils/preprocessing_utils.py
```diff
@@ -33,8 +34,53 @@
...
+        if data_args.validation_data_path and not is_eval_data_pretokenized:
+            raise ValueError(
+                "validation data should be pretokenized to be used \
+                along with pretokenized train data"
+            )
+
+        # packing wont be available for pretokenized datasets in trl library
+        # see: https://github.com/huggingface/trl/issues/1848
+        if packing:
+            logger.warning("packing will not be used when datasets are pretokenized")
```
Ssukriti, line 81: why is this only a warning? it is ok to be warning if code does not perform as well, like performance wont be good etc. We cant have warning when code just wont work. If flag is set to packing , it will be true even when passing to TRL

tuning/utils/preprocessing_utils.py
```diff
@@ -33,12 +34,59 @@
...
+
+
 def validate_data_args(data_args: configs.DataArguments, packing: bool):
 
     assert isinstance(
         data_args.training_data_path, str
     ), "Training data path has to be set and str"
 
+    is_train_data_pretokenized = is_pretokenized_dataset(data_args.training_data_path)
+    is_eval_data_pretokenized = is_pretokenized_dataset(data_args.validation_data_path)
```
Ssukriti, line 59: this is going to return False when validation file not set. later in line 77, it will return valueerror .

validation file is optional from top level API. Only when not None, these checks need to happen. Please add unit test

Ssukriti: oh I see you added `if data_args.validation_data_path` later on, thats fine then

tuning/utils/preprocessing_utils.py
kmehant: Thanks

tuning/utils/preprocessing_utils.py
kmehant: sure @Ssukriti thanks

tuning/utils/preprocessing_utils.py
kmehant: Agree with you, we should error out

Ssukriti:
tests are still incomplete for both validate_args and format_dataset . I understand its because the JSON is not readily available for tests. You can take a few examples from any JSON here https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/tests/data/twitter_complaints_input_output.json and convert it to pretokenized and contribute new JSON with pretokenized. Then it can be easily used in all tests

Ssukriti (approved):

alex-jw-brooks (approved):

PR #273, opened 2024-07-31
fix run evaluation to get base model path

anhuong:
### Description of the change

The base model name is now required when running `TunedCausalLM.load()` due to [recent change ](https://github.com/foundation-model-stack/fms-hf-tuning/commit/3be607bed91ca50ebc7ac16a8bf8bae7d3201545) that went in which is breaking the run_evaluation script. run_evaluation does not have a way of passing in the base model path and did not need to pass this in because `AutoPeftModelForCausalLM()` handled loading the peft model.

This PR fixes this by:
- Adding flag where user can pass in `base_model_name_or_path`
- If user does not pass in `base_model_name_or_path`, then this value is tried to be pulled from the checkpoint's adapter_config.json

The flag is not required since if a user is running with a fine tuned model, there will be no adapter config nor a base model needed.

### Was the PR tested


scripts/run_evaluation.py
```diff
@@ -446,7 +450,20 @@ def export_experiment_info(
...
+
+    base_model_name_or_path = args.base_model_name_or_path
+    if not base_model_name_or_path:
+        adapter_config_path = os.path.join(args.model, "adapter_config.json")
+        if os.path.exists(adapter_config_path):
+            with open(adapter_config_path, "r", encoding="utf-8") as config_file:
+                adapter_config = json.load(config_file)
+                base_model_name_or_path = adapter_config.get("base_model_name_or_path")
+
+    tuned_model = TunedCausalLM.load(
```
aluu317, line 462: If I'm reading this correctly, this code is missing the case that a user is running evaluation on a fine tuned model and there's no `adapter_config.json` file and they don't pass in any value for `base_model_name_or_path`. Then the model will fail to load, and an exception is raised, is that correct?

aluu317 (changes requested):
This is good to know, and a need to merge before I use evaluation script for quality testing. Just one comment inline for the fine tuning base model case

scripts/run_evaluation.py
anhuong: That's a good point, the problem is that fine tuned models don't have an adapter_config.json nor is a base model needed to load the model. So this may be another bug in the run_inference script for fine tuned models that needs to be fixed. Let me try it out with fine tuned model...

aluu317:
Should probably update https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/scripts/evaluation.md#running-evaluation  and put a note about fine-tuning vs. other cases

scripts/run_evaluation.py
anhuong: @aluu317 verified that both changes did not affect the fine tuning case. No change is needed for fine tuning because of [this line](https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/scripts/run_inference.py#L182) in `TunedCausalLM.load()`. Since no adapter_config.json is found, this jumps below to the except block where the model is correctly loaded as a merged model.

aluu317 (approved):
LGTM!

PR #274, opened 2024-07-31
deps: limit peft deps 

anhuong:
### Description of the change

We want to investigate the new changes in PEFT before upgrading.

### Was the PR tested


jbusche (approved):
LGTM

PR #276, opened 2024-07-31
Releasev1.1.0

jbusche:
### Description of the change

Release v1.1.0

### Related issue number

Closes #1135
### Was the PR tested
Yes, tested with rc.1 candidate image.


anhuong (approved):
LGTM, thanks Jim! Note that DCO had the same issue where we had to manually set it to pass. Looked through the commits and they look good so verified 👍

tedhtchang (approved):

PR #280, opened 2024-08-02
feat: Add DataClass Arguments to Activate Padding-Free and MultiPack Plugin and FastKernels

achew010:
### Description of the change

This PR adds two dataclass arguments to enable *padding free* and *multipack* for the `sft_trainer.py`, via the new `fms acceleration` [attention-and-distributed-packing](https://github.com/foundation-model-stack/fms-acceleration/tree/main/plugins/attention-and-distributed-packing) plugin and allows the current `--fastkernels` dataclass to support optimized full-finetuning:
-  `--padding_free`: technique to process multiple examples in single batch without adding padding tokens that waste compute.
-  `--multipack`: technique for *multi-gpu training* to balance out number of tokens processed in each device, to minimize waiting time.
- `--fast_kernels`: Previously limited only for QPEFT (used to raise if not activated with `--fast_lora`), Now allows for optimized full/standard LoRA finetuning.

These are extremely effective methods to improve training throughputs:
- see the section on [benchmarks](#benchmarks) below. Currently, either padding free is used **alone**, or **together** with multipack. We *do not currently support* the option of using multipack alone.
- padding free and multipack used in the *instructlab* (ILAB) work, see below on the section of the [early version](#early-version) of this plugin. For general use when producing this plugin, we have greatly simplified the user interface

**NOTE**: adhering to the design of [fms-acceleration](https://github.com/foundation-model-stack/fms-acceleration), the new plugin is **optional**, and separately installed. 
- there is no dependency on the [fms-acceleration-peft](https://github.com/foundation-model-stack/fms-acceleration/tree/main/plugins/accelerated-peft) for GPTQ LoRA. However we [have tested it with other plugins](https://github.com/foundation-model-stack/fms-acceleration/pull/66):
    * [accelerated-peft](https://github.com/foundation-model-stack/fms-acceleration/tree/main/plugins/accelerated-peft)
    * [fused-ops-and-kernels](https://github.com/foundation-model-stack/fms-acceleration/tree/main/plugins/fused-ops-and-kernels)
   

**Notes on Padding Free**
- works for both *single* and *multi-gpu*. 
- works on both *pretokenized* and *untokenized* datasets
- works with current transformers releases (`<=4.43`), when padding free is *not yet integrated* from our PR into Hugging Face: https://github.com/huggingface/transformers/pull/31629.
- is forward-compatible with future transformers that natively supports padding-free (`>= 4.44`).
- verified against the version found in HF main, merged in via PR https://github.com/huggingface/transformers/pull/31629.

**Notes on Multipack**
- works only for *multi-gpu*.
- currently only includes the version of *multipack* optimized for linear attention implementations like *flash-attn*.

**Notes on FastKernels**
- currently supports FastCrossEntropyLoss, FastRoPE, FastRMSLayerNorm but will include SwiGLU and Liger Kernels (e.g. FusedCrossEntropyLoss)  in the future
- Works for full-finetuning, LoRA and QPEFT, 
    - pass `--fast_kernels True True True` on full finetuning/LoRA runs
    - pass `--fast_kernels True True True --auto_gptq triton_v2 --fused_lora auto_gptq True` for GPTQ-LoRA
    - pass `--fast_kernels True True True --bitsandbytes nf4 --fused_lora bitsandbytes True` for QLoRA
- FastRoPE currently doesn't accept `positional_ids` but [this issue](https://github.com/foundation-model-stack/fms-acceleration/issues/33) will be addressed in the future 

#### Benchmarks

**PaddingFree and Multipack Benchmarks for Mistral 7B** 

Notes:
- *Shown below are the runtimes for running a subset of 6000 FLAN samples*.
- Tested two cases of per device batch sizes 4 and 8, for varying number gpus from 2 to 8
- Verified that untokenized dataset produces the same improvements for paddingfree and multipack

**Per Device Batch Size 4**

| Framework Config | Num Devices | Per Device Batch Size | Train Runtime (secs) | Speedups |
| -------- | ------- | ------- | ------- | ------- |
| full-FT | 2 | 4 | 1537 | baseline |
| padding-free | 2 | 4 | 859 | 1.79 x |
| padding-free + multipack | 2 | 4 | 751 | 2.05 x |
| full-FT | 4 | 4 | 932 | baseline |
| padding-free | 4 | 4 | 483 | 1.93 x |
| padding-free + multipack | 4 | 4 | 342 | 2.75 x |
| full-FT | 8 | 4 | 551 | baseline |
| padding-free | 8 | 4 | 275 | 2.00 x |
| padding-free + multipack | 8 | 4 | 163 | 3.38 x |

**Per Device Batch Size 8**
| Framework Config | Num Devices | Per Device Batch Size | Train Runtime (secs) | Speedup |
| -------- | ------- | ------- | ------- | ------- |
| full-FT | 2 | 8 | 1722 | baseline |
| padding-free | 2 | 8 | 678 | 2.54 x |
| padding-free + multipack | 2 | 8 | 603 | 2.86 x |
| full-FT | 4 | 8 | 1025 | baseline |
| padding-free | 4 | 8 | 380 | 2.70 x |
| padding-free + multipack | 4 | 8 | 289 | 3.55 x |
| full-FT | 8 | 8 | 611 | baseline |
| padding-free | 8 | 8 | 215 | 2.84 x |
| padding-free + multipack | 8 | 8 | 140 | 4.36 x |
 

 
**Verified Similar Improvements for Untokenized Dataset**
| Framework Config | Num Devices | Per Device Batch Size | Train Runtime (secs) | Speedups |
| -------- | ------- | ------- | ------- | ------- |
| full-FT | 2 | 4 | 1516 | baseline |
| padding-free | 2 | 4 | 848 | 1.78x |
| padding-free + multipack | 2 | 4 | 747 | 2.02x |

**Full Finetuning Benchmarks for Mistral 7B** 

#### Early Version Of This Plugin

We have an [unofficial version](https://github.com/foundation-model-stack/fms-acceleration/tree/refactor/attention) with more features than our present release. @kmehant  is currently using for ILAB work. It addition to the padding-free and multipack, it also has the additional two plugins below:

- [LossAccelerationPlugin](https://github.com/foundation-model-stack/fms-acceleration/blob/refactor/attention/plugins/accelerated-attn/src/fms_acceleration_attn/framework_plugin_loss.py): various methods of averaging losses across GPUs in distributed training, like straight per-token averaging.
- [MLPDropoutAccelerationPlugin](https://github.com/foundation-model-stack/fms-acceleration/blob/refactor/attention/plugins/accelerated-attn/src/fms_acceleration_attn/framework_plugin_mlp_dropout.py). Adding Dolomite-style residual dropout to MLP units. 

To use the early version a quick hack of `sft_trainer` with pretokenized + custom tokenizer: https://github.com/fabianlim/fms-hf-tuning/tree/attn-plugin . **This will be superceded by this PR in the near future**

Use with these command line arugments:
```
	  --padding_free huggingface-injected \
	  --loss_across_gpus mean token \
```

### How to verify the PR

Additional checks/tests were added to
1. Ensures parsing `--padding_free` and `multipack` is correct in `test_dataclass_parse_successfully`
2. Ensures wrong arguments to `--padding_free` are caught in `test_dataclass_will_fail_to_accept_illegal_args`
3. Ensures Plugin is successfully instantiated from dataclass in `test_framework_initialize_and_trains_with_aadp`
4. Ensure `--padding_free` must be used with flash-attn, otherwise error is raised
5. Ensure `--multi_pack` must be used with `--padding_free`, otherwise error is raised
6. Ensure `--packing True`  with `--padding_free` will raise an error
7. Ensure `--fast_kernels` works with full finetuning
8. Ensure that `--fast_lora` not called with either `--auto_gptq` or `--bitsandbytes` will raise an error

Ran the full suite of acceleration checks to verify all fms-acceleration unit tests passed
```
pytest tests/acceleration/
```
image

### Was the PR tested

### DIFF
--- a/README.md
+++ b/README.md
@@ -621,15 +621,26 @@ The list of configurations for various `fms_acceleration` plugins:
 - [fused_ops_and_kernels](./tuning/config/acceleration_configs/fused_ops_and_kernels.py) (experimental):
   - `--fused_lora`: fused lora for more efficient LoRA training.
   - `--fast_kernels`: fast cross-entropy, rope, rms loss kernels.
+- [attention_and_distributed_packing](./tuning/config/acceleration_configs/attention_and_distributed_packing.py) (experimental):
+  - `--padding_free`: technique to process multiple examples in single batch without adding padding tokens that waste compute.
+  - `--multipack`: technique for *multi-gpu training* to balance out number of tokens processed in each device, to minimize waiting time.
 
 Notes: 
  * `quantized_lora_config` requires that it be used along with LoRA tuning technique. See [LoRA tuning section](https://github.com/foundation-model-stack/fms-hf-tuning/tree/main?tab=readme-ov-file#lora-tuning-example) on the LoRA parameters to pass.
  * When setting `--auto_gptq triton_v2` plus note to also pass `--torch_dtype float16` and `--fp16`, or an exception will be raised. This is because these kernels only support this dtype.
- * Currently, the `fused_ops_and_kernels` is to be used used together QLoRA or GPTQ-LORA via the `quantized_lora_config`. In the future it may be made more flexible such that `fast_kernels` can even be used with full-finetuning.
  * When using `fused_ops_and_kernels` together with `quantized_lora_config`,
  make sure to appropriately set `--fused_lora auto_gptq True` or `bitsandbytes True`; the `True` sets `fast_lora==True`.
- * Currently `fused_ops_and_kernels` only supports activating `fast_loss,fast_rsm_layernorm,fast_rope_embeddings` all to `True`, so pass `--fast_kernels True True True`.
-
+ * `fused_ops_and_kernels` works for full-finetuning, LoRA, QLoRA and GPTQ-LORA, 
+    - pass `--fast_kernels True True True` for full finetuning/LoRA
+    - pass `--fast_kernels True True True --auto_gptq triton_v2 --fused_lora auto_gptq True` for GPTQ-LoRA
+    - pass `--fast_kernels True True True --bitsandbytes nf4 --fused_lora bitsandbytes True` for QLoRA
+ * Notes on Padding Free
+  - works for both *single* and *multi-gpu*. 
+  - works on both *pretokenized* and *untokenized* datasets
+  - verified against the version found in HF main, merged in via PR https://github.com/huggingface/transformers/pull/31629.
+ * Notes on Multipack
+  - works only for *multi-gpu*.
+  - currently only includes the version of *multipack* optimized for linear attention implementations like *flash-attn*.
 
 Activate `TRANSFORMERS_VERBOSITY=info` to see the huggingface trainer printouts and verify that `AccelerationFramework` is activated!
 
--- a/README.md
+++ b/README.md
@@ -635,12 +635,12 @@ Notes:
     - pass `--fast_kernels True True True --auto_gptq triton_v2 --fused_lora auto_gptq True` for GPTQ-LoRA
     - pass `--fast_kernels True True True --bitsandbytes nf4 --fused_lora bitsandbytes True` for QLoRA
  * Notes on Padding Free
-  - works for both *single* and *multi-gpu*. 
-  - works on both *pretokenized* and *untokenized* datasets
-  - verified against the version found in HF main, merged in via PR https://github.com/huggingface/transformers/pull/31629.
+    - works for both *single* and *multi-gpu*. 
+    - works on both *pretokenized* and *untokenized* datasets
+    - verified against the version found in HF main, merged in via PR https://github.com/huggingface/transformers/pull/31629.
  * Notes on Multipack
-  - works only for *multi-gpu*.
-  - currently only includes the version of *multipack* optimized for linear attention implementations like *flash-attn*.
+    - works only for *multi-gpu*.
+    - currently only includes the version of *multipack* optimized for linear attention implementations like *flash-attn*.
 
 Activate `TRANSFORMERS_VERBOSITY=info` to see the huggingface trainer printouts and verify that `AccelerationFramework` is activated!
 
--- a/tuning/config/acceleration_configs/acceleration_framework_config.py
+++ b/tuning/config/acceleration_configs/acceleration_framework_config.py
@@ -22,6 +22,7 @@ import yaml
 
 # Local
 from .fused_ops_and_kernels import FastKernelsConfig, FusedLoraConfig
+from .instruct_lab_config import PaddingFree
 from .quantized_lora_config import AutoGPTQLoraConfig, BNBQLoraConfig
 from tuning.utils.import_utils import is_fms_accelerate_available
 
@@ -95,6 +96,15 @@ class AccelerationFrameworkConfig:
             key="fused_ops_and_kernels",
             experimental=True,
             required_packages=["foak"],
+        ),
+    ] = None
+
+    padding_free: Annotated[
+        PaddingFree,
+        ConfigAnnotation(
+            path="training.attention",
+            experimental=True,
+            required_packages=["ilab"],
         ),
     ] = None
 
--- a/tuning/config/acceleration_configs/acceleration_framework_config.py
+++ b/tuning/config/acceleration_configs/acceleration_framework_config.py
@@ -22,7 +22,7 @@ import yaml
 
 # Local
 from .fused_ops_and_kernels import FastKernelsConfig, FusedLoraConfig
-from .instruct_lab_config import PaddingFree
+from .attention_and_distributed_packing import PaddingFree
 from .quantized_lora_config import AutoGPTQLoraConfig, BNBQLoraConfig
 from tuning.utils.import_utils import is_fms_accelerate_available
 
@@ -104,7 +104,7 @@ class AccelerationFrameworkConfig:
         ConfigAnnotation(
             path="training.attention",
             experimental=True,
-            required_packages=["ilab"],
+            required_packages=["aadp"],
         ),
     ] = None
 
--- a/tuning/config/acceleration_configs/acceleration_framework_config.py
+++ b/tuning/config/acceleration_configs/acceleration_framework_config.py
@@ -22,7 +22,7 @@ import yaml
 
 # Local
 from .fused_ops_and_kernels import FastKernelsConfig, FusedLoraConfig
-from .attention_and_distributed_packing import PaddingFree
+from .attention_and_distributed_packing import PaddingFree, MultiPack
 from .quantized_lora_config import AutoGPTQLoraConfig, BNBQLoraConfig
 from tuning.utils.import_utils import is_fms_accelerate_available
 
@@ -108,6 +108,16 @@ class AccelerationFrameworkConfig:
         ),
     ] = None
 
+    multipack: Annotated[
+        MultiPack,
+        ConfigAnnotation(
+            path="training.dataloader",
+            experimental=True,
+            required_packages=["aadp"],
+        ),
+    ] = None
+
+
     @staticmethod
     def from_dataclasses(*dataclasses: Type):
         "Convert one or many FMS config dataclasses to a monolithic AccelerationConfig"
@@ -125,6 +135,7 @@ class AccelerationFrameworkConfig:
 
         # first unroll all the dataclases into a single level
         nested_dataclasses = []
+
         for dc in dataclasses:
             if dc is None:
                 continue
@@ -172,7 +183,6 @@ class AccelerationFrameworkConfig:
         return config
 
     def get_framework(self):
-
         if is_fms_accelerate_available():
 
             # to be eventually be made to be passed as a dict to Acceleration
--- a/tuning/config/acceleration_configs/acceleration_framework_config.py
+++ b/tuning/config/acceleration_configs/acceleration_framework_config.py
@@ -84,7 +84,7 @@ class AccelerationFrameworkConfig:
         ConfigAnnotation(
             path="peft.quantization",
             key="fused_ops_and_kernels",
-            experimental=True,
+            experimental=False,
             required_packages=["foak"],
         ),
     ] = None
@@ -94,7 +94,7 @@ class AccelerationFrameworkConfig:
         ConfigAnnotation(
             path="training",
             key="fused_ops_and_kernels",
-            experimental=True,
+            experimental=False,
             required_packages=["foak"],
         ),
     ] = None
--- a/tuning/config/acceleration_configs/acceleration_framework_config.py
+++ b/tuning/config/acceleration_configs/acceleration_framework_config.py
@@ -117,6 +117,14 @@ class AccelerationFrameworkConfig:
         ),
     ] = None
 
+    def _verify_configured_dataclasses(self):
+        if self.multipack is not None:
+            # ensure if multipack is set, padding free is also turned on as well
+            # this also ensures that the attention implementation for multipack
+            # will be flash attention as sfttrainer will enforce flash attn to be
+            # set for padding free
+            assert self.padding_free is not None, "`--multipack` is currently only supported with `--padding_free`"
+
     @staticmethod
     def from_dataclasses(*dataclasses: Type):
         "Convert one or many FMS config dataclasses to a monolithic AccelerationConfig"
@@ -179,6 +187,8 @@ class AccelerationFrameworkConfig:
             setattr(config, fi.name, dc)
             del rem_fields[fi.name]  # remove the field
 
+        # perform some checks on dataclasse
+        config._verify_configured_dataclasses()
         return config
 
     def get_framework(self):
--- a/tuning/config/acceleration_configs/acceleration_framework_config.py
+++ b/tuning/config/acceleration_configs/acceleration_framework_config.py
@@ -123,7 +123,9 @@ class AccelerationFrameworkConfig:
             # this also ensures that the attention implementation for multipack
             # will be flash attention as sfttrainer will enforce flash attn to be
             # set for padding free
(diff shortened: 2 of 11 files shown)


kmehant (changes requested):
@achew010 lets gracefully handle the case when use_flash_attn is set to False and padding free is being used. 

https://github.com/foundation-model-stack/fms-hf-tuning/blob/d7280073f8121ff2d0b160302092ddc13010852f/tuning/config/configs.py#L39

tuning/sft_trainer.py
```diff
@@ -123,6 +128,16 @@ def train(
...
 
+    if (
+        attention_and_distributed_packing_config is not None
+        and attention_and_distributed_packing_config.padding_free is not None
+        and model_args.use_flash_attn is False
+    ):
+        raise ValueError(
+            "`--padding_free` argument was called without enabling \
+            flash attention, ensure `use_flash_attn = True` to use padding-free flash attention"
+        )
```
fabianlim, line 139: Lets also have a `packing=False` check when `padding_free` is activated. Put a comment to explain that when using `padding_free`, then we should not attempt to perform any packing using `sft_trainer`

tuning/config/acceleration_configs/attention_and_distributed_packing.py
```diff
@@ -0,0 +1,35 @@
...
+
+@parsable_dataclass
+@dataclass
+class PaddingFree:
+
+    method: str = "huggingface"
+
+    def __post_init__(self):
+        if self.method != "huggingface":
+            raise ValueError("only 'huggingface' method currently supported.")
```
anhuong, line 16: Can you describe what "huggingface" method padding free means compared to other methods?

tuning/config/acceleration_configs/attention_and_distributed_packing.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+@parsable_dataclass
+@dataclass
+class MultiPack:
+
+    num_processes: int = 16
```
anhuong, line 23: Is there any guidance on what this number should be set to?

tuning/config/acceleration_configs/attention_and_distributed_packing.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+@dataclass
+class AttentionAndDistributedPackingConfig:
+
+    padding_free: PaddingFree = None
+
+    multipack: MultiPack = None
```
anhuong, line 31: So this would be called `python sft_trainer.py --multipack <int> --padding_free "huggingface"` correct?

tuning/config/acceleration_configs/acceleration_framework_config.py
```diff
@@ -91,13 +92,48 @@ class AccelerationFrameworkConfig:
...
+            # set for padding free
+            assert (
+                self.padding_free is not None
+            ), "`--multipack` is currently only supported with `--padding_free`"
+
+        # Check that fused lora must be activated with either auto_gptq or bitsandbytes
+        if self.fused_lora is not None:
+            assert self.bitsandbytes is not None or self.auto_gptq is not None, (
+                "`--fused_lora` must be accompanied by a quantized base layer"
+                " `--auto_gptq` or `--bitsandbytes`."
```
anhuong, line 134: I appreciate the checks but think these should be raised ValueErrors instead of AssertionErrors

tuning/config/acceleration_configs/acceleration_framework_config.py
```diff
@@ -91,13 +92,48 @@ class AccelerationFrameworkConfig:
...
+        MultiPack,
+        ConfigAnnotation(
+            path="training.dataloader",
+            experimental=True,
+            required_packages=["aadp"],
+        ),
+    ] = None
+
+    def _verify_configured_dataclasses(self):
+        if self.multipack is not None:
```
anhuong, line 121: What happens if multipack is set but only on single-GPU? Should there be a check to fail if a user sets this?

tuning/sft_trainer.py
```diff
@@ -174,7 +197,9 @@ def train(
             trainer_callbacks.append(cb)
 
     framework = AccelerationFrameworkConfig.from_dataclasses(
-        quantized_lora_config, fusedops_kernels_config
+        quantized_lora_config,
+        fusedops_kernels_config,
+        attention_and_distributed_packing_config,
```
anhuong, line 202: Just for my understanding, so these are all model loader augmentors that change how the model is loaded based on the acceleration framework configurations? Although padding free and multipack are both dataset augmentors? How does setting the acceleration framework here affect the dataset loading?

tuning/sft_trainer.py
```diff
@@ -123,6 +128,24 @@ def train(
...
+        attention_and_distributed_packing_config is not None
+        and attention_and_distributed_packing_config.padding_free is not None
+    ):
+        if model_args.use_flash_attn is False:
+            raise ValueError(
+                "`--padding_free` argument was called without enabling flash attention, "
+                "ensure `use_flash_attn = True` to use padding-free flash attention"
+            )
+
+        if train_args.packing is True:
```
anhuong, line 141: nit: can simplify to `if train_args.packing`

anhuong (commented):
Thank you for the excellent change and description! Had a few questions...also am wondering, should the plugin be installed by default so users can utilize these new parameters? Looks like a very useful addition. 

Also please add some of the great description from this PR into the readme.

tuning/config/acceleration_configs/attention_and_distributed_packing.py
fabianlim: "huggingface" means the recent PR made by the team https://github.com/huggingface/transformers/pull/31629 that uses FA2. I left it extensible in the potential event that we want to support other variants of padding free attention (e.g., native torch SDPA with nested tensors). No concrete plans now, but future proofing.

tuning/config/acceleration_configs/attention_and_distributed_packing.py
fabianlim: this number is a reasonable one for most datasets of reasonable size (e.g. under a million examples). The packing algorithm is relatively fast, but in the event the dataset is too large, then our plugin will raise a warning
https://github.com/foundation-model-stack/fms-acceleration/blob/4e81c64453ec5d2b06a8d14a2a72374cc736098a/plugins/attention-and-distributed-packing/src/fms_acceleration_aadp/framework_plugin_multipack.py#L117-L123

that advises the user to increase this number if  the process times out.

tuning/config/acceleration_configs/attention_and_distributed_packing.py
fabianlim: correct

tuning/config/acceleration_configs/acceleration_framework_config.py
fabianlim: ok @achew010 can you help make the changes?

tuning/config/acceleration_configs/acceleration_framework_config.py
fabianlim: we have a check in our plugin to raise https://github.com/foundation-model-stack/fms-acceleration/blob/4e81c64453ec5d2b06a8d14a2a72374cc736098a/plugins/attention-and-distributed-packing/src/fms_acceleration_aadp/framework_plugin_multipack.py#L95-L98

tuning/sft_trainer.py
fabianlim: you are right in saying padding free and multipack affect the dataloading, but more specifically
- padding free only requires modifications to data collation.
- multpack requires modification to dataloader


Both we handle by our [`AccelerationPatcher`](https://github.com/foundation-model-stack/fms-acceleration/blob/main/plugins/framework/src/fms_acceleration/accelerator_patcher.py), which is a component that we wrote to allow controlled replacements of the data collator and data loader.

fabianlim:
@anhuong thanks for the review. For making this default, I drafted out various possibilities in this issue here https://github.com/foundation-model-stack/fms-hf-tuning/issues/334. We can discuss offline,

anhuong:
Also we added the new automation that ensure PRs follow convention commits which you can see is failing -- https://github.com/foundation-model-stack/fms-hf-tuning/actions/runs/10920573842/job/30310716778?pr=280 please address the change

README.md
```diff
@@ -455,15 +455,26 @@ The list of configurations for various `fms_acceleration` plugins:
...
- * Currently `fused_ops_and_kernels` only supports activating `fast_loss,fast_rsm_layernorm,fast_rope_embeddings` all to `True`, so pass `--fast_kernels True True True`.
-
+ * `fused_ops_and_kernels` works for full-finetuning, LoRA, QLoRA and GPTQ-LORA, 
+    - pass `--fast_kernels True True True` for full finetuning/LoRA
+    - pass `--fast_kernels True True True --auto_gptq triton_v2 --fused_lora auto_gptq True` for GPTQ-LoRA
+    - pass `--fast_kernels True True True --bitsandbytes nf4 --fused_lora bitsandbytes True` for QLoRA
+ * Notes on Padding Free
+  - works for both *single* and *multi-gpu*. 
+  - works on both *pretokenized* and *untokenized* datasets
+  - verified against the version found in HF main, merged in via PR https://github.com/huggingface/transformers/pull/31629.
```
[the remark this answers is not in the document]
achew010: Oops, i fixed it

README.md
```diff
@@ -455,15 +455,26 @@ The list of configurations for various `fms_acceleration` plugins:
...
  * When setting `--auto_gptq triton_v2` plus note to also pass `--torch_dtype float16` and `--fp16`, or an exception will be raised. This is because these kernels only support this dtype.
- * Currently, the `fused_ops_and_kernels` is to be used used together QLoRA or GPTQ-LORA via the `quantized_lora_config`. In the future it may be made more flexible such that `fast_kernels` can even be used with full-finetuning.
  * When using `fused_ops_and_kernels` together with `quantized_lora_config`,
  make sure to appropriately set `--fused_lora auto_gptq True` or `bitsandbytes True`; the `True` sets `fast_lora==True`.
- * Currently `fused_ops_and_kernels` only supports activating `fast_loss,fast_rsm_layernorm,fast_rope_embeddings` all to `True`, so pass `--fast_kernels True True True`.
-
+ * `fused_ops_and_kernels` works for full-finetuning, LoRA, QLoRA and GPTQ-LORA, 
+    - pass `--fast_kernels True True True` for full finetuning/LoRA
+    - pass `--fast_kernels True True True --auto_gptq triton_v2 --fused_lora auto_gptq True` for GPTQ-LoRA
+    - pass `--fast_kernels True True True --bitsandbytes nf4 --fused_lora bitsandbytes True` for QLoRA
```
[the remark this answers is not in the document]
fabianlim: Yes that is correct, but unfortunately that will be more complicated than the current implementation. 
- Consider the plugin `dataclass` (e.g., `FusedOpsAndKernelsConfig`), see [here](https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/tuning/config/acceleration_configs/fused_ops_and_kernels.py#L72) 
- the plugin dataclass is a nested `dataclass`; this is because it has `dataclasses` as members.
- each member dataclass needs to be parsable by `HfArgumentParser`, which actually does not support parsing a `dataclass` type.
- hence, we made it possible due to our  [parsable_dataclass](https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/tuning/config/acceleration_configs/utils.py#L62) decorator, that
  - masquarades the member dataclass as a `List`, which `HfArgumentParser` does support lists of a *uniform type*.
  - allows our member dataclass to contain *mixed types* by the casting logic implemented in `parsable_dataclass` via [EnsureTypes](https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/tuning/config/acceleration_configs/utils.py#L38C7-L38C18).

Hence this allows `--fast_kernels False True False` to be parsed into the dataclass `FusedOpsAndKernelsConfig`. 

To support parsing of the kind `--fast_kernels FastRoPE`, we need to handle
- handling of different types. "FastRoPE" is clearly a boolean type, but we also need to handle `str` inputs, `float` inputs etc, where it would need to be a `key=value` pair
- handling different orders, we need to be able to parse `--dataclass_key key_a=a key_b` and `--dataclass key_b key_a=a` equivalently.
- handling defaults, if `key_a` is missing then we need to set it a default. We need to raise if users try to define the dataclass without defaults.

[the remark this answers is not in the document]
fabianlim: I suggest to merge this PR first, we can come back to this again when we do the fused cross entropy. BTW i left a comment for @achew010 to help upgrade `FusedOpsAndKernelsConfig` to non-experimental status in this PR by deleting these `experimental=True` entries, see [here](https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/tuning/config/acceleration_configs/acceleration_framework_config.py#L86)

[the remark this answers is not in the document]
anhuong: Thank you for the details, I agree getting this merged and thinking about improvements for fast_kernels later makes sense. Is just the `FusedOpsAndKernelsConfig` ready to move out of experimental or can this also be done for PaddingFree and Multipack?

anhuong:
Please update the branch with the new changes from `main` and then once the experimental fields are updated this is good to merge in to me 👍

anhuong:
Note @kmehant I think since you requested changes, an approval is needed from your side as well before this can merge

kmehant (approved):
@anhuong thanks for letting me know. Its really annoying that I am not able to dismiss my review someway so that I do not stand as a blocker :( forcing me to push a approval. 

Nonetheless, I have used most of these features as part of iLab and undoubtedly vouch for the changes. Thanks.

anhuong (approved):
We can also mark paddingFree and multiPack as not experimental but LGTM

PR #269, opened 2024-07-29
Set default value of target_modules to be None in LoraConfig

willmj:
### Description of the change
Original issue (from Slack):
> By default target_modules None should work, and it will use the default values [HF has here](https://github.com/huggingface/peft/blob/7b1c08d2b5e13d3c99b7d6ee83eab90e1216d4ba/src/peft/utils/constants.py#L70) for each architecture when not specified  . In [PEFT LoraConfig](https://huggingface.co/docs/peft/en/package_reference/lora#peft.LoraConfig) default is None.
> In our Config also we should set it to None in [tuning/config/peft_config.py](https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/tuning/config/peft_config.py#L48) and ensure it still works with None through unit tests.

In short, changed
`default_factory=lambda: ["q_proj", "v_proj"],` to `default=None,` in the LoraConfig class.

### Related issue number
N/A

### How to verify the PR
Run unit tests.
Additionally, you can run tuning on a model, and on the output model check target_modules in adapter_config.json matches the intended target_modules of the model used.

### Was the PR tested
- [ ] I have added >=1 unit test(s) for every new method I have added. 
- [x] I have ensured all unit tests pass

If you have any questions or concerns, please respond below. Thanks!

PR #281, opened 2024-08-02
Add unit test to verify target_modules defaults correctly

willmj:
### Description of the change
Add unit test for recent change (PR #269) to verify that HF library correctly changes target_modules to the default values.

### How to verify the PR
Run unit tests.

### Was the PR tested

### DIFF
--- a/tests/build/test_launch_script.py
+++ b/tests/build/test_launch_script.py
@@ -138,31 +138,6 @@ def test_successful_lora():
         assert os.path.exists(tempdir + "/adapter_config.json") is True
 
 
-def test_successful_lora_target_modules_defaults():
-    """Check if target modules automatically set to their default values"""
-    with tempfile.TemporaryDirectory() as tempdir:
-        setup_env(tempdir)
-        TRAIN_KWARGS = {**BASE_LORA_KWARGS, **{"output_dir": tempdir}}
-        serialized_args = serialize_args(TRAIN_KWARGS)
-        os.environ["SFT_TRAINER_CONFIG_JSON_ENV_VAR"] = serialized_args
-
-        assert main() == 0
-        # check termination log and .complete files
-        assert os.path.exists(tempdir + "/adapter_config.json") is True
-
-        # Load the adapter_config.json file and check target_modules
-        with open(tempdir + "/adapter_config.json", "r", encoding="utf-8") as f:
-            config = json.load(f)
-
-        assert (
-            "target_modules" in config
-        ), "target_modules not found in adapter_config.json."
-        assert set(config["target_modules"]) == {
-            "q_proj",
-            "v_proj",
-        }, "target_modules are not set to the default values."
-
-
 def test_bad_script_path():
     """Check for appropriate error for an invalid training script location"""
     with tempfile.TemporaryDirectory() as tempdir:
--- a/tests/build/test_launch_script.py
+++ b/tests/build/test_launch_script.py
@@ -19,7 +19,6 @@
 import os
 import tempfile
 import glob
-import json
 
 # Third Party
 import pytest
--- a/tests/build/test_launch_script.py
+++ b/tests/build/test_launch_script.py
@@ -19,6 +19,7 @@
 import os
 import tempfile
 import glob
+import json
 
 # Third Party
 import pytest
@@ -135,6 +136,31 @@ def test_successful_lora():
         assert os.path.exists(os.path.join(tempdir, ".complete")) is True
         assert os.path.exists(tempdir + "/adapter_model.safetensors") is True
         assert os.path.exists(tempdir + "/adapter_config.json") is True
+
+
+def test_successful_lora_target_modules_defaults():
+    """Check if target modules automatically set to their default values"""
+    with tempfile.TemporaryDirectory() as tempdir:
+        setup_env(tempdir)
+        TRAIN_KWARGS = {**BASE_LORA_KWARGS, **{"output_dir": tempdir}}
+        serialized_args = serialize_args(TRAIN_KWARGS)
+        os.environ["SFT_TRAINER_CONFIG_JSON_ENV_VAR"] = serialized_args
+
+        assert main() == 0
+        # check termination log and .complete files
+        assert os.path.exists(tempdir + "/adapter_config.json") is True
+
+        # Load the adapter_config.json file and check target_modules
+        with open(tempdir + "/adapter_config.json", "r", encoding="utf-8") as f:
+            config = json.load(f)
+
+        assert (
+            "target_modules" in config
+        ), "target_modules not found in adapter_config.json."
+        assert set(config["target_modules"]) == {
+            "q_proj",
+            "v_proj",
+        }, "target_modules are not set to the default values."
 
 
 def test_bad_script_path():
--- a/tests/test_sft_trainer.py
+++ b/tests/test_sft_trainer.py
@@ -31,6 +31,7 @@ import transformers
 # First Party
 from build.utils import serialize_args
 from scripts.run_inference import TunedCausalLM
+from tests.build.test_launch_script import BASE_LORA_KWARGS
 from tests.data import (
     EMPTY_DATA,
     MALFORMATTED_DATA,
@@ -39,10 +40,6 @@ from tests.data import (
     TWITTER_COMPLAINTS_DATA_INPUT_OUTPUT,
     TWITTER_COMPLAINTS_JSON_FORMAT,
     TWITTER_COMPLAINTS_TOKENIZED,
-)
-from tests.build.test_launch_script import (
-    BASE_KWARGS, 
-    BASE_LORA_KWARGS
 )
 
 # Local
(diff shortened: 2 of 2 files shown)


tests/build/test_launch_script.py
```diff
@@ -137,6 +138,31 @@ def test_successful_lora():
         assert os.path.exists(tempdir + "/adapter_config.json") is True
 
 
+def test_successful_lora_target_modules_defaults():
+    """Check if target modules automatically set to their default values"""
```
aluu317, line 142: Please be more specific about what this means. Something like: "Check that if target_modules is not set, or set to None, the default value by model type will be using in training". And explain the test:
"We use `TRAIN_KWARGS` to run training with, which has no `target_modules` set. During HF training process, the correct default target modules will be used for model type Llama and "q_proj", "v_proj" will then exist in the resulting in adapter_config.json."

And link to https://github.com/huggingface/peft/blob/7b1c08d2b5e13d3c99b7d6ee83eab90e1216d4ba/src/peft/tuners/lora/model.py#L432 for how it is handled in HF

aluu317 (changes requested):
Thank you @willmj . 
This is helpful! 
- I would argue this test isn't just to test a launch script functionality, but rather should exist in `test_sft_trainer`. It is more about `sft_trainer`.
- I would also suggest to add another test in `test_config_utils.py` to test that when nothing is set in for `target_modules` by our local peft_config.LoraConfig, HF peft.LoraConfig still has `target_modules` None set.
- comment in line to be more specific/readable for future maintenance.

willmj:
It seems like the [lora test](https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/tests/test_sft_trainer.py#L352) in `test_sft_trainer.py` are already testing default target_modules being None, with additional testing for specific configs. Should I still add the above test?

tests/build/test_launch_script.py
willmj: It seems like the [lora test](https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/tests/test_sft_trainer.py#L352) in test_sft_trainer.py are already testing default target_modules being None, with additional testing for specific configs. Should I still add the above test?

willmj:
It seems like the second test you mentioned to add is covered already in [test_get_hf_peft_config_returns_lora_config_correctly()](https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/tests/utils/test_config_utils.py#L41). Let me know if there's any additional functionality there that needs testing.
```
assert (
        config.target_modules is None
    )  # default value from local peft_config.LoraConfig
```

tests/build/test_launch_script.py
aluu317: I have a feeling that this test [linked](https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/tests/test_sft_trainer.py#L381) is bypassing the argument parsing from CLI or from JSON and just calling the `train` function directly with the None argument for target_modules. 

We want to test that even when user passes None or not passing anything from CLI or from JSON, it still works. Like the test you  wrote for this PR. 
(The problem with the test you wrote for this PR is it tests the accelerate launch script only, so someone who calls `tuning.sft_trainer` in cli will bypass your test as they're not calling the accelerate launch script. Hence I requested for the test to be moved to sft_trainer). 

I see that there's no testing of `main` in `test_sft_trainer`, this looks like it would be first!

aluu317 (approved):
This looks good thanks @willmj . I approved but please wait for Anh or Sukriti!

tests/test_sft_trainer.py
```diff
@@ -29,7 +29,9 @@
 import transformers
 
 # First Party
+from build.utils import serialize_args
 from scripts.run_inference import TunedCausalLM
+from tests.build.test_launch_script import BASE_LORA_KWARGS
 from tests.data import (
     EMPTY_DATA,
     MALFORMATTED_DATA,
```
anhuong, line 37: Would be good to add a test where run parse_arguments and verify that not passing in target_modules gives None still. This can be added to [existing test on line 149](https://github.com/foundation-model-stack/fms-hf-tuning/pull/281/files#diff-406c3c4d02c1ff158009b04d26a6b2d357f05d11e8c35c3f2ddfc1b54649a022R149)

tests/test_sft_trainer.py
```diff
@@ -403,6 +405,39 @@ def test_run_causallm_lora_and_inference(request, target_modules, expected):
         assert "Simply put, the theory of relativity states that" in output_inference
 
 
+def test_successful_lora_target_modules_default_from_main():
+    """Check that if target_modules is not set, or set to None in KWARGS sent to main,
+    the default value by model type will be using in training.
+    We use TRAIN_KWARGS from test_launch_script.py to run training with, which has no
+    target_modules set. During HF training process, the correct default target modules
+    will be used for model type Llama and "q_proj", "v_proj" will then exist in the
+    resulting in adapter_config.json.
```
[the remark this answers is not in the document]
willmj: Thanks for looking out! Yeah it's definitely confusing to pull from test_launch_script, I didn't want to add too much to test_sft_trainer for just one test, but that makes sense - I'll go ahead and change that.

> Good to be more specific, use LlamaForCausalLM

I actually think the [model type](https://huggingface.co/Maykeye/TinyLLama-v0/blob/main/config.json) (the one that [HF uses for defaulting](https://github.com/huggingface/peft/blob/7b1c08d2b5e13d3c99b7d6ee83eab90e1216d4ba/src/peft/utils/constants.py#L70)) I mean is llama, let me know what you think.

tests/test_sft_trainer.py
```diff
@@ -403,6 +446,46 @@ def test_run_causallm_lora_and_inference(request, target_modules, expected):
...
+    We use MODEL_ARGS to run tuning, which has no target_modules set. During HF
+    training process, the correct default target modules will be used for model
+    type llama and will then exist in the resulting in adapter_config.json.
+    To see how HF handles target_module defaults:
+    https://github.com/huggingface/peft/blob/7b1c08d2b5e13d3c99b7d6ee83eab90e1216d4ba/
+    src/peft/tuners/lora/model.py#L432
+    """
+    with tempfile.TemporaryDirectory() as tempdir:
+        TRAIN_KWARGS = {
+            **MODEL_ARGS.__dict__,
```
anhuong, line 461: Ahh I see since these aren't dicts, you can't combine them as easily. Nice way of combining them! Although a small refactor would be to combine your custom ones 
```py
**PEFT_LORA_ARGS.__dict__,
**{"peft_method": "lora", "output_dir": tempdir},
```

tests/test_sft_trainer.py
```diff
@@ -403,6 +446,46 @@ def test_run_causallm_lora_and_inference(request, target_modules, expected):
         assert "Simply put, the theory of relativity states that" in output_inference
 
 
+def test_successful_lora_target_modules_default_from_main():
+    """Check that if target_modules is not set, or set to None via JSON, the
+    default value by model type will be using in LoRA tuning.
+    We use MODEL_ARGS to run tuning, which has no target_modules set. During HF
+    training process, the correct default target modules will be used for model
+    type llama and will then exist in the resulting in adapter_config.json.
```
anhuong, line 454: nit: slight rewording for clarity and making a little more succinct
```suggestion
    """Check if target_modules is not set or set to None, the default
    target_modules by model type will be using in LoRA tuning.
    The correct default target modules will be used for model type
    LlamaForCausalLM and will exist in the resulting adapter_config.json.
```

tests/test_sft_trainer.py
```diff
@@ -403,6 +446,46 @@ def test_run_causallm_lora_and_inference(request, target_modules, expected):
...
+    with tempfile.TemporaryDirectory() as tempdir:
+        TRAIN_KWARGS = {
+            **MODEL_ARGS.__dict__,
+            **TRAIN_ARGS.__dict__,
+            **DATA_ARGS.__dict__,
+            **PEFT_LORA_ARGS.__dict__,
+            **{"peft_method": "lora"},
+            **{"output_dir": tempdir},
+        }
+        # TRAIN_KWARGS = {**BASE_LORA_KWARGS, **{"output_dir": tempdir}}
```
anhuong, line 468: Can remove comment

tests/test_sft_trainer.py
```diff
@@ -44,6 +45,47 @@
...
+        "peft_method": "pt",
+        "prompt_tuning_init": "RANDOM",
+        "num_virtual_tokens": 8,
+        "prompt_tuning_init_text": "hello",
+        "save_strategy": "epoch",
+        "output_dir": "tmp",
+    },
+}
+BASE_LORA_KWARGS = {
+    **BASE_KWARGS,
```
anhuong, line 81: This isn't used so can remove

anhuong (changes requested):

tests/test_sft_trainer.py
```diff
@@ -159,6 +201,7 @@ def test_parse_arguments_peft_method(job_config):
         parser, job_config_lora
     )
     assert isinstance(tune_config, peft_config.LoraConfig)
+    assert "target_modules" not in job_config_lora
```
anhuong, line 204: can also `assert not hassattr(tune_config, "target_modules")` so that it verifies the input from job_config_lora and this above assertion verifies that after parse_arguments run, the value is still None for target_modules

willmj: When I add this check I get the following error:
```
>       assert not hasattr(tune_config, "target_modules")
E       AssertionError: assert not True
E        +  where True = hasattr(LoraConfig(r=8, lora_alpha=32, target_modules=None, lora_dropout=0.05), 'target_modules')
```
I think this is because the attribute target_modules exists but is None. Would this suffice?
`assert tune_config.target_modules is None`

anhuong: yes that makes sense and yes that check looks good! i think you can refactor it down to `assert not tune_config.target_modules`

anhuong (approved):
LGTM, thanks Will!

PR #283, opened 2024-08-04
[Draft] Add dev stage to dockerfile

alex-jw-brooks:
### Was the PR tested

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -19,6 +19,7 @@ ARG USER=tuning
 ARG USER_UID=1000
 ARG PYTHON_VERSION=3.11
 ARG WHEEL_VERSION=""
+ARG ENABLE_AIM=false
 
 ## Base Layer ##################################################################
 FROM registry.access.redhat.com/ubi9/ubi:${BASE_UBI_IMAGE_TAG} as base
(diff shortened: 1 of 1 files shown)


build/Dockerfile
```diff
@@ -127,50 +132,95 @@ RUN if [[ -z "${WHEEL_VERSION}" ]]; \
...
+
+## Stages for release images ##################################
+FROM files-common as python-rel-installations
+ARG USER
+ARG USER_UID
+ARG ENABLE_AIM
+
 # Install from the wheel
+# TODO - probably a good idea to install most stuff in common and copy it out
+# in both this stage and dev.
```
anhuong, line 187: Or similarly, could the dev stage come after the python building stage? But the dev stage adds the additional tools and then release stage removes the unnecessary parts? I guess it may be hard to install the dev tools and selectively remove them...

build/Dockerfile
```diff
@@ -98,20 +92,31 @@ RUN dnf config-manager \
...
+## Python dep management / common files for dev & release #####
+FROM cuda-devel as files-common
 
 ARG WHEEL_VERSION
 ARG USER
 ARG USER_UID
-
-## Enable Aimstack if requested via ENABLE_AIM set to "true"
-ARG ENABLE_AIM=false
+ARG ENABLE_AIM
```
anhuong, line 101: Probably no longer need this build arg in this stage since its used in later stages unless we want to setup the aim_profile in this stage and copy it in other stages. But I think you'd still need the if block so doesn't save too much

Issue #217, opened 2024-06-27
bug: Pod runs out of ephemeral storage (disk) space because of the temporary directory.


HarikrishnanBalagopal:
# Overview
https://github.com/foundation-model-stack/fms-hf-tuning/blob/09496999edbd02d656ae2fef778b30c137afc433/build/accelerate_launch.py#L93

While running the image in a K8s/Openshift Pod, since the intermediate checkpoints are being written to a temporary folder, the Pod runs out of disk space.

# Steps

1. Start a pod with the https://github.com/foundation-model-stack/fms-hf-tuning/tree/main/build image and decent disk space. Example: `ephemeralStorage: "20Gi"` 
2. Let training run for 10 epochs with https://huggingface.co/ibm-granite/granite-7b-base model and twitter complaints dataset https://www.kaggle.com/datasets/thoughtvector/customer-support-on-twitter/data
3. At around Epoch 7 training crashes with "cannot write to file/disk space" error.

# Expected Behaviour

We should be able to run the training without running out of space.

# Proposed Fix
Let the user configure the temporary folder. That way we can use a folder inside a PVC so that we don't run out of disk space on the Pod.

anhuong:
PR that addresses issue merged in

PR #291, opened 2024-08-08
feat: add save_model_dir flag where final checkpoint saved

anhuong:
### Description of the change
- Add optional `save_model_dir` flag where final checkpoint can be saved to using `trainer.save_model()`
   - Note that this only saves the model, not the optimizer states
   - Adds minimal `save()` to sft_trainer
- `output_dir` is reserved for checkpoint saving and training logs. This param is still required to pass in even if no checkpoints are saved using `save_strategy="no"`
   - Note that with this update, the training logs will be streamed to output_dir
- Update accelerate_launch.py:
   - Remove tempdir which caused issues with ephemeral storage
   - Remove copy final checkpoint since this is now moved into sft_trainer
   - For lm_head removal, removes lm_head from `save_model_dir`if exists otherwise removes from final checkpoint

### Related issue number

#217 

### How to verify the PR

Tested:
1. `save_strategy=“no”` and `save_model_dir` to diff path than `output_dir` --> verified saves final model and does not save any checkpoints in ´output_dir` only logs

2. `save_total_limit=2` and `output_dir` set (aka`save_model_dir` not set) --> only checkpoints are saved with logs

3. `save_strategy=“no”` and `output_dir==save_model_dir` --> verified that logs and model saved to path

4. `save_strategy="epoch”` and  save_total_limit=2` and`output_dir==save_model_dir` -->  checkpoint dirs, model, and training logs are all written to path 

5. accelerate_launch.py
   - `save_total_limit=3` and `save_model_dir==output_dir` --> same as 4, checkpoints, training logs, and model outputted to path
   - `save_strategy="no"` and `save_model_dir==output_dir` --> same as 3, only model and logs outputted to path
   - `save_total_limit=1` and `output_dir` subdir of `save_model_dir` --> output_dir with checkpoints and logs inside of save_model_dir
   - `save_total_limit=1` and `save_model_dir` subdir of `output_dir` --> output_dir has checkpoints, logs, and dir with model

6. accelerate_launch: Finally I also verified that the lm_head removal continued to work as expected:
- save_total_limit=1, save_model_dir==output_dir, granite-3b-code-base --> verified that with lora and ft the model was saved to given path with lm_head removed but the checkpoint didn't have lm_head removed
- save_strategy="no", save_model_dir==output_dir, granite-3b-code-base --> verified lm_head removed and no addt checkpoints saved
- save_total_limit=1, output_dir, granite-3b-code-base  (aka no save_model_dir given) --> verified lm_head removed from final checkpoint

Ran vLLM inference on LoRA and fine tuned llama-13b-base model that was saved in separate and same dir. Fine tuning got good "no complaint" inference result, LoRA got poor results after tuning but marginal improvement from base model.

### Was the PR tested


anhuong:
Wanted to call out a note from the description: For lm_head removal, removes lm_head from `save_model_dir` if exists, otherwise removes from final checkpoint. Is this the behavior we want or do we only want to remove lm_head if `save_model_dir` passed? Not fully sure if the saving

tests/test_sft_trainer.py
```diff
@@ -444,10 +445,32 @@ def test_successful_lora_target_modules_default_from_main():
...
+
+        trainer = sft_trainer.train(MODEL_ARGS, DATA_ARGS, save_model_args, None)
+        logs_path = os.path.join(
+            tempdir, FileLoggingTrackerConfig.training_logs_filename
+        )
+        _validate_logfile(logs_path)
+        # validate that no checkpoints created
+        assert not any(x.startswith("checkpoint-") for x in os.listdir(tempdir))
+
+        sft_trainer.save(tempdir, trainer)
```
anhuong, line 471: This was the best way I could think of to test the `save()` without having to try to mock up the SFTTrainer. I had tried doing something like...
```py
    training_args = SFTConfig(**transformer_kwargs)
    trainer = SFTTrainer(
        model=MODEL_NAME,
        args=TRAIN_ARGS,
        train_dataset=TWITTER_COMPLAINTS_DATA,
    )
    with tempfile.TemporaryDirectory() as tempdir:
        sft_trainer.save(tempdir, trainer)
```

but this requires needing to preprocess the train args so it can be parsed into SFTConfig and requires data preprocessing to run otherwise hits error `AttributeError: 'str' object has no attribute 'column_names'` when trying to tokenize the dataset

README.md
```diff
@@ -225,6 +226,40 @@ tuning/sft_trainer.py \
...
+$ ls /tmp/same_dir
+
+added_tokens.json	model-00001-of-00006.safetensors  model-00006-of-00006.safetensors  tokenizer_config.json
+checkpoint-16		model-00002-of-00006.safetensors  model.safetensors.index.json	    training_args.bin
+checkpoint-20		model-00003-of-00006.safetensors  special_tokens_map.json	    training_logs.jsonl
+config.json		model-00004-of-00006.safetensors  tokenizer.json
+generation_config.json	model-00005-of-00006.safetensors  tokenizer.model
+```
+
+LoRA tuning example with `save_strategy="epoch”`, `save_total_limit=1`, and `output_dir==save_model_dir==/tmp/same_dir_lora`.
```
ashokponkumar, line 254: Should we instead show an example of save_model_dir being a subfolder of output_dir? That way the the logs, checkpoints and final model area easily segregable.

build/accelerate_launch.py
```diff
@@ -102,143 +97,111 @@ def main():
...
 
-        try:
-            last_checkpoint_dir = get_highest_checkpoint(tempdir)
-            last_checkpoint_path = os.path.join(tempdir, last_checkpoint_dir)
+    # remove lm_head from granite with llama arch models
+    try:
+        checkpoint_dir = job_config.get("save_model_dir")
+        if not checkpoint_dir:
+            checkpoint_dir = os.path.join(
+                output_dir, get_highest_checkpoint(output_dir)
```
ashokponkumar, line 127: Do we know if removing the lm_head is required while resuming a training, if so we should think about removing the lm_head from each intermediate checkpoint as it is being written.

ashokponkumar (changes requested):
Thanks Anh! It looks good. Just a few questions.

README.md
```diff
@@ -225,6 +226,40 @@ tuning/sft_trainer.py \
...
+
+A useful flag to set to limit the number of checkpoints saved is [`save_total_limit`](https://huggingface.co/docs/transformers/main_classes/trainer#transformers.TrainingArguments.save_total_limit). Older checkpoints are deleted from the `output_dir`. For example if `save_total_limit=1`, this will only save the last checkpoint. However, while tuning, two checkpoints will exist in `output_dir` for a short time as the new checkpoint is created and then the older one will be deleted.
+
+`save_model_dir` can optionally be set to save the final checkpoint using `SFTTrainer.save_model()`. This can be used in tandem with `save_strategy="no"` to only save the final checkpoint and not any intermediate checkpoints, which can help to save space.
+
+`save_model_dir` can be set to a different directory than `output_dir`. If set to the same directory, the final checkpoint, training logs, and any intermediate checkpoints will all be saved to the same directory as seen below.
+
+Fine tuning example with `save_strategy="epoch”`, `save_total_limit=2`, and `output_dir==save_model_dir==/tmp/same_dir`. Note the checkpoint directories as well as the `training_logs.jsonl`:
```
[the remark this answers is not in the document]
anhuong: Yes can do! I agree this is recommended behavior, but wanted to point out the edge case that may cause the most confusion.

build/accelerate_launch.py
anhuong: Removing lm_head is required for loading the tuned model with vLLM, but I don't think it's required for resuming training. Is there a way I could check resuming training? If you have a command to run with sft_trainer, that would be super helpful!

README.md
```diff
@@ -225,6 +226,40 @@ tuning/sft_trainer.py \
...
+
+Checkpoints are saved to the given `output_dir`, which is a required field. If `save_strategy="no"`, the `output_dir` will only contain the training logs with loss details.
+
+A useful flag to set to limit the number of checkpoints saved is [`save_total_limit`](https://huggingface.co/docs/transformers/main_classes/trainer#transformers.TrainingArguments.save_total_limit). Older checkpoints are deleted from the `output_dir`. For example if `save_total_limit=1`, this will only save the last checkpoint. However, while tuning, two checkpoints will exist in `output_dir` for a short time as the new checkpoint is created and then the older one will be deleted.
+
+`save_model_dir` can optionally be set to save the final checkpoint using `SFTTrainer.save_model()`. This can be used in tandem with `save_strategy="no"` to only save the final checkpoint and not any intermediate checkpoints, which can help to save space.
+
+`save_model_dir` can be set to a different directory than `output_dir`. If set to the same directory, the final checkpoint, training logs, and any intermediate checkpoints will all be saved to the same directory as seen below.
```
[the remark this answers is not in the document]
ashokponkumar: sure Anh.

README.md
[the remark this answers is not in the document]
ashokponkumar: sure.

build/accelerate_launch.py
ashokponkumar: I think we have to stop a run midway and then run the training again.

We have to ensure we use https://huggingface.co/docs/transformers/en/main_classes/trainer#transformers.Trainer.train.resume_from_checkpoint argument. Seems like we are not currently setting it. We should possibly set it. This will resume from the previously stopped last checkpoint.

Ssukriti: hi @ashokponkumar we have not heard of any requirement to remove lm_head to resume training. @Abhishek-TAMU will be working on resume training from checkpoint in follow up PR as per issue https://github.ibm.com/ai-foundation/watson-fm-stack-tracker/issues/1007 . Can we wait for him to add the change?

ypu can post any tips on the issue @ashokponkumar

tuning/config/configs.py
```diff
@@ -119,6 +119,13 @@ class TrainingArguments(transformers.TrainingArguments):
             'steps' (save is done every `save_steps`)"
         },
     )
+    save_model_dir: str = field(
+        default=None,
+        metadata={
+            "help": "Directory where final checkpoint will be saved to \
```
Ssukriti, line 125:
```suggestion
            "help": "Directory where final tuned model will be saved to \
```

README.md
```diff
@@ -225,6 +226,44 @@ tuning/sft_trainer.py \
...
+
+By default, [`save_strategy`](tuning/config/configs.py) is set to `"epoch"` in the TrainingArguments. This means that checkpoints will be saved on each epoch. This can also be set to `"steps"` to save on every `"save_steps"` or `"no"` to not save any checkpoints.
+
+Checkpoints are saved to the given `output_dir`, which is a required field. If `save_strategy="no"`, the `output_dir` will only contain the training logs with loss details.
+
+A useful flag to set to limit the number of checkpoints saved is [`save_total_limit`](https://huggingface.co/docs/transformers/main_classes/trainer#transformers.TrainingArguments.save_total_limit). Older checkpoints are deleted from the `output_dir`. For example if `save_total_limit=1`, this will only save the last checkpoint. However, while tuning, two checkpoints will exist in `output_dir` for a short time as the new checkpoint is created and then the older one will be deleted.
+
+`save_model_dir` can optionally be set to save the designated checkpoint using `SFTTrainer.save_model()`. This can be used in tandem with `save_strategy="no"` to only save the designated checkpoint and not any intermediate checkpoints, which can help to save space.
```
Ssukriti, line 239: I think there has been some confusion here. 

If the user sets a validation dataset and [`load_best_model_at_end`](https://huggingface.co/docs/transformers/en/main_classes/trainer#transformers.TrainingArguments.load_best_model_at_end), then the best checkpoint will be saved. If no additional flags are set, the final checkpoint will be saved.

This is for when save_total_limit = 1 , then you can. control best or last 


save_model()  is always for saving tuned model at end after training. By then any checkpoint information is lost. Can you confirm

build/accelerate_launch.py
ashokponkumar: Sure Sukriti.

ashokponkumar (dismissed):

README.md
```diff
@@ -225,6 +226,44 @@ tuning/sft_trainer.py \
 
 To summarize you can pick either python for single-GPU jobs or use accelerate launch for multi-GPU jobs. The following tuning techniques can be applied:
 
+### Tips on Parameters to Set
+
+#### Saving models
```
Ssukriti, line 231: This README is confusing and can be simplified. 

We can have. 2 subsections 

1. Save checkpoints while training

- save_strategy = set to epoch , can also be set to steps or no (as described)
- checkpoints are saved to output_dir 
- save_total_limit para

 load_best_model para applies when save_total_limit=1 , as it is a training argument . it does not apply to trainer.save_model as that is after saving. we can skip this para on load_best_model as users can look up TrainingArguments themselves. It si always described in URL you linked https://huggingface.co/docs/transformers/main_classes/trainer#transformers.TrainingArguments.save_total_limit

2. Save model after training - 
save_model_dir` can optionally be set to save the tuned model using `SFTTrainer.save_model()`. This can be used in tandem with `save_strategy="no"` to only save the final tuned  and not any intermediate checkpoints, which can help to save space.

ways you can use save_model_dir and more tips (collapse this section)

examples on expanding:

README.md
```diff
@@ -225,6 +226,44 @@ tuning/sft_trainer.py \
...
+
+A useful flag to set to limit the number of checkpoints saved is [`save_total_limit`](https://huggingface.co/docs/transformers/main_classes/trainer#transformers.TrainingArguments.save_total_limit). Older checkpoints are deleted from the `output_dir`. For example if `save_total_limit=1`, this will only save the last checkpoint. However, while tuning, two checkpoints will exist in `output_dir` for a short time as the new checkpoint is created and then the older one will be deleted.
+
+`save_model_dir` can optionally be set to save the designated checkpoint using `SFTTrainer.save_model()`. This can be used in tandem with `save_strategy="no"` to only save the designated checkpoint and not any intermediate checkpoints, which can help to save space.
+
+If the user sets a validation dataset and [`load_best_model_at_end`](https://huggingface.co/docs/transformers/en/main_classes/trainer#transformers.TrainingArguments.load_best_model_at_end), then the best checkpoint will be saved. If no additional flags are set, the final checkpoint will be saved.
```
Ssukriti, line 241: better to skip this section as described above, does not apply to save_model. It applies to save checkpoint only

Ssukriti (changes requested):
Have you verified saved model, using save() still infers on vLLM?

Comments on re-organizing README

anhuong:
> Have you verified saved model, using save() still infers on vLLM?

Yes this is described in the description of the PR above

>> Ran vLLM inference on LoRA and fine tuned llama-13b-base model that was saved in separate and same dir. Fine tuning got good "no complaint" inference result, LoRA got poor results after tuning but marginal improvement from base model.

README.md
anhuong: i ended up keeping it but moving it within the save_total_limits section and not in the save_model_dir section

Ssukriti (approved):
Nice work @anhuong !

PR #293, opened 2024-08-08
feat: Added additional events such as on_step_begin, on_optimizer_step, on_substep_end

seshapad:
### Description of the change

Some events like on_step_begin, on_optimizer_step, on_substep_end have been added to the trainer controller callback.

### How to verify the PR

By using the events in the trainer controller configuration as `trigger`. For example:
  ```
  controller_metrics:
    - name: trainer_state
      class: TrainingState
  controllers:
    - name: stop_on_training_loss_on_save
      triggers:
        - on_optimizer_step
      rule: trainer_state["epoch"] >= 0.5
      operations:
        - hfcontrols.should_training_stop
  ```

### Was the PR tested

### DIFF
--- a/tuning/trainercontroller/callback.py
+++ b/tuning/trainercontroller/callback.py
@@ -582,3 +582,45 @@ class TrainerControllerCallback(TrainerCallback):
         kwargs["state"] = state
         kwargs["control"] = control
         self._actions_on_event(event_name="on_save", **kwargs)
+
+    def on_step_begin(
+        self,
+        args: TrainingArguments,
+        state: TrainerState,
+        control: TrainerControl,
+        **kwargs,
+    ):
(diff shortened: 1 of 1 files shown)


ashokponkumar (approved):

ashokponkumar:
@seshapad Does this require a particular transformer version to work? And if so, is it compatible with the version used by fms-hf-tuning?

seshapad:
> @seshapad Does this require a particular transformer version to work? And if so, is it compatible with the version used by fms-hf-tuning?

@ashokponkumar Only `on_optimizer_step()` has been added in [4.42.0](https://github.com/huggingface/transformers/releases/tag/v4.42.0) release. The rest have been available even in older versions and have been around for 2 years.

PR #294, opened 2024-08-08
FIX: Metrics file epoch indexing starting from 0

Abhishek-TAMU:
### Description of the change

Training metrics are being written to a file by the fms-hf-tuning trainer. Currently the epochs in the metrics file starts from 1 and the change requested is to have epoch starting from 0-indexed. 

### Related issue number

#[1163](https://github.ibm.com/ai-foundation/watson-fm-stack-tracker/issues/1163)

### How to verify the PR

Running the test command below to tune a small model will result in output file `training_logs.jsonl`. Verify the file if the epoch is 0-indexed.

```
--model_name_or_path Maykeye/TinyLLama-v0  \
--training_data_path tests/data/twitter_complaints_small.json  \
--output_dir outputs/full-tuning  \
--num_train_epochs 5  \
--per_device_train_batch_size 2  \
--gradient_accumulation_steps 1  \
--learning_rate 1e-5  \
--use_flash_attn false \
--torch_dtype "float32"
```

### Was the PR tested


tuning/trackers/filelogging_tracker.py
```diff
@@ -56,7 +56,9 @@ def _track_loss(self, loss_key, log_name, log_file, logs, state):
             log_obj = {
                 "name": log_name,
                 "data": {
-                    "epoch": round(logs["epoch"], 2),
+                    "epoch": round((logs["epoch"] - 1), 2)
+                    if logs["epoch"] != 1.0
+                    else 0,
```
aluu317, line 61: For consistency, let's do "0.0" instead of "0".

aluu317 (changes requested):
Is it possible to add a unit test to check that the first log line is printing epoch 0.0 (and not 1.0)?

Also I put a comment in line, let's do `0.0` for consistency.

tuning/trackers/filelogging_tracker.py
Abhishek-TAMU: Okay Sure. I can remove the conversion from 0.0 to 0

Abhishek-TAMU: Also regarding writing unit case for the same, I can think of is to read the log file and check the first line of the log file. Does this sound good to you for a base logic of unit test case ?

aluu317: yea that sounds good!

tuning/trackers/filelogging_tracker.py
```diff
@@ -56,7 +56,9 @@ def _track_loss(self, loss_key, log_name, log_file, logs, state):
             log_obj = {
                 "name": log_name,
                 "data": {
-                    "epoch": round(logs["epoch"], 2),
+                    "epoch": round((logs["epoch"] - 1), 2)
```
anhuong: Also I know the issue was for this to specifically be in the training logs to start at 0 indexing, but the loss that is printed out will still start at 1 correct? Since this change would only affect the logging and not how tranformers is calculating and logging the epochs

Abhishek-TAMU: Yes. I didn't find any argument or function which makes the transformer epoch to start from 0 instead of 1. This change is just a change in log file, but transformers will still start from 1.

Abhishek-TAMU: Also, I do have asked Mynhardt Burger, regarding clarification, if they just expects changes in epoch-index in the logging file of fms-hf-tuning repo or does it expects to change the transformers calculations and working to start from epoch 0 ?
[Redhat issue](https://issues.redhat.com/browse/RHOAIENG-10792)

Ssukriti:
hi all, please hold off in this PR. If there is no way for transformers to actually print out from epoch 0, doing postprocessing is incorrect as 1 epoch was complete . we can index from 0 only if there is a way to print loss at epoch 0 -> no training, just calculates loss first

Abhishek-TAMU:
Yea it makes sense.

ashokponkumar:
Just noticed that this PR is merged. Just confirming, hope we are not changing the way we call the first finished epoch as 1. It will have a implication in a lot of places and will be inconsistent with how the community calls it.

Abhishek-TAMU:
This PR is not merged instead a different PR is merged where changes from there PR were there in local fork as a mistake but it was fixed before merging into fms-hf-tuning repo, though the commits and revert commits still remains there. So nothing related to this PR is merged. Sorry for any inconvenience.

kmehant:
@Abhishek-TAMU I think you working on your main branch for your features which is not a suggested way given all these issues that you could come across. You should always branch out from main and then work on your feature and raise a PR with it. 

FYA @Ssukriti 

cc: @ashokponkumar

PR #295, opened 2024-08-08
Rename all fixtures with correct .jsonl extension

willmj:
### Description of the change

Current fixtures are in jsonl format but are .json file types. Since we want to add json format functionality, should rename current fixtures to appropriate format.
Renamed jsonl fixtures with .json file types to .jsonl, updating  __init__.py with file type changes.
### Was the PR tested


anhuong (approved):
Thanks Will! In regards to the `malformatted_data.json` and the `empty_data.json`, I know that the fact that these are JSON files isn't the issue but the contents of the files are the issues. I was just wondering if its clearer to have these be JSONL or JSON files. Just a thought

willmj:
If we wanted to fully test both data types it might make sense to test malformatted data and empty data in both a .jsonl and .json. I think if we only did one though, I think leaving them as .json files would be fine.

anhuong:
>  I think if we only did one though, I think leaving them as .json files would be fine.
 I don't think we need two because the type of file is a separate test. These tests care about the contents of the files and I believe they are loading the datasets correctly, the issue isn't with the JSON vs JSONL file type

PR #264, opened 2024-07-25
feat: logging control operation

seshapad:
### Description of the change

Logging control operation to inject logs into the training loop.

### Related issue number

"Closes #179"

### How to verify the PR

Use the following trainer controller configuration to try out the logging control:
```
controller_metrics:
  - name: state
    class: TrainingState
operations:
  - name: logcontrolstep
    class: LogControl
    arguments:
      log_format: 'This is a test log format [{event_name}] => {tc_metrics[state]}'
      log_level: warning
controllers:
  - name: log-controller-step
    triggers:
      - on_step_end
    rule: 'True'
    operations:
      - logcontrolstep.should_log
```

### Was the PR tested

- [x] I have added >=1 unit test(s) for every new method I have added.
- [x] I have ensured all unit tests pass

PR #296, opened 2024-08-12
feat: Example log controller yaml with training state

seshapad:
### Description of the change

This PR introduces a more involved example for log controller which uses multiple trainer controller metrics.

### Related issue number

Closes #264 

### Was the PR tested

### DIFF
--- a/examples/trainercontroller_configs/logging_controller_with_rank.yaml
+++ b/examples/trainercontroller_configs/logging_controller_with_rank.yaml
@@ -1,27 +1,33 @@
 controller_metrics:
   - name: trainer_state
     class: TrainingState
+  - name: training_loss
+    class: Loss
+  - name: per_process_state
+    class: PerProcessState
 operations:
-  - name: logcontroleval
+  - name: logger_on_log
     class: LogControl
     arguments:
-      log_format: 'This is a test log format [{event_name}] => {metrics}'
-      log_level: WARNING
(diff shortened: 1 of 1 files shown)


examples/trainercontroller_configs/logging_controller_with_rank.yaml
```diff
@@ -0,0 +1,33 @@
...
+  - name: logger_on_log
+    class: LogControl
+    arguments:
+      log_format: 'Epoch: {trainer_state[epoch]:.0f}, Step: {trainer_state[global_step]}, Rank: {per_process_state[rank]}, loss = {training_loss}'
+      log_level: INFO
+  - name: logger_on_save
+    class: LogControl
+    arguments:
+      log_format: |
+        Saving model in huggingface format at samples_seen: {trainer_state[global_step]}
```
ashokponkumar, line 18:
```suggestion
        Saving model in huggingface format at samples_seen: {trainer_state[global_step]}
```

It should not be samples_seen, its on a step, isn't it?

ashokponkumar (changes requested):
nit

examples/trainercontroller_configs/logging_controller_with_rank.yaml
seshapad: Addressed.

ashokponkumar (approved):

PR #297, opened 2024-08-12
feat: Enable JSON dataset compatibility

willmj:
### Description of the change
While JSON documentation exists, there have been issues with loading JSON files, as most testing has been focused on JSONL files. To address this:

- Create JSON versions of the JSONL fixtures.
- Add JSON fixtures to `__init__.py` and imports
- Add JSON tests to functions leading to the following failures (added 31 total, from 167 -> 198 tests):
```
tests/acceleration/test_acceleration_dataclasses.py ....                                                                                                              [  2%]
tests/acceleration/test_acceleration_framework.py s.ssssssss                                                                                                          [  7%]
tests/build/test_launch_script.py .........                                                                                                                           [ 11%]
tests/build/test_utils.py .........                                                                                                                                   [ 16%]
tests/test_sft_trainer.py ..................................ss......F......                                                                                           [ 41%]
tests/trackers/test_aim_tracker.py sss                                                                                                                                [ 43%]
tests/trackers/test_file_logging_tracker.py ..                                                                                                                        [ 44%]
tests/trackers/test_tracker_api.py ..                                                                                                                                 [ 45%]
tests/trainercontroller/test_tuning_trainercontroller.py ......................                                                                                       [ 56%]
tests/utils/test_config_utils.py ...............                                                                                                                      [ 64%]
tests/utils/test_data_type_utils.py ...                                                                                                                               [ 65%]
tests/utils/test_data_utils.py ...                                                                                                                                    [ 67%]
tests/utils/test_embedding_resize.py .                                                                                                                                [ 68%]
tests/utils/test_evaluator.py .                                                                                                                                       [ 68%]
tests/utils/test_preprocessing_utils.py .........F.......FFFF..................................F.....                                                                 [100%]
```
- Modify `preprocessing_utils.py` to handle both JSONL and JSON files based on file extension. The main change is to change the `load_hf_dataset_from_jsonl_file` function to `load_hf_dataset_from_file`. When format_dataset is called, if the dataset is not pretokenized or single sequence, it calls this function. The problem was that it only supported JSONL files. By looking at the file extension to determine if it is JSON or JSONL, we can return either a JSON or JSONL object.

### Related issue

> While JSONL file format works, JSON file format is currently broken at least for a few kinds of data formats we support . [https://github.com/foundation-model-stack/fms-hf-tuning/tree/main?tab=readme-ov-file#1-json-formats-with-a-single-s[…]-use-for-masking-on-completion](https://github.com/foundation-model-stack/fms-hf-tuning/tree/main?tab=readme-ov-file#1-json-formats-with-a-single-sequence-and-a-specified-response_template-to-use-for-masking-on-completion)
>We have documented JSON, but it doesn't actually work. Investigate why this is broken and fix.

### How to verify the PR

Run unit tests added in `test_sft_trainer.py`, `test_preprocessing_utils.py`
If any new tests need to be added or removed, let me know!

### Was the PR tested


tuning/utils/preprocessing_utils.py
```diff
@@ -355,6 +354,14 @@ def load_hf_dataset_from_jsonl_file(
     if input_field_name == output_field_name:
         raise ValueError("Input field name and output field name should not match!")
 
+    file_extension = os.path.splitext(data_path)[-1].lower()
+    if file_extension == ".jsonl":
+        file_format = "jsonl"
+    elif file_extension == ".json":
+        file_format = "json"
+    else:
+        raise ValueError("Unsupported file format! Use 'json' or 'jsonl'.")
```
anhuong, line 363: This can be refactored to be at the bottom and do the if statements together:
```py
    file_extension = os.path.splitext(data_path)[-1].lower()
    if file_extension == ".jsonl":
        return Dataset.from_generator(get_jsonl_object)
    elif file_extension == ".json":
        return Dataset.from_generator(get_json_object)
    else:
        raise ValueError("Unsupported file format! Use 'json' or 'jsonl'.")
```

tuning/utils/preprocessing_utils.py
```diff
@@ -364,7 +371,18 @@ def get_jsonl_object():
                     output_field_name: data[output_field_name],
                 }
 
-    return Dataset.from_generator(get_jsonl_object)
+    def get_json_object():
```
anhuong, line 374: Potentially can refactor in to the same method....
```py
   def get_json_object():
        with open(data_path, "r", encoding="utf-8") as json_file:
            if file_extension == ".jsonl":
                data_stream = [json.loads(line) for line in jsonl_file]
            elif file_extension == ".json":
                data_stream = json.load(json_file)
            for data in data_stream:
                yield {
                    input_field_name: data[input_field_name],
                    output_field_name: data[output_field_name],
                }
```

is `file_extension` accessible in the method given its a function defined inside a function?

anhuong (changes requested):
Lots of great work Will! Appreciate the solid test coverage but I think we should think about which ones we need to verify JSON vs JSONL format. Perhaps we have too many...

tests/test_sft_trainer.py
```diff
@@ -824,14 +867,18 @@ def test_pretokenized_dataset_bad_args(dataset_text_field, response_template):
         data_args = copy.deepcopy(DATA_ARGS)
         data_args.dataset_text_field = dataset_text_field
         data_args.response_template = response_template
-        data_args.training_data_path = TWITTER_COMPLAINTS_DATA_INPUT_OUTPUT
+        data_args.training_data_path = validation_data_path
```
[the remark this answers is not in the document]
willmj: Yeah I would agree that it isn't needed here since it's more about the data than than the data type.

tuning/utils/preprocessing_utils.py
willmj: Absolutely - very nice!

tuning/utils/preprocessing_utils.py
willmj: Even nicer! `file_extension` is not accessible in the method, but it could be something like this:
```
    def get_json_object():
        with open(data_path, "r", encoding="utf-8") as json_file:
            file_extension = os.path.splitext(data_path)[-1].lower()
            if file_extension == ".jsonl":
                data_stream = (json.loads(line) for line in json_file)
            elif file_extension == ".json":
                data_stream = json.load(json_file)
            else:
                raise ValueError("Unsupported file format! Use 'json' or 'jsonl'.")
            for data in data_stream:
                yield {
                    input_field_name: data[input_field_name],
                    output_field_name: data[output_field_name],
                }

    return Dataset.from_generator(get_json_object)
```

willmj:
[Changes](https://github.com/foundation-model-stack/fms-hf-tuning/wiki/PiPY-Wheel-Release-Instructions/_compare/5bc96e5a3c3a270537dc92257e02e5eb0411fac8...2d50fff37e480af1609b630e33d657ab3a6de222)

tests/test_sft_trainer.py
```diff
@@ -311,27 +313,35 @@ def test_run_causallm_pt_invalid_train_params(param_name, param_val, exc_msg):
             sft_trainer.train(MODEL_ARGS, DATA_ARGS, invalid_params, PEFT_PT_ARGS)
 
 
-def test_run_causallm_pt_with_validation():
+@pytest.mark.parametrize(
+    "validation_data_path",
+    [TWITTER_COMPLAINTS_DATA_JSONL, TWITTER_COMPLAINTS_DATA_JSON],
+)
```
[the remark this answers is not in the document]
anhuong: Wow that is a minimal difference, good to know! Thanks for the info. Will still try to think about which tests are needed and most useful

anhuong (dismissed):
This looks great Will! Sorry for delay in re-review. Went through the list of tests and I think you could add back in a test in `test_preprocessing_utils.py` to `test_get_preprocessed_dataset()` to test JSON with method `get_preprocessed_dataset()` but I think the tests have great coverage.

anhuong:
I think the final thing you need to update is the docs, the main place i see is [here](https://github.com/foundation-model-stack/fms-hf-tuning?tab=readme-ov-file#2-jsonl-with-input-and-output-fields-no-response-template) since this references only using JSONL but now this accepts JSON

PR #299, opened 2024-08-14
release: merge set of changes for v1.2.0

willmj:
### Description of the change
v1.2.0 release

### Was the PR tested


anhuong:
@willmj I looked through all of the commits and they look good except for the most recent commit from you. This does not look to be signed, 
> Merge tag 'v1.2.0-rc.1' into release

Some of the PRs that were done via merge commit instead of squash commit also have lots of commits spread out which makes it harder to read. So we should do squash commits going forward.

anhuong (approved):

PR #300, opened 2024-08-14
feat: Add support for smoothly resuming training from a saved checkpoint

Abhishek-TAMU:
### Description of the change

In this PR, have added feature of resuming the tuning from the last saved checkpoint of last tuning using the flag 
`--output_dir` which stores all the checkpoints.

### Related issue number

#[1007](https://github.ibm.com/ai-foundation/watson-fm-stack-tracker/issues/1007)

### How to verify the PR

Resuming the tuning of model after initial tuning and verify the output checkpoints. 

### Was the PR tested


tuning/utils/config_utils.py
```diff
@@ -135,3 +135,32 @@ def txt_to_obj(txt):
     except UnicodeDecodeError:
         # Otherwise the bytes are a pickled python dictionary
         return pickle.loads(message_bytes)
+
+
+def get_last_checkpoint(train_args):
```
fabianlim, line 140: why is this logic needed? since it seems to be [duplicated from that inside HF trainer](https://github.com/huggingface/transformers/blob/8820fe8b8c4b9da94cf1e4761876f85c562e0efe/src/transformers/trainer.py#L1930)

anhuong: This is also doing similar work to what exists in [build/utils](https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/build/utils.py#L43-L63) so that or this method could be refactored to be used in both places

kmehant (changes requested):
How is this different from simply passing the flag https://github.com/huggingface/transformers/blob/8820fe8b8c4b9da94cf1e4761876f85c562e0efe/src/transformers/training_args.py#L663?

ashokponkumar:
Why not just set the [`resume_from_checkpoint`](https://github.com/huggingface/transformers/blob/984bc11b0882ff1e5b34ba717ea357e069ceced9/src/transformers/trainer.py#L1858) field to `True` and let HF trainer take care of it?

tuning/utils/config_utils.py
Abhishek-TAMU: That's correct. I missed that one. Now importing it from `transformers.trainer_utils import get_last_checkpoint`

kmehant: I think it would be good to reuse this flag to pass to train() function as is the way user provides, this would save us from too much of custom handling which limit other potential usecases.

https://github.com/huggingface/transformers/blob/8820fe8b8c4b9da94cf1e4761876f85c562e0efe/src/transformers/training_args.py#L663

Abhishek-TAMU: Just to get clarity @kmehant, does user passes the flag `resume_from_checkpoint` if users wants to resume from particular checkpoint ?
And in particular case where user doesn't pass the flag, do we assume that user doesn't intend on resuming the tuning and wants to start a new one, OR we still do need to check for latest checkpoint path and use it to resume tuning. 
Thanks!

tuning/sft_trainer.py
```diff
@@ -360,7 +361,11 @@ def train(
         for x in framework.get_callbacks_and_ready_for_train(model, accelerator):
             trainer.add_callback(x)
 
-    trainer.train()
+    # Check if tuning has to be resumed from last saved checkpoint
+    is_checkpoint_available = get_last_checkpoint(training_args.output_dir)
+    if is_checkpoint_available:
+        logger.info("Tuning resumes from the last checkpoint")
+    trainer.train(resume_from_checkpoint=is_checkpoint_available)
```
kmehant, line 368: `resume_from_checkpoint=True` gives same effect isn't it?

tuning/utils/config_utils.py
kmehant: > Just to get clarity @kmehant, does user passes the flag resume_from_checkpoint if users wants to resume from particular checkpoint ?

Yes, `--resume_from_checkpoint=true` if to start from last checkpoint or to a path if it has to be some arbitrary checkpoint. And then we pass this as is to the train()

> And in particular case where user doesn't pass the flag, do we assume that user doesn't intend on resuming the tuning and wants to start a new one, OR we still do need to check for latest checkpoint path and use it to resume tuning.

I have no particular suggestion on the default flow when user does not pass the flag 🙂

Abhishek-TAMU: >Yes, --resume_from_checkpoint=true if to start from last checkpoint or to a path if it has to be some arbitrary checkpoint. And then we pass this as is to the train()

Okay

tuning/sft_trainer.py
Abhishek-TAMU: `resume_from_checkpoint=True`, Yes I tested it. It gives the same effect if the `output_dir` has checkpoints available or else throws error.

Abhishek-TAMU: It would be helpful for me to know your requirements on default flow when user does not pass the flag as then could be able to finalize the code changes for review.

Abhishek-TAMU: @ashokponkumar Any comment/input on above ?

ashokponkumar: What error do we get when `resume_from_checkpoint` is set to True and there is no checkpoint? We can go with this fix for now, but it would be good to check in the HF Trainer, if they are okay with gracefully failing. Even if `resume_from_checkpoint` is set, if there is no checkpoint in the output dir, it should start like a normal run from scratch.

ashokponkumar: Also, if user actually provides a false explicitly we should honor it.

Abhishek-TAMU: >What error do we get when resume_from_checkpoint is set to True and there is no checkpoint

Error output and failing of code and exiting.
`ValueError: No valid checkpoint found in output directory (outputs/lora-tuning)`

Abhishek-TAMU: >Even if resume_from_checkpoint is set, if there is no checkpoint in the output dir, it should start like a normal run from scratch.

>Also, if user actually provides a false explicitly we should honor it.

Okay.

Abhishek-TAMU: Based on above comments, I have pushed few changes.
@kmehant Feel free to review it. 

Cases covered and tested:

1- **If user passes the flag `resume_from_checkpoint` with value** `False`: Tuning will start from scratch.

2- **If user passes the flag `resume_from_checkpoint` with value** `True`: if `output_dir` has checkpoints, then tuning will resume from last checkpoint, else tuning will start from scratch.

3- **If user passes the flag `resume_from_checkpoint` with value** `String: Entire Path of checkpoint-x`: The model would be loaded from `Path of checkpoint-x` and tuning will resume from the `checkpoint-x`.

4- **If user doesn't pass flag `resume_from_checkpoint`** : if `output_dir` has checkpoints, then tuning will resume from last checkpoint, else tuning will start from scratch.

kmehant:
Thanks @Abhishek-TAMU 

@ashokponkumar FYA https://github.com/foundation-model-stack/fms-hf-tuning/pull/300#issuecomment-2302860601

ashokponkumar:
Sounds good.

> Based on above comments, I have pushed few changes.
> @kmehant  Feel free to review it. 
> 
> Cases covered and tested:
> 
> 1- **If user passes the flag `resume_from_checkpoint` with value** `False`: Tuning will start from scratch.
> 
> 2- **If user passes the flag `resume_from_checkpoint` with value** `True`: if `output_dir` has checkpoints, then tuning will resume from last checkpoint, else tuning will start from scratch.
> 
> 3- **If user passes the flag `resume_from_checkpoint` with value** `String: Entire Path of checkpoint-x`: The model would be loaded from `Path of checkpoint-x` and tuning will resume from the `checkpoint-x`.
> 
> 4- **If user doesn't pass flag `resume_from_checkpoint`** : if `output_dir` has checkpoints, then tuning will resume from last checkpoint, else tuning will start from scratch.
> 
> 
> 
> 
>

kmehant:
@Abhishek-TAMU can we add simple unit tests for each of the cases you mentioned above? Thanks

Abhishek-TAMU:
>@Abhishek-TAMU can we add simple unit tests for each of the cases you mentioned above? Thanks

Sure! Have added the test cases and pushed the changes. Thanks!

tuning/sft_trainer.py
```diff
@@ -360,7 +361,24 @@ def train(
...
+    ):
+        resume_from_checkpoint = get_last_checkpoint(training_args.output_dir)
+    else:
+        # `training_args.resume_from_checkpoint` gives string values
+        # Check if flag is false OR flag has checkpoint value for resuming tuning
+        resume_from_checkpoint = (
+            training_args.resume_from_checkpoint
+            if training_args.resume_from_checkpoint.lower() != "false"
+            else False
+        )
```
ashokponkumar, line 379: Why not simplify this logic even further and move it into __post_init__ of training_args. 

We can check if `resume_from_checkpoint` is `None` or `True`, and if so, we check if `get_last_checkpoint(training_args.output_dir)` returns a non empty string. If it returns a non empty string, then we set `resume_from_checkpoint` to `True`, else we set `resume_from_checkpoint` to `False`.  In the current condition added, if `get_last_checkpoint` does not return a checkpoint, doesn't it get set to empty string, which is not what we want, isn't it?

We can leave every other case untouched, Isn't it? HF will already handle the other cases well, or is it not handling when `resume_from_checkpoint` is a string `false`?

ashokponkumar (commented):
one query..

tuning/sft_trainer.py
Abhishek-TAMU: >if `get_last_checkpoint` does not return a checkpoint, doesn't it get set to empty string, which is not what we want, isn't it?

`get_last_checkpoint` either returns a checkpoint string or None.

Abhishek-TAMU: > is it not handling when `resume_from_checkpoint` is a string false ?

When `resume_from_checkpoint` is a string false, `trainer()` needs argument `resume_from_checkpoint` value as boolean.

`trainer()` accepts argument `resume_from_checkpoint` value as boolean True/False, None (none_type), or string with path of checkpoint.

ashokponkumar: So we are adding support for string `true` and `false` additionally, even though HF Trainer does not support it?

ashokponkumar: What is the behaviour if we set the Training argument `resume_from_checkpoint` to be `True` or `False`? Isn't it handling it as a boolean `True` or `False` and behave accordingly?

Abhishek-TAMU: Training argument `resume_from_checkpoint` takes in the path to a folder with a valid checkpoint for the model, as mentioned [here](https://github.com/huggingface/transformers/blob/7591ca5bc583aadbdfe6b2fb36a12c874233323c/src/transformers/training_args.py#L665), and doesn't work with `boolean/String` value `True/False`

ashokponkumar: Then lets go ahead with this change. LGTM. We can refine later if required.

ashokponkumar (dismissed):

tuning/sft_trainer.py
```diff
@@ -360,7 +361,24 @@ def train(
         for x in framework.get_callbacks_and_ready_for_train(model, accelerator):
             trainer.add_callback(x)
 
-    trainer.train()
+    resume_from_checkpoint = None
+    # Check if resume flag is not passed (None), or if flag is true and
+    # output_dir has checkpoints then get last checkpoint from output_dir
+    if (
+        training_args.resume_from_checkpoint is None
```
anhuong, line 368: Is this saying that if resume_from_checkpoint is not set, then it would enable resume from checkpoint? That seems off...If a user doesn't pass anything in, why would it be enabled?

tuning/sft_trainer.py
```diff
@@ -360,7 +361,24 @@ def train(
...
+        or training_args.resume_from_checkpoint.lower() == "true"
+    ):
+        resume_from_checkpoint = get_last_checkpoint(training_args.output_dir)
+    else:
+        # `training_args.resume_from_checkpoint` gives string values
+        # Check if flag is false OR flag has checkpoint value for resuming tuning
+        resume_from_checkpoint = (
```
anhuong, line 375: Am I understanding the logic here correctly -- that `resume_from_training` can be passed in as a boolean value or as a path to a directory to resume training from? Would be good to provide docs on this if the behavior is different from how HF transformers handles the argument as I see HF transformers only accepts path to checkpoint
> resume_from_checkpoint (str, optional) — The path to a folder with a valid checkpoint for your model.

Abhishek-TAMU: Sure!

Mentioning `trainer.train()` [arguments](https://github.com/huggingface/transformers/blob/db70426854fe7850f2c5834d633aff637f14772e/src/transformers/trainer.py#L1901).

```
resume_from_checkpoint (`str` or `bool`, *optional*):
If a `str`, local path to a saved checkpoint as saved by a previous instance of [`Trainer`]. If a
`bool` and equals `True`, load the last checkpoint in *args.output_dir* as saved by a previous instance
of [`Trainer`]. If present, training will resume from the model/optimizer/scheduler states loaded here.
```

If the value of `resume_from_checkpoint` is passed `None` or `False` to `trainer.train()`, then it would behave similiar as not passing this flag to `trainer.train()`

tests/test_sft_trainer.py
```diff
@@ -79,6 +79,197 @@
...
+        final_training_logs = _get_training_logs(tempdir, epoch=1)
+        assert len(final_training_logs) == 2
+
+
+def test_resume_training_from_checkpoint_with_flag_checkpoint_path():
+    """
+    Test feature of resume training from checkpoint using resume_from_checkpoint flag
+    to ensure that when the value of flag is a checkpoint-x path, the tuning will resume
+    from that checkpoint. This test checks if total_flos of checkpoint-x has not changed
+    after resuming tuning, hence confirming that the tuning has started from checkpoint-x.
```
anhuong, line 210: Try to make this one more succinct, Also what is `total_flos`?

anhuong: Let's make this a LoRA test to test resuming training from diff tuning techniques

anhuong (changes requested):
Great changes and much appreciate the tests Abhishek! Have a few refactors that can be made to the tests to create cleaner code. Do you know how much time this adds to the tests by running tuning an additional 7 times?

tuning/sft_trainer.py
anhuong: Ah I was looking at the docs here - https://huggingface.co/docs/transformers/main_classes/trainer#transformers.TrainingArguments.resume_from_checkpoint

tuning/sft_trainer.py
```diff
@@ -360,7 +361,24 @@ def train(
         for x in framework.get_callbacks_and_ready_for_train(model, accelerator):
             trainer.add_callback(x)
 
-    trainer.train()
+    resume_from_checkpoint = None
+    # Check if resume flag is not passed (None), or if flag is true and
+    # output_dir has checkpoints then get last checkpoint from output_dir
+    if (
+        training_args.resume_from_checkpoint is None
+        or training_args.resume_from_checkpoint.lower() == "true"
```
anhuong, line 369: Since this does accept a boolean value, shouldn't this just check the bool value? A user isn't passing in a string "True" but the boolean "True" right?

tests/test_sft_trainer.py
```diff
@@ -79,6 +79,197 @@
...
+            trainer_state = json.load(f)
+    return trainer_state, last_checkpoint
+
+
+def _get_training_logs(dir_path, epoch=None):
+    data_list = []
+    with open(f"{dir_path}/training_logs.jsonl", "r", encoding="utf-8") as file:
+        for line in file:
+            json_data = json.loads(line)
+            data_list.append(json_data)
```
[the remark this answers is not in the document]
anhuong: Can be a future improvement

tuning/sft_trainer.py
Abhishek-TAMU: Yes, I can update the Readme. In this [readme](https://github.com/Abhishek-TAMU/fms-hf-tuning/blob/main/README.md), is there any specific section recommended, where I can mention above details for users (regarding info of `resume_from_checkpoint` flag) ?

tests/test_sft_trainer.py
```diff
@@ -79,6 +79,197 @@
...
+            entry["loss"]
+            for entry in init_trainer_state["log_history"]
+            if entry["epoch"] == 1.0
+        ][0]
+        final_loss_epoch = [
+            entry["loss"]
+            for entry in final_trainer_state["log_history"]
+            if entry["epoch"] == 1.0
+        ][0]
+        assert init_loss_epoch == final_loss_epoch
```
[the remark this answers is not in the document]
Abhishek-TAMU: It's definitely a good test to add in general to test tuning, but not sure how it is related to resuming from checkpoint functionality.

tests/test_sft_trainer.py
```diff
@@ -79,6 +79,197 @@
...
+        # Resume training with higher epoch and same output dir
+        train_args.num_train_epochs += 5
+        sft_trainer.train(MODEL_ARGS, DATA_ARGS, train_args, None)
+        _validate_training(tempdir)
+
+        # Get trainer state of latest checkpoint
+        final_trainer_state, _ = _get_latest_checkpoint_trainer_state(tempdir)
+        assert final_trainer_state is not None
+
+        assert final_trainer_state["epoch"] > init_trainer_state["epoch"]
```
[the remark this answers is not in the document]
Abhishek-TAMU: Yes! This is addressed!

tests/test_sft_trainer.py
```diff
@@ -79,6 +79,197 @@
...
+        # Get total_flos from trainer state of checkpoint_path and check if its same
+        final_trainer_state = None
+        trainer_state_file = os.path.join(checkpoint_path, "trainer_state.json")
+        with open(trainer_state_file, "r", encoding="utf-8") as f:
+            final_trainer_state = json.load(f)
+
+        assert final_trainer_state["total_flos"] == init_trainer_state["total_flos"]
+
+
+def _get_latest_checkpoint_trainer_state(dir_path, checkpoint_index=-1):
```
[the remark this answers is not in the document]
Abhishek-TAMU: Added the description!

tests/test_sft_trainer.py
```diff
@@ -79,6 +79,197 @@
...
+        last_checkpoint = sorted(checkpoints, key=lambda x: int(x.split("-")[-1]))[
+            checkpoint_index
+        ]
+        trainer_state_file = os.path.join(last_checkpoint, "trainer_state.json")
+        with open(trainer_state_file, "r", encoding="utf-8") as f:
+            trainer_state = json.load(f)
+    return trainer_state, last_checkpoint
+
+
+def _get_training_logs(dir_path, epoch=None):
```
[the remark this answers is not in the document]
Abhishek-TAMU: Added the description!

tests/test_sft_trainer.py
```diff
@@ -79,6 +79,197 @@
...
+    data_list = []
+    with open(f"{dir_path}/training_logs.jsonl", "r", encoding="utf-8") as file:
+        for line in file:
+            json_data = json.loads(line)
+            data_list.append(json_data)
+
+    if epoch is not None:
+        mod_data_list = []
+        for value in data_list:
+            if int(value["data"]["epoch"]) == int(epoch):
```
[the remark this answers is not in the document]
Abhishek-TAMU: Yes this change is helpful. Thanks!
Also I guess the apt function name could be `_get_training_logs_by_epoch`. Do you agree with this ?

tests/test_sft_trainer.py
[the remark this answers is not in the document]
Abhishek-TAMU: Good point. Can look into this one as a future improvement.

tests/test_sft_trainer.py
```diff
@@ -79,6 +79,197 @@
...
+
+
+def _get_training_logs(dir_path, epoch=None):
+    data_list = []
+    with open(f"{dir_path}/training_logs.jsonl", "r", encoding="utf-8") as file:
+        for line in file:
+            json_data = json.loads(line)
+            data_list.append(json_data)
+
+    if epoch is not None:
```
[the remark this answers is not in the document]
Abhishek-TAMU: Thank you for reviewing test cases. This is done!

tests/test_sft_trainer.py
```diff
@@ -79,6 +79,197 @@
...
+        assert final_trainer_state["epoch"] > init_trainer_state["epoch"]
+        assert final_trainer_state["global_step"] > init_trainer_state["global_step"]
+
+        # Check if loss of 1st epoch after first tuning is same after
+        # resuming tuning and not overwritten
+        assert len(init_trainer_state["log_history"]) > 0
+        init_loss_epoch = [
+            entry["loss"]
+            for entry in init_trainer_state["log_history"]
+            if entry["epoch"] == 1.0
```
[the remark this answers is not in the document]
Abhishek-TAMU: Yes I checked it and the first one would always be epoch 1. And taking in your refactor, I can just modify it to something like below: (as I was taking out the loss in the code, but now just taking log history)

```
init_log_history = init_trainer_state["log_history"][0]
assert init_log_history["epoch"] == 1

final_log_history = final_trainer_state["log_history"][0]
assert final_log_history["epoch"] == 1

assert init_log_history["loss"] == final_log_history["loss"]
```

tests/test_sft_trainer.py
```diff
@@ -79,6 +79,197 @@
 PEFT_LORA_ARGS = peft_config.LoraConfig(r=8, lora_alpha=32, lora_dropout=0.05)
 
 
+def test_resume_training_from_checkpoint():
+    """
+    Test feature of resume training from checkpoint to ensure that the tuning resumes
+    from the latest checkpoint creating new checkpoints and the checkpoints created
+    before resuming tuning is not afected. Also the tuning is resumed with increased
+    number of epoch and ensures that the loss of 1st epoch is unchanged.
```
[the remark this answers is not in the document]
Abhishek-TAMU: Make sense. Thanks!

tests/test_sft_trainer.py
```diff
@@ -79,6 +79,197 @@
...
+    trainer_state = None
+    checkpoints = [
+        os.path.join(dir_path, d)
+        for d in os.listdir(dir_path)
+        if d.startswith("checkpoint")
+    ]
+    if checkpoints:
+        last_checkpoint = sorted(checkpoints, key=lambda x: int(x.split("-")[-1]))[
+            checkpoint_index
+        ]
```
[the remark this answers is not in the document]
Abhishek-TAMU: In line 248-250, we don't necessarily take out last checkpoint, instead get checkpoint based on `checkpoint_index`, so I guess `transformers.utils.get_last_checkpoint()` may not be useful here.

tests/test_sft_trainer.py
```diff
@@ -79,6 +79,197 @@
...
+
+def test_resume_training_from_checkpoint_with_flag_true():
+    """
+    Test feature of resume training from checkpoint using resume_from_checkpoint flag
+    to ensure that when the value of flag is True and output_dir has checkpoints, the
+    tuning resumes from the latest checkpoint creating new checkpoints and the checkpoints
+    created before resuming tuning is not affected. When the value of flag is True and
+    output_dir does not have checkpoints, then the tuning will start from scratch.
+    This test also checks if timestamp of 1st epoch after first tuning is same after
+    resuming tuning and hence confirming that earlier checkpoints are not overwritten.
```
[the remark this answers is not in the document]
Abhishek-TAMU: Sure! Have removed the description.

tests/test_sft_trainer.py
```diff
@@ -79,6 +79,197 @@
...
+            == final_training_logs[0]["data"]["timestamp"]
+        )
+
+
+def test_resume_training_from_checkpoint_with_flag_false():
+    """
+    Test feature of resume training from checkpoint using resume_from_checkpoint flag
+    to ensure that when the value of flag is False the tuning will start from scratch.
+    This test also checks multiple entry of 1st epoch in training logs after resuming
+    tuning and hence confirming that the tuning has started from scratch.
```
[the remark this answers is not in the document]
Abhishek-TAMU: Okay got it. Thanks!

tests/test_sft_trainer.py
```diff
@@ -79,6 +79,197 @@
...
+
+        # Get trainer state of latest checkpoint
+        init_trainer_state, _ = _get_latest_checkpoint_trainer_state(tempdir)
+        assert init_trainer_state is not None
+
+        # Get Training log entry for epoch 1
+        init_training_logs = _get_training_logs(tempdir, epoch=1)
+        assert len(init_training_logs) == 1
+
+        # Resume training with higher epoch and same output dir
```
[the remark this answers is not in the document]
Abhishek-TAMU: Yes correct, that's the logic. Also comment seems misleading, hence changed it to:

`# Training again with higher epoch and same output dir`

tests/test_sft_trainer.py
Abhishek-TAMU: `total_flos` is basically Total Floating Point Operations (FLOPs).
Formal def:
`total_flos` is a measure of the computational workload performed by the model during training. It is computed as the cumulative sum of FLOPs (Floating Point Operations) for each training step.

Abhishek-TAMU: > Let's make this a LoRA test to test resuming training from diff tuning techniques

Sure! Made this test as LoRA test. Thanks!

Abhishek-TAMU:
> Great changes and much appreciate the tests Abhishek! Have a few refactors that can be made to the tests to create cleaner code. Do you know how much time this adds to the tests by running tuning an additional 7 times?

Thank you @anhuong for the review on test cases. I have made changes mentioned and pushed it. Also I calculated the time these 4 test cases adds to the tests. Difference in running with and without the 4 test case is 15 seconds. Thanks!

tuning/sft_trainer.py
Abhishek-TAMU: Yes, `resume_from_checkpoint` does accepts bool in `trainer.train()`. Its just that, when user passes `resume_from_checkpoint` flag as bool, it gets taken as string as it gets treated as flag in [training arguments](https://huggingface.co/docs/transformers/main_classes/trainer#transformers.TrainingArguments.resume_from_checkpoint) which just accepts string.
image

tests/test_sft_trainer.py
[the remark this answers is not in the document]
anhuong: i could go either way on this, its only getting by epoch if the user provides an epoch, otherwise it returns the full training logs. so its not necessarily by epoch but is fine either way to me

tests/test_sft_trainer.py
```diff
@@ -238,8 +221,27 @@ def test_resume_training_from_checkpoint_with_flag_checkpoint_path():
...
 
 
-def _get_latest_checkpoint_trainer_state(dir_path, checkpoint_index=-1):
+def _get_latest_checkpoint_trainer_state(dir_path: str, checkpoint_index: int = -1):
+    """
+    Get the trainer state from the specified checkpoint directory.
+    This function gets the latest or specific checkpoint based on the
+    provided checkpoint_index from the checkpoint directory, and loads
+    the `trainer_state.json` file from that checkpoint. The trainer
+    state is returned along with the path to the checkpoint.
```
anhuong, line 230: nit: kinda repetitive so makes more succinct
```suggestion
    Get the trainer state from the latest or specified checkpoint directory.
    The trainer state is returned along with the path to the checkpoint.
```

tests/test_sft_trainer.py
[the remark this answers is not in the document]
anhuong: true

tests/test_sft_trainer.py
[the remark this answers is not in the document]
anhuong: Not the description, I was referring to the test. All tests should have descriptions but this test looks to be doing the same work as the one above

tests/test_sft_trainer.py
```diff
@@ -190,31 +174,30 @@ def test_resume_training_from_checkpoint_with_flag_false():
...
 
 
 def test_resume_training_from_checkpoint_with_flag_checkpoint_path():
     """
-    Test feature of resume training from checkpoint using resume_from_checkpoint flag
-    to ensure that when the value of flag is a checkpoint-x path, the tuning will resume
-    from that checkpoint. This test checks if total_flos of checkpoint-x has not changed
-    after resuming tuning, hence confirming that the tuning has started from checkpoint-x.
+    Test when setting resume_from_checkpoint=path/to/checkpoint-x
+    that the tuning will resume from the checkpoint-x.
```
anhuong, line 193: nit: Could be a little clearer and should update that this is for LoRA
```suggestion
    Test resume checkpoint from a specified checkpoint path for LoRA tuning.
```

tests/test_sft_trainer.py
```diff
@@ -190,31 +174,30 @@ def test_resume_training_from_checkpoint_with_flag_false():
...
         sft_trainer.train(MODEL_ARGS, DATA_ARGS, train_args, None)
         _validate_training(tempdir)
 
         # Get Training log entry for epoch 1
-        final_training_logs = _get_training_logs(tempdir, epoch=1)
+        final_training_logs = _get_training_logs_by_epoch(tempdir, epoch=1)
         assert len(final_training_logs) == 2
 
 
 def test_resume_training_from_checkpoint_with_flag_checkpoint_path():
```
anhuong, line 190: SHould include LoRA here
```suggestion
def test_resume_training_from_checkpoint_with_flag_checkpoint_path_lora():
```

tuning/sft_trainer.py
anhuong: ahh thanks for the clarity

anhuong (commented):
Not sure why all of my comments got split up in different reviews, sorry about that. There are a few nits on the descriptions and small changes that can be addressed as well as a test case that I'm not sure if its useful. In addition, a quick note should be added to docs that explains how this works but point at the HF.TrainingArguments / Trainer docs

anhuong:
In terms of the linting and too many lines in test_sft_trainer.py, I think its okay to increase this value in pylint. We can create an issue to look into the tests and see how we can split them up into separate files. For now, lets increase to 1200

tests/test_sft_trainer.py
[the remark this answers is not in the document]
Abhishek-TAMU: Okay got it. Yes this test also checks if tuning is resuming from checkpoint but here I am marking `resume_from_checkpoint` flag as true, as to test the logic in the `sft_trainer` when flag value is True.

tests/test_sft_trainer.py
```diff
@@ -255,17 +257,32 @@ def _get_latest_checkpoint_trainer_state(dir_path, checkpoint_index=-1):
     return trainer_state, last_checkpoint
 
 
-def _get_training_logs(dir_path, epoch=None):
+def _get_training_logs_by_epoch(dir_path: str, epoch: int = None):
+    """
+    Load and optionally filter training logs from a training_logs JSON Lines file.
+    This function reads a JSON Lines (`.jsonl`) file containing training logs and
```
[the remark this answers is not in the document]
Abhishek-TAMU: Thank you!

tests/test_sft_trainer.py
Abhishek-TAMU: Thank you!

tests/test_sft_trainer.py
Abhishek-TAMU: Thank you!

Abhishek-TAMU:
Have pushed changes related to function description, adding ReadMe and increase in line value in pylint. 
Feel free to review once again. @anhuong 
Thanks!

tests/test_sft_trainer.py
```diff
@@ -122,6 +122,11 @@ def test_resume_training_from_checkpoint():
 
 
 def test_resume_training_from_checkpoint_with_flag_true():
```
anhuong, line 124: I'm not sure if this method is useful or if it;s doing the same as the one above.

anhuong (approved):
Other than question on test, change looks good. Please answer question but i'll merge this as-is

anhuong:
@kmehant Since you reviewed this PR and requested changes, the PR is waiting for your approval before it can be merged

tests/test_sft_trainer.py
Abhishek-TAMU: Yes, this test also checks if tuning is resuming from checkpoint but here I am marking `resume_from_checkpoint` flag as true, as to test the logic in the `sft_trainer.py` when flag value is True, though it's behavior would be similar to when value of `resume_from_checkpoint` is `None`.

Let me know, if you think we should remove this test.

kmehant (approved):
@Abhishek-TAMU lets rebase thank you.

tests/test_sft_trainer.py
anhuong: It's a bit redundant as resume_from_checkpoint=None creates the same behavior as resume_from_checkpoint=True but fine to leave for now

PR #302, opened 2024-08-15
fix: list type flags like LoRA --target_modules where not being handled properly by the fire CLI argument parser

HarikrishnanBalagopal:
### Was the PR tested


HarikrishnanBalagopal:
remove `fire` instead.

PR #304, opened 2024-08-15
fix: setting log level in save()

anhuong:
### Description of the change

Bug where log level set in `save()` fails if lowercase values passed in for log_level. This fixes the issue and adds a small addition to unit tests to ensure this change fixes it.

### Was the PR tested


Abhishek-TAMU:
Looks good to me!

aluu317 (dismissed):
LGTM

aluu317 (commented):
It'd be nice to add "log_level" to the doc string of `save( )` function since this might have been missed in previous PR. But could be done in your next refactor PR!

aluu317 (approved):

PR #305, opened 2024-08-16
doc: How to tune LoRA lm_head

aluu317:
### Description of the change

We need to document on how to LoRA tune `lm_head` layer

### Was the PR tested


README.md
```diff
@@ -387,7 +390,34 @@ For example for LLaMA model the modules look like:
...
+As per [LoRA paper](https://arxiv.org/pdf/2106.09685), section 4.2 , by using the query and value projection matrices, we can achieve reasonable quality with efficient GPU utilization. Hence, while thinking about what LoRA adapters to specify, we recommend starting with query and value matrices. You could also refer to the defaults specified by PEFT library for popular model architectures in section [TRANSFORMERS_MODELS_TO_LORA_TARGET_MODULES_MAPPING](https://github.com/huggingface/peft/blob/7b1c08d2b5e13d3c99b7d6ee83eab90e1216d4ba/src/peft/utils/constants.py#L70) as a good starting point.
+
+#### How to specify lm_head as a target module
```
Ssukriti, line 395: could you add this to a collapse section https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/organizing-information-with-collapsed-sections 

as its advanced usage to use lm_head and wasnt supported by vLLM anyway for inference

aluu317: good idea! Will do

README.md
```diff
@@ -331,8 +331,11 @@ Equally you can pass in a JSON configuration for running tuning. See [build doc]
 }
 ```
 
-Notice the `target_modules` that are set are the default values. `target_modules` are the names of the modules to apply the adapter to. If this is specified, only the modules with the specified names will be replaced. When passing a list of strings, either an exact match will be performed or it is checked if the name of the module ends with any of the passed strings. If this is specified as `all-linear`, then all linear/Conv1D modules are chosen, excluding the output layer. If this is not specified, modules will be chosen according to the model architecture. If the architecture is not known, an error will be raised — in this case, you should specify the target modules manually. See [HuggingFace docs](https://huggingface.co/docs/peft/en/package_reference/lora#peft.LoraConfig) for more details.
+Notice the `target_modules` are the names of the modules to apply the adapter to. See below [section]() for how to specify `lm_head` as a target module.
```
anhuong, line 334: `[section]()` is missing the section link

anhuong: I also think the note on lm_head should be in the first bullet to show all the different options - specify your own modules, all-linear, and then note that specifying lm_head is a special case

README.md
```diff
@@ -387,7 +390,38 @@ For example for LLaMA model the modules look like:
...
+As per [LoRA paper](https://arxiv.org/pdf/2106.09685), section 4.2 , by using the query and value projection matrices, we can achieve reasonable quality with efficient GPU utilization. Hence, while thinking about what LoRA adapters to specify, we recommend starting with query and value matrices. You could also refer to the defaults specified by PEFT library for popular model architectures in section [TRANSFORMERS_MODELS_TO_LORA_TARGET_MODULES_MAPPING](https://github.com/huggingface/peft/blob/7b1c08d2b5e13d3c99b7d6ee83eab90e1216d4ba/src/peft/utils/constants.py#L70) as a good starting point.
+
+<details>
+
+<summary>How to specify lm_head as a target module</summary>
+
+Since `lm_head` is an output layer, it will _not_ be included as target module if you specify `all-linear`. You can, however, specify to apply the LoRA adapter to `lm_head` layer by explicit naming it in the `target_modules` arg.
```
anhuong, line 399: small fixes:
```suggestion
Since `lm_head` is an output layer, it will **not** be included as a target module if you specify `all-linear`. You can, however, specify to apply the LoRA adapter to the `lm_head` layer by explicitly naming it in the `target_modules` arg.
```

README.md
```diff
@@ -387,7 +390,38 @@ For example for LLaMA model the modules look like:
...
+
+*NB*: Specifying `["lm_head", "all-linear"]` will not produce `lm_head` layer, but only the equivalent of running with `["all-linear"]`. Explicitly specify the layers needed. Using the example of Llama model above, you would need to list `"q_proj" "v_proj" "k_proj" "o_proj" "lm_head"` and these 5 layers will be produced with applied LoRA adapter.
```
anhuong, line 401: few nits
- I hadn't heard of NB and had to look it up to see what it meant 😅  might be easier for users to use something like "NOTE:"
- small rewording suggestion for clarity
```suggestion
**NOTE**: Specifying `["lm_head", "all-linear"]` will not tune the `lm_head` layer, but will run the equivalent of `["all-linear"]`. To include `lm_head`, you must explicitly specify all of the layers to tune on. Using the example of the Llama model above, you would need to list `"q_proj" "v_proj" "k_proj" "o_proj" "lm_head"` to tune the linear layers and `lm_head`. These 5 layers will be produced in the LoRA adapter.
```

anhuong (approved):
Thanks Angel for the doc update! Very useful for users

Issue #307, opened 2024-08-16
bug: UnboundLocalError, `logger` is not defined


HarikrishnanBalagopal:
# Bug

## Overview

Commit `a6d093eb96e62311ffa02cdeb97d78c90fc86e74`

This bug happens if there is an exception in `parse_arguments` function call because the logger is only bound after it https://github.com/foundation-model-stack/fms-hf-tuning/blob/a6d093eb96e62311ffa02cdeb97d78c90fc86e74/tuning/sft_trainer.py#L530

So the `logger` variable here is unbound
https://github.com/foundation-model-stack/fms-hf-tuning/blob/a6d093eb96e62311ffa02cdeb97d78c90fc86e74/tuning/sft_trainer.py#L550

## Steps to reproduce

```
$ python3 -m tuning.sft_trainer --output_dir foo --use_flash_attn false --bnb_qlora true
Traceback (most recent call last):
  File "/Users/haribala/Documents/code/remote/github.com/foundation-model-stack/fms-hf-tuning/tuning/sft_trainer.py", line 527, in main
    ) = parse_arguments(parser, job_config)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/haribala/Documents/code/remote/github.com/foundation-model-stack/fms-hf-tuning/tuning/sft_trainer.py", line 485, in parse_arguments
    ) = parser.parse_args_into_dataclasses(return_remaining_strings=True)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/haribala/Documents/code/remote/github.com/foundation-model-stack/fms-hf-tuning/.venv/lib/python3.11/site-packages/transformers/hf_argparser.py", line 339, in parse_args_into_dataclasses
    obj = dtype(**inputs)
          ^^^^^^^^^^^^^^^
  File "<string>", line 5, in __init__
  File "/Users/haribala/Documents/code/remote/github.com/foundation-model-stack/fms-hf-tuning/tuning/config/acceleration_configs/quantized_lora_config.py", line 72, in __post_init__
    ensure_nested_dataclasses_initialized(self)
  File "/Users/haribala/Documents/code/remote/github.com/foundation-model-stack/fms-hf-tuning/tuning/config/acceleration_configs/utils.py", line 34, in ensure_nested_dataclasses_initialized
[...]

PR #308, opened 2024-08-16
fix: logger is unbound error

HarikrishnanBalagopal:
### Description of the change

Fixes https://github.com/foundation-model-stack/fms-hf-tuning/issues/307

### Related issue number

https://github.com/foundation-model-stack/fms-hf-tuning/issues/307

### How to verify the PR

```
python3 -m tuning.sft_trainer --output_dir foo --use_flash_attn false --bnb_qlora true
```

### Was the PR tested


Ssukriti:
@aluu317 could you review this as you recently went over logging code?

anhuong (approved):
Need to initialize logger so it can be used in the exception if parse_argumetns fails

ashokponkumar:
Thanks @anhuong !

Issue #194, opened 2024-06-12
feat: use native python logger instead of transformers logger


kmehant:
Move away from using logger modules from transformers as it was strictly meant to be used (https://github.com/huggingface/transformers/blob/1c73d85b86e8afa549eb1209b3e62afdaa664cad/src/transformers/utils/logging.py#L151) when writing a submodule within transformers module. Use python native logger instead. 

Also, setting levels on this logger does not seem to work and is quite buggy. It only works when we get logger with name `transformers`. Setting verbosity as well does not set the level for the logger which has been created using `get_logger()`. We should instead use python native logger and for transformer specific logging, we should simply set its log verbosity level but not use it as a logger in our code (https://huggingface.co/docs/transformers/en/main_classes/logging#transformers.utils.logging.set_verbosity_error)

PR #309, opened 2024-08-19
feat: Migrating the trainer controller to python logger

seshapad:
### Description of the change

This PR switches the logging module from tranformer logger to python logger.

### Related issue number

Closes https://github.com/foundation-model-stack/fms-hf-tuning/issues/194

### Was the PR tested

### DIFF
--- a/tuning/trainercontroller/callback.py
+++ b/tuning/trainercontroller/callback.py
@@ -20,6 +20,7 @@ from typing import Dict, List, Union
 import inspect
 import os
 import re
+import logging
 
 # Third Party
 from simpleeval import EvalWithCompoundTypes, FeatureNotAvailable, NameNotDefined
@@ -29,7 +30,6 @@ from transformers import (
     TrainerState,
     TrainingArguments,
 )
-from transformers.utils import logging
 import yaml
 
 # Local
@@ -45,7 +45,7 @@ from tuning.trainercontroller.operations import (
 from tuning.trainercontroller.patience import PatienceControl
 from tuning.utils.evaluator import MetricUnavailableError, RuleEvaluator
 
-logger = logging.get_logger(__name__)
+logger = logging.getLogger()
 
 # Configuration keys
 CONTROLLER_METRICS_KEY = "controller_metrics"
--- a/tuning/trainercontroller/callback.py
+++ b/tuning/trainercontroller/callback.py
@@ -305,7 +305,7 @@ class TrainerControllerCallback(TrainerCallback):
         kwargs["state"] = state
         kwargs["control"] = control
 
-        log_levels = logging._levelToName
+        log_levels = dict((v,k) for k,v in logging._levelToName.items())
         # Check if there any metrics listed in the configuration
         if (
             CONTROLLER_METRICS_KEY not in self.trainer_controller_config
--- a/tuning/trainercontroller/callback.py
+++ b/tuning/trainercontroller/callback.py
@@ -46,6 +46,12 @@ from tuning.trainercontroller.patience import PatienceControl
 from tuning.utils.evaluator import MetricUnavailableError, RuleEvaluator
 
 logger = logging.getLogger(__name__)
+logger.setLevel(logging.DEBUG)
+
+logging.getLogger('hfcontrols').setLevel(logging.DEBUG)
+logging.getLogger('logcontrol').setLevel(logging.DEBUG)
+logging.getLogger('operation').setLevel(logging.DEBUG)
+logging.getLogger('patience').setLevel(logging.DEBUG)
 
 # Configuration keys
 CONTROLLER_METRICS_KEY = "controller_metrics"
--- a/tuning/trainercontroller/callback.py
+++ b/tuning/trainercontroller/callback.py
@@ -48,11 +48,6 @@ from tuning.utils.evaluator import MetricUnavailableError, RuleEvaluator
 logger = logging.getLogger(__name__)
 logger.setLevel(logging.DEBUG)
 
-logging.getLogger('hfcontrols').setLevel(logging.DEBUG)
-logging.getLogger('logcontrol').setLevel(logging.DEBUG)
-logging.getLogger('operation').setLevel(logging.DEBUG)
-logging.getLogger('patience').setLevel(logging.DEBUG)
-
 # Configuration keys
 CONTROLLER_METRICS_KEY = "controller_metrics"
 OPERATIONS_KEY = "operations"
--- a/tuning/trainercontroller/callback.py
+++ b/tuning/trainercontroller/callback.py
@@ -18,9 +18,9 @@
 # Standard
 from typing import Dict, List, Union
 import inspect
+import logging
 import os
 import re
-import logging
 
 # Third Party
 from simpleeval import EvalWithCompoundTypes, FeatureNotAvailable, NameNotDefined
@@ -305,7 +305,7 @@ class TrainerControllerCallback(TrainerCallback):
         kwargs["state"] = state
         kwargs["control"] = control
 
-        log_levels = dict((v,k) for k,v in logging._levelToName.items())
+        log_levels = dict((v, k) for k, v in logging._levelToName.items())
         # Check if there any metrics listed in the configuration
         if (
             CONTROLLER_METRICS_KEY not in self.trainer_controller_config
--- a/tuning/trainercontroller/callback.py
+++ b/tuning/trainercontroller/callback.py
@@ -45,7 +45,7 @@ from tuning.trainercontroller.operations import (
 from tuning.trainercontroller.patience import PatienceControl
 from tuning.utils.evaluator import MetricUnavailableError, RuleEvaluator
 
-logger = logging.getLogger()
+logger = logging.getLogger(__name__)
 
 # Configuration keys
 CONTROLLER_METRICS_KEY = "controller_metrics"
--- a/tuning/trainercontroller/callback.py
+++ b/tuning/trainercontroller/callback.py
@@ -417,7 +417,6 @@ class TrainerControllerCallback(TrainerCallback):
                             config_log_level_str = DEFAULT_TRIGGER_LOG_LEVEL
                     else:
                         control.config = DEFAULT_CONFIG
-                    logger.warning(f"LOG LEVELS: {log_levels}")
                     control.config[CONTROLLER_CONFIG_TRIGGER_LOG_LEVEL] = log_levels[
                         config_log_level_str
                     ]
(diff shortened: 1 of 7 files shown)


tuning/trainercontroller/callback.py
```diff
@@ -417,6 +417,7 @@ def on_init_end(
                             config_log_level_str = DEFAULT_TRIGGER_LOG_LEVEL
                     else:
                         control.config = DEFAULT_CONFIG
+                    logger.warning(f"LOG LEVELS: {log_levels}")
```
ashokponkumar, line 420: Why do we have a warning level comment here?

tuning/trainercontroller/operations/hfcontrols.py
```diff
@@ -10,6 +10,8 @@
 # Local
 from .operation import Operation
 
+logger = logging.getLogger()
```
ashokponkumar, line 13: Is this the standard way logging is being initialized elsewhere? Is there no logger name given?

ashokponkumar (changes requested):
Few changes

tuning/trainercontroller/callback.py
seshapad: Addressed.

tuning/trainercontroller/operations/hfcontrols.py
seshapad: Added `__name__`. Addressed.

ashokponkumar (approved):

kmehant (approved):
successfully transitions from transformers logger to python logging for trainer controller

Abhishek-TAMU:
Thank you for the PR. As a curiosity, I have one question here. 

`logger` defined [here](https://github.com/foundation-model-stack/fms-hf-tuning/blob/54f0ac978b6018f3cc2dcb6bf8611281ddb9cea1/tuning/trainercontroller/operations/hfcontrols.py#L13C1-L13C37) and [here](https://github.com/foundation-model-stack/fms-hf-tuning/blob/54f0ac978b6018f3cc2dcb6bf8611281ddb9cea1/tuning/trainercontroller/patience.py#L34), would have default log level of `warning`, and it its log level could be modified in 2 ways:

1- Setting root log level in a main file (OR in the file where these module are imported for usage) using `(logging.basicConfig(level=logging.DEBUG))` and then importing `hfcontrols/patience` module. 

2- Setting log level of `hfcontrols/patience` from main file using : `logging.getLogger('hfcontrols').setLevel(logging.DEBUG)` / `logging.getLogger('patience').setLevel(logging.DEBUG)`

Is the use case of these modules includes any of the above, to set log level of `hfcontrols/patience` module ?
Other than this changes look fine to me.

kmehant:
@seshapad FYA 🔝

seshapad:
> Thank you for the PR. As a curiosity, I have one question here.
> 
> `logger` defined [here](https://github.com/foundation-model-stack/fms-hf-tuning/blob/54f0ac978b6018f3cc2dcb6bf8611281ddb9cea1/tuning/trainercontroller/operations/hfcontrols.py#L13C1-L13C37) and [here](https://github.com/foundation-model-stack/fms-hf-tuning/blob/54f0ac978b6018f3cc2dcb6bf8611281ddb9cea1/tuning/trainercontroller/patience.py#L34), would have default log level of `warning`, and its log level could be modified in 2 ways:
> 
> 1- Setting root log level in a main file (OR in the file where these module are imported for usage) using `(logging.basicConfig(level=logging.DEBUG))` and then importing `hfcontrols/patience` module.
> 
> 2- Setting log level of `hfcontrols/patience` from main file using : `logging.getLogger('hfcontrols').setLevel(logging.DEBUG)` / `logging.getLogger('patience').setLevel(logging.DEBUG)`
> 
> Is the use case of these modules includes any of the above, to set log level of `hfcontrols/patience` module ? Other than this changes look fine to me.

Thank you @Abhishek-TAMU . Will look into these changes and get back to you.

seshapad:
> Thank you for the PR. As a curiosity, I have one question here.
> 
> `logger` defined [here](https://github.com/foundation-model-stack/fms-hf-tuning/blob/54f0ac978b6018f3cc2dcb6bf8611281ddb9cea1/tuning/trainercontroller/operations/hfcontrols.py#L13C1-L13C37) and [here](https://github.com/foundation-model-stack/fms-hf-tuning/blob/54f0ac978b6018f3cc2dcb6bf8611281ddb9cea1/tuning/trainercontroller/patience.py#L34), would have default log level of `warning`, and its log level could be modified in 2 ways:
> 
> 1- Setting root log level in a main file (OR in the file where these module are imported for usage) using `(logging.basicConfig(level=logging.DEBUG))` and then importing `hfcontrols/patience` module.
> 
> 2- Setting log level of `hfcontrols/patience` from main file using : `logging.getLogger('hfcontrols').setLevel(logging.DEBUG)` / `logging.getLogger('patience').setLevel(logging.DEBUG)`
> 
> Is the use case of these modules includes any of the above, to set log level of `hfcontrols/patience` module ? Other than this changes look fine to me.

@Abhishek-TAMU I have addressed the above comments. Request you to review.

Abhishek-TAMU:
> @Abhishek-TAMU I have addressed the above comments. Request you to review.

Thank you for the changes. I can see setting of log level for individual loggers in different modules in your changes. This looks good to me!
Feel free to move ahead, no further questions/comments from my side. Thanks!

tuning/trainercontroller/callback.py
```diff
@@ -45,7 +45,8 @@
 from tuning.trainercontroller.patience import PatienceControl
 from tuning.utils.evaluator import MetricUnavailableError, RuleEvaluator
 
-logger = logging.get_logger(__name__)
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.DEBUG)
```
anhuong, line 49: Should this reuse the `DEFAULT_TRIGGER_LOG_LEVEL` env var?

tuning/trainercontroller/operations/hfcontrols.py
```diff
@@ -10,6 +10,9 @@
 # Local
 from .operation import Operation
 
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.DEBUG)
```
anhuong, line 14: Why is the default log level for the trainer controller debug? In general it's warning to print minimal print messages and the user can increase the level as needed

ashokponkumar:
> Thank you for the PR. As a curiosity, I have one question here.
> 
> `logger` defined [here](https://github.com/foundation-model-stack/fms-hf-tuning/blob/54f0ac978b6018f3cc2dcb6bf8611281ddb9cea1/tuning/trainercontroller/operations/hfcontrols.py#L13C1-L13C37) and [here](https://github.com/foundation-model-stack/fms-hf-tuning/blob/54f0ac978b6018f3cc2dcb6bf8611281ddb9cea1/tuning/trainercontroller/patience.py#L34), would have default log level of `warning`, and its log level could be modified in 2 ways:
> 
> 1- Setting root log level in a main file (OR in the file where these module are imported for usage) using `(logging.basicConfig(level=logging.DEBUG))` and then importing `hfcontrols/patience` module.
> 
> 2- Setting log level of `hfcontrols/patience` from main file using : `logging.getLogger('hfcontrols').setLevel(logging.DEBUG)` / `logging.getLogger('patience').setLevel(logging.DEBUG)`
> 
> Is the use case of these modules includes any of the above, to set log level of `hfcontrols/patience` module ? Other than this changes look fine to me.

@Abhishek-TAMU There is no need to change the logging level within patience, isn't? It should just inherit whatever is set in logging module, isn't it?

Abhishek-TAMU:
> @Abhishek-TAMU There is no need to change the logging level within patience, isn't? It should just inherit whatever is set in logging module, isn't it?

@ashokponkumar Just to clarify, can I ask if you are mentioning the fact that, after the setting of root log level by `main` function of `SFTTrainer.py`, that root log level should be inherited by `patience.py` ?

ashokponkumar:
> > @Abhishek-TAMU There is no need to change the logging level within patience, isn't? It should just inherit whatever is set in logging module, isn't it?
> 
> @ashokponkumar Just to clarify, can I ask if you are mentioning the fact that, after the setting of root log level by `main` function of `SFTTrainer.py`, that root log level should be inherited by `patience.py` ?

Yes @Abhishek-TAMU . If we are setting the logging level for class/module logging, then every logger inherited from it will already have the level set, isn't it?

Abhishek-TAMU:
@ashokponkumar 

Yes, if **root** log level is set in any class/module using `logging.basicConfig(level=logging.DEBUG/WARNING/ERROR)` as I have mentioned [above](https://github.com/foundation-model-stack/fms-hf-tuning/pull/309#issuecomment-2299190335), and this code runs initially setting up the root log level of the repo, then subsequently any logger used in any of the other class/module in the same repo will **inherit log level from root logger**. 
Unless if you would want to specify the log level of logger of any module different than the root logger, then as mentioned above, `logger.setLevel()` function could be used, for that specific logger.

ashokponkumar:
> @ashokponkumar 
> 
> Yes, if **root** log level is set in any class/module using `logging.basicConfig(level=logging.DEBUG/WARNING/ERROR)` as I have mentioned [above](https://github.com/foundation-model-stack/fms-hf-tuning/pull/309#issuecomment-2299190335), and this code runs initially setting up the root log level of the repo, then subsequently any logger used in any of the other class/module in the same repo will **inherit log level from root logger**. 
> Unless if you would want to specify the log level of logger of any module different than the root logger, then as mentioned above, `logger.setLevel()` function could be used, for that specific logger.
> 
> 

@seshapad then we can remove the part where we are setting the logging levels.

seshapad:
> > @ashokponkumar
> > Yes, if **root** log level is set in any class/module using `logging.basicConfig(level=logging.DEBUG/WARNING/ERROR)` as I have mentioned [above](https://github.com/foundation-model-stack/fms-hf-tuning/pull/309#issuecomment-2299190335), and this code runs initially setting up the root log level of the repo, then subsequently any logger used in any of the other class/module in the same repo will **inherit log level from root logger**.
> > Unless if you would want to specify the log level of logger of any module different than the root logger, then as mentioned above, `logger.setLevel()` function could be used, for that specific logger.
> 
> @seshapad then we can remove the part where we are setting the logging levels.

This is done. Thank you @ashokponkumar @Abhishek-TAMU @anhuong

ashokponkumar:
@anhuong can we merge this PR now?

anhuong:
@Abhishek-TAMU please review this again when you get a chance, it looks good to me. If it looks good to you, you can approve and squash-merge it.

Abhishek-TAMU:
LGTM!!

Abhishek-TAMU (approved):
LGTM!!

PR #310, opened 2024-08-20
fix: do not resize embedding layer by default

kmehant:
Defaults multiple to 1 thereby not resizing the embedding unless user wishes to resize it by explicitly using this feature. 

Motivation is to support vLLM inference for LoRA which needs merging of adapter when embedding layer is resized with no real tokens. Findings from @Ssukriti  and team.


Ssukriti (approved):
Thank you! 

@aluu317 @willmj the fix is merged to default value 1 . You should be able to proceed with QLoRA testing on vLLM

PR #313, opened 2024-08-23
refactor: move removal of lm_head to save method

anhuong:
### Description of the change

- Utilize the removal of duplicated lm_head weight in granite models with llama arch in sft_trainer instead of just in accelerate_launch script. Removal of lm_head weight still occurs in the save_model_dir OR the last checkpoint.
- Now on call to `sft_trainer.save()` the lm_head weight will automatically be removed
- No longer need to reload the checkpoint to remove lm_head

### How to verify the PR

Currently testing

### Was the PR tested

### DIFF
--- a/tuning/sft_trainer.py
+++ b/tuning/sft_trainer.py
@@ -63,6 +63,7 @@ from tuning.utils.error_logging import (
     write_termination_log,
 )
 from tuning.utils.logging import set_log_level
+from tuning.utils.postprocessting_utils import get_highest_checkpoint
 from tuning.utils.preprocessing_utils import (
     format_dataset,
     get_data_collator,
@@ -367,6 +368,8 @@ def train(
 
 def save(path: str, trainer: SFTTrainer, log_level="WARNING"):
     """Saves model and tokenizer to given path.
+    If model is of type granite with llama arch and lm_head weight
+    is a duplicate of embeddings weight, remove lm_head.
 
     Args:
         path: str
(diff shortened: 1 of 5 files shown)


tuning/sft_trainer.py
```diff
@@ -613,19 +648,27 @@ def main(**kwargs):  # pylint: disable=unused-argument
...
-        except Exception as e:  # pylint: disable=broad-except
-            logger.error(traceback.format_exc())
-            write_termination_log(
-                f"Failed to save model to {training_args.save_model_dir}: {e}"
+        else:
+            # if granite with llama arch, remove lm_head in last checkpoint
+            save(
+                path=get_highest_checkpoint(training_args.output_dir),
+                trainer=trainer,
+                log_level=training_args.log_level,
```
anhuong, line 663: This is currently removing lm_head from the model that is being saved at save_model_dir OR the last checkpoint. A few questions...
- Do we need to remove from last checkpoint? This is for the use case if someone doesn't specify save_model_dir and retains current behavior but it is a little weird for the final checkpoint to be different than the rest of the checkpoints.
- Do we want to remove from each checkpoint? This would require loading up each of the checkpoints again and resaving it, unless SFTTrainer.trainer has access to the checkpoints still but AFAIK it only has access to the last checkpoint

anhuong:
will close in favor of PR: https://github.com/foundation-model-stack/fms-hf-tuning/pull/333

PR #314, opened 2024-08-23
refactor: set log level to separate out train args

anhuong:
### Description of the change

Separate out python logging and setting log level with setting log level using train_args and altering `train_args.log_level` to match python logging level. 

### Was the PR tested

### DIFF
--- a/tuning/utils/logging.py
+++ b/tuning/utils/logging.py
@@ -53,12 +53,24 @@ def set_log_level(train_args, logger_name=None):
             else os.environ.get("TRANSFORMERS_VERBOSITY")
         )
 
-    logging.basicConfig(
-        format="%(levelname)s:%(filename)s:%(message)s", level=log_level.upper()
-    )
-
+    train_logger = logging.getLogger()
     if logger_name:
         train_logger = logging.getLogger(logger_name)
-    else:
-        train_logger = logging.getLogger()
+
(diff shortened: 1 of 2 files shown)


tuning/utils/logging.py
```diff
@@ -53,12 +53,24 @@ def set_log_level(train_args, logger_name=None):
...
-    )
-
+    train_logger = logging.getLogger()
     if logger_name:
         train_logger = logging.getLogger(logger_name)
-    else:
-        train_logger = logging.getLogger()
+
+    set_python_log_level(log_level, train_logger)
+    set_python_log_level(log_level)
```
Abhishek-TAMU, line 61: Nice to have a generic function `set_python_log_level` to set log level of any logger passed to it from any module of the code. Thank you!

Regarding function `set_log_level`, according to me, it is called to **set root log level in the code** based on either CLI using `train_args`or ENV variable. Hence, I believe, if we don't want to call `set_python_log_level` twice in the function `set_log_level`, we can modify the code in function `set_log_level` as below:

```
set_python_log_level(log_level)
train_logger = logging.getLogger()
if logger_name:
    train_logger = logging.getLogger(logger_name)

return train_args, train_logger
```
Call `set_python_log_level(log_level)`, will set the root log level and `train_logger` with or without `logger_name`, will inherit the root log level.

PR #315, opened 2024-08-26
docs: Add findings from exploration into model tuning performance degradation

willmj:
### Description of the change
Add our findings from exploration into performance degradation. If anything else needs to be added or expanded upon, let me know!


README.md
```diff
@@ -270,6 +270,13 @@ generation_config.json	model-00005-of-00006.safetensors  tokenizer.model
 
 </details>
 
+#### Optimizing training time
+Writing models to Cloud Object Storage (COS) is an expensive operation. Saving model checkpoints to a local directory causes much faster training times than writing to COS. You can use `output_dir` and `save_model_dir` to control which type of storage you write your checkpoints and final model to. Using a local directory over COS can lead to up to a 40% train time reduction.
```
anhuong, line 274: I do like the addition of numbers although we only tested this on one model type that is not super big and only on a small dataset. I'm wary of including the numbers even though saying "up to" is key here. What do you think?

anhuong (commented):
Thanks for adding this doc Will!

README.md
willmj: Yeah I also had this concern, I think we can remove it until further testing is done by RH team.

anhuong (approved):
Thanks Will!

PR #321, opened 2024-08-30
feat: Add subpackage for gptq dependencies to use in evaluation/inference

aluu317:
### Was the PR tested


aluu317:
Duplicate, just opened to test

PR #322, opened 2024-08-30
doc: Add qLoRA README

aluu317:
### Description of the change

We are now supporting qLoRA tuning technique, so this PR adds documentation on that.

### Was the PR tested


README.md
```diff
@@ -9,6 +9,7 @@
   - [Tips on Parameters to Set](#tips-on-parameters-to-set)
 - [Tuning Techniques](#tuning-techniques)
   - [LoRA Tuning Example](#lora-tuning-example)
+  - [qLoRA Tuning Example](#qlora-tuning-example)
```
anhuong, line 12: nit: is technically QLoRA
```suggestion
  - [QLoRA Tuning Example](#qlora-tuning-example)
```

README.md
```diff
@@ -432,6 +433,79 @@ Example 3:
 
 _________________________
 
+
+### qLoRA Tuning Example
```
anhuong, line 437: same nit:
```suggestion
### QLoRA Tuning Example
````

Comment (anhuong) [reply README.md:437]: I actually think we want to be more specific here. If these docs are just for auto_gptq and not bnb_qlora then we should note that this is 4bit GPTQ-LoRA with AutoGPTQ. Whereas bnb_qlora is 4bit QLoRA with bitsandbytes as both are named QLoRA

File: README.md [hunk 3]
```diff
@@ -432,6 +433,79 @@ Example 3:
...
+
+This method is similar to LoRA Tuning, but the base model is a quantized model.
```
anhuong, line 439: Would be useful to add a note pointing to the fms-acceleration section and note that is how qlora is enabled - https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/README.md#fms-acceleration

README.md
```diff
@@ -432,6 +433,79 @@ Example 3:
...
+Set `peft_method` to `"lora"`. You can pass any of LoraConfig, see section on [LoRA Example](#lora-tuning-example).
```
anhuong, line 440: Would be good to note the additional qlora flags first and then note that the lora params are the same. Would be useful to have more details on what the LoRA quantization config flags mean and what kernels are supported. Right now only triton is supported

README.md
```diff
@@ -432,6 +433,79 @@ Example 3:
...
+class AutoGPTQLoraConfig:
+
+  # auto_gptq supports various kernels, to select the kernel to use.
+  kernel: str = "triton_v2"
+
+  # allow auto_gptq to quantize a model before training commences.
+  # NOTE: currently this is not allowed.
+  from_quantized: bool = True
+
+```
```
anhuong, line 460: I would not include this as it causes more confusion. Right now we only support triton kernel and from_quantized has the note that this is currently not allowed so we should not expose it to users

README.md
```diff
@@ -432,6 +433,79 @@ Example 3:
...
+    "peft_method": "lora",
+    "r": 8,
+    "lora_dropout": 0.05,
+    "lora_alpha": 16,
+    "target_modules": ["c_attn", "c_proj"]
+    "auto_gptq": ["triton_v2"]
+}
+```
+
+Similarly to LoRA, the `target_modules` are the names of the modules to apply the adapter to. See the LoRA [section](#lora-tuning-example) on `target_modules` for more info.
```
anhuong, line 503: You can link directly to the target modules section - https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/README.md#fms-acceleration

anhuong (changes requested):
Thanks for getting the docs started Angel! Would be good to clarify what the flags we can pass are and what is currently supported.

README.md
anhuong: My bad I think this is the link I wanted -- https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/README.md#how-to-get-list-of-lora-target_modules-of-a-model

anhuong: But I see you have the `Lora-tuning-example` link and note the target modules section so that works

anhuong (approved):
LGTM, thanks Angel! although i see the DCO failed on your latest commit not being signed. I think this can also be brought out of review phase

PR #324, opened 2024-08-30
fix: remove fire ported from Hari's PR #303

HarikrishnanBalagopal:
### Description of the change

Remove the 2nd CLI parser (fire) because it's not necessary and throws errors when training with LoRA.

### Related issue number

Fixes https://github.com/foundation-model-stack/fms-hf-tuning/issues/47

### How to verify the PR

`make test`

### Was the PR tested

- [x] I have added >=1 unit test(s) for every new method I have added.
- [x] I have ensured all unit tests pass

PR #326, opened 2024-09-02
fix: need to pass skip_prepare_dataset for pretokenized dataset due to breaking change in HF SFTTrainer

HarikrishnanBalagopal:
### Description of the change

Fix test failure

### Related issue number

Fixes https://github.com/foundation-model-stack/fms-hf-tuning/pull/324#issuecomment-2322823510

### How to verify the PR

`make test`

### Was the PR tested

### DIFF
--- a/tests/test_sft_trainer.py
+++ b/tests/test_sft_trainer.py
@@ -35,7 +35,6 @@ from tests.data import (
     EMPTY_DATA,
     MALFORMATTED_DATA,
     MODEL_NAME,
-    TWITTER_COMPLAINTS_DATA_INPUT_OUTPUT_JSON,
     TWITTER_COMPLAINTS_DATA_INPUT_OUTPUT_JSONL,
     TWITTER_COMPLAINTS_DATA_JSON,
     TWITTER_COMPLAINTS_DATA_JSONL,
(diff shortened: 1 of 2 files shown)


willmj:
This seems to be the line in the SFTTrainer code that caused the error on our side:
https://github.com/huggingface/trl/compare/v0.9.6...v0.10.1#diff-29597a530a3b595f2140e524e585a0e1458983b8c1c382e3fca48b314892dcf8L204

willmj (approved):
LGTM!

anhuong:
The issue is that trl created a release 5 days ago that our PRs are now picking up. v0.10.1 includes this line change https://github.com/huggingface/trl/blob/d57e4b726561e5ae58fdc335f34029052944a4a3/trl/trainer/sft_trainer.py#L348 where if packing=False (set by default in our code) then it expects that data preparation must be completed. For pretokenized datasets, data preparation is not needed, thus we need to set `skip_prepare_dataset=False`. For the other dataset formats we accept, `dataset_text_field` is set and thus data processing is run in SFTTrainer.

Angel had the idea that we could only add `skip_prepare_dataset=False` in our pretokenized dataset tests. Although this is true and would resolve the issue, this is something we need to pass into SFTTrainer for all pretokenized datasets, that the user wouldn't be able to set on their own.

anhuong (approved):

PR #331, opened 2024-09-04
fix: avoid docker push timeout by splitting the single big layer into 2 smaller layers

HarikrishnanBalagopal:
### Description of the change

Split the 6GB layer into a two around 3GB layers

### Was the PR tested


kmehant:
cherry picked

PR #333, opened 2024-09-05
fix: remove lm_head post processing

Abhishek-TAMU:
### Description of the change

Removal of [lm_head hack](https://github.com/foundation-model-stack/fms-hf-tuning/pull/258) which was made to fix lm_head issue and now fixed with newer vllm versions, the [change](https://github.com/vllm-project/vllm/pull/6758) coming in as of v0.5.4 

### Related issue number

[#1166](https://github.ibm.com/ai-foundation/watson-fm-stack-tracker/issues/1166)

### How to verify the PR

Running LoRA and full fine tuning of granite-3b and llama-8b model without removal of lm_head able to run inference on.

### Was the PR tested

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -21,7 +21,7 @@ ARG PYTHON_VERSION=3.11
 ARG WHEEL_VERSION=""
 ## Enable Aimstack if requested via ENABLE_AIM set to "true"
 ARG ENABLE_AIM=false
-ARG ENABLE_FMS_ACCELERATION=false
+ARG ENABLE_FMS_ACCELERATION=true
 
 ## Base Layer ##################################################################
 FROM registry.access.redhat.com/ubi9/ubi:${BASE_UBI_IMAGE_TAG} AS base
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -105,7 +105,7 @@ FROM cuda-devel AS python-installations
 ARG WHEEL_VERSION
 ARG USER
 ARG USER_UID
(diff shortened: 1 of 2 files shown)


build/Dockerfile
```diff
@@ -105,7 +105,7 @@ FROM cuda-devel AS python-installations
 ARG WHEEL_VERSION
 ARG USER
 ARG USER_UID
-ARG ENABLE_FMS_ACCELERATION=false
+ARG ENABLE_FMS_ACCELERATION=true
```
anhuong, line 108: Note that with this change, fms-acceleration can be installed by default thus enabling QLoRA support by default. The lm_head removal was blocking QLoRA enablement.

anhuong:
After testing, found that accelerate version is not working as expected. 

New logic intorduced in get_state_dict, also removes the top-level FSDP wrapper from the model. So then since FSDP keeps flattened params, all the parameters managed by the top-level wrapper will now remained flattened when model.state_dict is called. The other child FSDP wrappers will protect their parameters, since when the state_dict call recurses to them, they will use the FSDP version of state_dict to unwrap the wrappers.

This results in error:
```
size mismatch for model.embed_tokens.weight: copying a param with shape torch.Size([62915840]) from checkpoint, the shape in current model is torch.Size([49152, 2560]).
size mismatch for model.norm.weight: copying a param with shape torch.Size([0]) from checkpoint, the shape in current model is torch.Size([2560]). 
```

anhuong (approved):

PR #335, opened 2024-09-10
dep: cap transformers version due to FSDP bug

anhuong:
### Description of the change

v0.34.0 and v0.34.1 of accelerate has a bug where when training with FSDP, the weight is being flattened and is unable to be loaded. Capping at v0.33 before the change went in and at a verified point.

### Was the PR tested

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,7 +27,7 @@ classifiers=[
 ]
 dependencies = [
 "numpy>=1.26.4,<2.0",
-"accelerate>=0.20.3,<0.40",
+"accelerate>=0.20.3,<0.34",
 "transformers>4.41,<5.0",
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -27,7 +27,7 @@ classifiers=[
 ]
 dependencies = [
 "numpy>=1.26.4,<2.0",
-"accelerate>=0.20.3,<0.40",
+"accelerate==0.33.0",
```
willmj, line 30: Could we do
```
"accelerate>=0.20.3,<0.34"
```

willmj (changes requested):

willmj (approved):
LGTM!

PR #328, opened 2024-09-03
deps: Add protobuf to support ALLaM models

willmj:
### Description of the change
Add protobuf v5.28.0 to fms-hf-tuning for compatibility with certain models

### Was the PR tested

- [ ] I have added >=1 unit test(s) for every new method I have added.
- [ ] I have ensured all unit tests pass

PR #336, opened 2024-09-10
deps: Add protobuf to enable compatibility with certain models

willmj:
### Description of the change
Add protobuf to enable compatibility with certain models, as done in patch release [v1.2.2](https://github.com/foundation-model-stack/fms-hf-tuning/releases/tag/v1.2.2)

### Was the PR tested


anhuong:
Related PR into release branch - https://github.com/foundation-model-stack/fms-hf-tuning/pull/328

anhuong:
I initially was thinking we could pull in the same commit onto release branch to try to minimize the merge conflict at release time by doing something like `git cherry-pick 1107e0042e7e4fdfa8e60f8fb33a03327b2278f6` but since this PR will squash merge into main branch, I don't think we can avoid having the same change but different commits on main and release branch. I don't think it should be too much of an issue at release time since they are the same change, but we shall see.

anhuong (approved):
Thanks for adding this in and not forgetting Will!

PR #337, opened 2024-09-10
fix: add enable_aim build args in all stages needed

anhuong:
### Description of the change

When I tried to build the image with AIM, it installed AIM correctly but didn't have the aim directory from the final stage since the build arg wasn't available in the final stage. This fixes the issue by defining enable_aim at the top and then enabling the flag in the stages where it is used. 

### Was the PR tested


willmj (approved):
LGTM!

PR #338, opened 2024-09-10
fix: utilities to post process checkpoint for LoRA

Ssukriti:
### Description of the change

utility function to post process checkpoint after LoRA tuning to convert to format required by vLLM.
This will need to be called at end of LoRA tuning to allow inferencing on LoRA, **for models for which we have added new tokens.** 

Since it is fast enough to load adapters.safetensors , added it as a post-processing function.

This PR adds a script that can be called after tuning to do the processing.

### Related issue number

https://github.ibm.com/ai-foundation/watson-fm-stack-tracker/issues/1210
Details on vLLM issue: https://github.com/vllm-project/vllm/issues/2816#issuecomment-1934607503

**Context:** Embedding vectors for new tokens need to be placed in `new_embeddings.safetensors` and lm_head.weight should be deleted from `adapter_model.safetensors` as per  https://github.com/vllm-project/vllm/issues/2816#issuecomment-1934607503

### How to verify the PR

1. Function called on llama model which was LoRA tuned - post-processed and being tested on vLLM
2. Verified PR with every unique model architecture we support

### Was the PR tested


tuning/utils/merge_model_utils.py
```diff
@@ -102,3 +102,77 @@ def fetch_base_model_from_checkpoint(checkpoint_model: str) -> str:
...
+                embeddings_weights_in_adapters = True
+                if len(sorted_token_indexes) >=1:
+                    raise NotImplementedError("Seems like embeddings are resized without adding new tokens. \
+                                               Cannot be post-processed to load on vLLM. Try setting \
+                                               parameter `embedding_size_multiple_of` to 1" )
+
+        # Post-processing is needed to copy out new vectors
+        if embeddings_weights_in_adapters:
+            for k in f.keys():
+                if 'lm_head.weight' in k:
```
fabianlim, line 148: `lm_head` is not a standard name. It is the name used in most models, but there might be differences say from LLama and GPTBIgCode (which the older granite models depend on)

tuning/utils/merge_model_utils.py
```diff
@@ -102,3 +102,77 @@ def fetch_base_model_from_checkpoint(checkpoint_model: str) -> str:
...
+    import shutil
+    src_files = os.listdir(src)
+    for file_name in src_files:
+        if file_name in exclude_files:
+            continue
+        full_file_name = os.path.join(src, file_name)
+        if os.path.isfile(full_file_name):
+            shutil.copy(full_file_name, dest)
+
+def post_process_vLLM_adapters_new_tokens(path_to_checkpoint: str, modified_checkpoint_path:str = None):
```
fabianlim, line 116: I think its a good idea to document down 
1. why we need this function
2. what this function tries to do, and description of the rough logic.

tuning/utils/merge_model_utils.py
```diff
@@ -102,3 +102,77 @@ def fetch_base_model_from_checkpoint(checkpoint_model: str) -> str:
...
+        adapters = {}
+        embeddings_weights_in_adapters = False
+        # Quickly check if post-processing is needed by checking adapters file for weights
+        for k in f.keys():
+            if 'lm_head.weight' in k or 'embed_tokens.weight' in k:
+                embeddings_weights_in_adapters = True
+                if len(sorted_token_indexes) >=1:
+                    raise NotImplementedError("Seems like embeddings are resized without adding new tokens. \
+                                               Cannot be post-processed to load on vLLM. Try setting \
+                                               parameter `embedding_size_multiple_of` to 1" )
```
kmehant, line 143: I am little confused, this check would not convey that embeddings are resized without adding new tokens isn't it? 

Should not it be this?
```suggestion
                if len(sorted_token_indexes) ==0:
                    raise NotImplementedError("Seems like embeddings are resized without adding new tokens. \
                                               Cannot be post-processed to load on vLLM. Try setting \
                                               parameter `embedding_size_multiple_of` to 1" )
```

tuning/utils/merge_model_utils.py
```diff
@@ -102,3 +102,77 @@ def fetch_base_model_from_checkpoint(checkpoint_model: str) -> str:
...
+        # Post-processing is needed to copy out new vectors
+        if embeddings_weights_in_adapters:
+            for k in f.keys():
+                if 'lm_head.weight' in k:
+                    lm_head = f.get_tensor(k)
+                    # pull out tensor values of new tokens
+                    if len(sorted_token_indexes)==1:
+                        new_output_embeddings = lm_head[sorted_token_indexes[0]:sorted_token_indexes[0]+1]
+                    elif len(sorted_token_indexes)>1:
+                        new_output_embeddings = lm_head[sorted_token_indexes[0]:sorted_token_indexes[-1]]
```
kmehant, line 154: This should work right? 🤔 
```suggestion
                    new_output_embeddings = lm_head[sorted_token_indexes[0]:sorted_token_indexes[-1]+1]
```

tuning/utils/merge_model_utils.py
```diff
@@ -102,3 +102,77 @@ def fetch_base_model_from_checkpoint(checkpoint_model: str) -> str:
...
+                    # vLLM requires renaming to output_embeddings
+                    new_embeddings['output_embeddings'] = new_output_embeddings
+            
+                elif 'embed_tokens.weight' in k:
+                    embed_tokens = f.get_tensor(k)
+                    # pull out tensor values of new tokens
+                    if len(sorted_token_indexes)==1:
+                        new_input_embeddings = embed_tokens[sorted_token_indexes[0]:sorted_token_indexes[0]+1]
+                    elif len(sorted_token_indexes)>1:
+                        new_input_embeddings = embed_tokens[sorted_token_indexes[0]:sorted_token_indexes[-1]]
```
kmehant, line 164: WDYT?
```suggestion
                    new_input_embeddings = embed_tokens[sorted_token_indexes[0]:sorted_token_indexes[-1]+1]
```

kmehant (changes requested):
Couple of comments, thanks

tuning/utils/merge_model_utils.py
fabianlim: maybe a suggestion

```python
def infer_head_embedding_names_from_vocab_size(vocab_size):
    try:
        # get (names, shapes) of all the modules that have one dimension that matches the vocab size
        name_and_sizes = [
            (name.split('.'), tuple(mod.weight.shape)) for name, mod in model.named_modules()
            if hasattr(mod, 'weight') and vocab_size in tuple(mod.weight.shape)
        ]

        # if there are more than 2 items left, then all bets are off
        if len(name_and_sizes) != 2:
            raise IndexError

        # sort them by the number of levels
        name_and_sizes = sorted(sizes, key= lambda x: len(x[0]))

        # the heuristic is that the lm_head will usually be at the CausalLM, so it will be at 
        # the lowest level, and the embedding will be inside the LM and at a higher level
        lm_head_name, _ = name_and_sizes[0]
        embed_name, _  = name_and_sizes[1]
        return lm_head_name[-1], embed_name[-1]
    except IndexError:
        return 'lm_head', 'embed_tokens' 

# usage, either use model.config.vocab_size, or if you know the resized vocab, use that
infer_head_embedding_names_from_vocab_size(model.config.vocab_size)
```

Ssukriti: thats a good idea, thank you!

tuning/utils/merge_model_utils.py
Ssukriti: oops yes, while running it I corrected, and forgot to commit the change, thanks

tuning/utils/merge_model_utils.py
```diff
@@ -102,3 +105,100 @@ def fetch_base_model_from_checkpoint(checkpoint_model: str) -> str:
             "Base model adapter config exists, but has no base_model_name_or_path!"
         )
     return adapter_dict["base_model_name_or_path"]
+
+
+def _copy_files_to_directory(src: str, dest: str, exclude_files: list[str] = None):
```
aluu317, line 110: Default value of `excluded_files` is None and not `[ ]` is because https://nikos7am.com/posts/mutable-default-arguments/

aluu317:
I added a commit to add unit test in `test_merge_model_utils.py`. Steps to run it (since Github Action skips it due to not having cuda);
- Start a sleep pod with volume mount to our `fmaas-integration-tests` COS bucket.
- Change the value in test file just for this run: 
```
DUMMY_TUNED_LLAMA_WITH_ADDED_TOKENS = "/testing/tuning/output/llama_dummy_LoRA/checkpoint-35"
```
- Copy test file into pod: 
```
oc cp tests/utils/test_merge_model_utils.py angel-sleep-pod-sft-trainer:/home/tuning/.local/lib/python3.11/site-packages/tuning/utils/test_merge_model_utils.py
```
- In sleep pod, `pip install pytest`. Then run:
```
[tuning@angel-sleep-pod-sft-trainer app]$ cd /home/tuning/.local/lib/python3.11/site-packages/tuning/utils
[tuning@angel-sleep-pod-sft-trainer utils]$ pytest test_merge_model_utils.py 
======================================= test session starts ========================================
platform linux -- Python 3.11.7, pytest-8.3.3, pluggy-1.5.0
rootdir: /home/tuning/.local/lib/python3.11/site-packages/tuning/utils
collected 1 item                                                                                   

test_merge_model_utils.py .                                                                  [100%]

======================================== 1 passed in 3.50s =========================================
```

tuning/sft_trainer.py
```diff
@@ -651,6 +655,25 @@ def main():
...
+    # post process lora
+    if isinstance(tune_config, peft_config.LoraConfig):
+        try:
+            checkpoint_dir = job_config.get("save_model_dir")
+            if not checkpoint_dir:
+                checkpoint_dir = os.path.join(
+                    training_args.output_dir,
+                    get_highest_checkpoint(training_args.output_dir),
+                )
+            print(training_args)
```
Ssukriti, line 667: is print(training_args) added on purpose?

tuning/sft_trainer.py
```diff
@@ -651,6 +655,25 @@ def main():
             )
             sys.exit(INTERNAL_ERROR_EXIT_CODE)
 
+    # post process lora
+    if isinstance(tune_config, peft_config.LoraConfig):
```
Ssukriti, line 659: can we do this with a flag 'vLLM_format_LoRA' ? I am not sure the formatted files will work with HF inference and the postprocessing shouldnt be needed for all use cases - which dont need vLLM for example

tuning/sft_trainer.py
willmj: No, thanks

willmj:
Note these tests are failing because they are running old unit tets (see changes in previous [commit](https://github.com/foundation-model-stack/fms-hf-tuning/pull/338/commits/57cadc3c3b5379b6352697fbfee8188369a272b9))

willmj:
Good news! llama3 and mistral both inference with the changes in the branch!
If you want to test for yourself:
llama model location: `/fmaas-integration-tests/tuning/output/llama3-8b_pre_trained/lora/20240920143503-save_model/`
mistral model location: `/fmaas-integration-tests/tuning/output/mistral-7b-v0.1/lora/20240919184910-t1-save/`
Screenshot 2024-09-20 at 3 37 38 PM
Screenshot 2024-09-20 at 2 20 50 PM

tuning/utils/merge_model_utils.py
Ssukriti: thanks for suggestion @fabianlim . We tested with every unique model architecture we support as @willmj documented https://github.com/foundation-model-stack/fms-hf-tuning/pull/338#issuecomment-2364505870 and all the models had these standard names. 

Hence we decided that inferring names is lower priority and we will add it an enhancement in future PR after this release - this PR is currently blocking release

Ssukriti:
@fabianlim @kmehant the PR is ready for a second review. I had to make some changes since you last reviewed, which came out of the team's testing (@Abhishek-TAMU @willmj ). Describing changes here:

1. I cannot rely on file 'added_tokens.json' , as the file is not consistently produced for the new PretrainedTokenizerFast tokenizer types. Only few tokenizer types result in that file being produced with the resize function. Hence I created my own artifact 'added_tokens_info.json' containing the information we need for post-processing - how many tokens were aded. 

2. To get information of how many new tokens were added while tuning, the train() function now returns additional metadata containing that information. The utility function for postprocessing consumes that metadata. 

3. I could not call the post-processing from main() , as @Abhishek-TAMU and @willmj pointed out errors due to multi-processing. We launch the main script using accelerate that uses multiple processes. This post-processing needs to happen only 1 time after training is complete, hence I added a new step and script to be called after tuning is complete using accelerate launch. 

Next PR: to complete this contribution, we have to add the post-processing step in the build /accelerate_launch.py script under a flag so watsonX can utilize it. This will be done next. 


NOTE: While there may be ways to figure out what tokens were added at end of tuning even without 'added_tokens.json' file, it was turning out to be complicated as some tokenizers just append new tokens to existing 'special_tokens.json' file. Also while tuning we have this information, so I just chose to write it to a file anyway.

tuning/sft_trainer.py
```diff
@@ -242,23 +242,16 @@ def train(
...
         if isinstance(tokenizer, (LlamaTokenizer, LlamaTokenizerFast)):
-            tokenizer.add_special_tokens(
-                {
-                    "bos_token": "<s>",
-                    "eos_token": "</s>",
-                    "unk_token": "<unk>",
-                    "pad_token": "<pad>",
-                }
-            )
+            special_tokens_dict["bos_token"] = "<s>"
```
Ssukriti, line 249: this was just done to ensure we add and resize embeddings in same function

tuning/sft_trainer.py
```diff
@@ -418,8 +410,9 @@ def train(
         )
 
     trainer.train(resume_from_checkpoint)
-
-    return trainer
+    additional_metadata = {}
+    additional_metadata["added_tokens_info"] = added_tokens_dict
+    return trainer, additional_metadata
```
Ssukriti, line 415: Returning additional metadata from train() containing information of newly added tokens. @kmehant let me know if this change is ok with you - it might be disruptive if using SDK and if you were relying on 1 return value. We can make this a major release with the change.

Alternatively, I can write the file artifacts I want inside 'output_dir' and later can copy it to save_model_dir as well, if return value is an issue. I just thought this return value could be extendable to any other metadata we need in future

tests/artifacts/tuned_llama_with_added_tokens/adapter_config.json
```diff
@@ -0,0 +1,29 @@
+{
```
Ssukriti, line 1: These files are all just dummy LoRA artifacts needed for unit tests

anhuong (changes requested):
Great work all! Few questions and changes, thanks for adding docs! I know you talked about adding some unit tests for the tokenizer_data_utils which would be good as well

scripts/post_process_adapters_vLLM.py
```diff
@@ -0,0 +1,80 @@
...
+    parser.add_argument(
+        "--output_model_path",
+        help="Output directory where post-processed artifacts will be stored. \
+                                    If not provided, artifacts will be modified in place",
+        default=None,
+    )
+    args = parser.parse_args()
+
+    if args.output_model_path is None:
+        output_model_path = args.model_path
```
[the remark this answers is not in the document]
Ssukriti: good catch

scripts/post_process_adapters_vLLM.py
```diff
@@ -0,0 +1,80 @@
...
+    if os.path.exists(os.path.join(args.model_path, "added_tokens_info.json")):
```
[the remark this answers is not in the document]
Ssukriti: I did mention in help as well `Path to tuned model containing either one or multiple checkpoints. \
                              Path should have file added_tokens_info.json produced by tuning. \`
                              
so if users face error that file is not found they can check the help?

scripts/post_process_adapters_vLLM.py
```diff
@@ -0,0 +1,80 @@
...
+    else:
+        print("file added_tokens_info.json not in model_path. Cannot post-processes")
+
+    if os.path.exists(os.path.join(args.model_path, "adapter_model.safetensors")):
+        post_process_vLLM_adapters_new_tokens(
+            args.model_path, output_model_path, num_added_tokens
+        )
+    # if multiple checkpoints in directory, process each checkpoint
+    found_checkpoints = 0
+    for _, dirs, _ in os.walk(args.model_path, topdown=False):
```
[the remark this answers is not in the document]
Ssukriti: I think its ok, if path they provided has any adapters it will be found or if it has any checkpoints it will be found and processed. In this code I dont know if its output-dir or save_model_dir , its just a path with some adapters

scripts/post_process_adapters_vLLM.py
```diff
@@ -0,0 +1,80 @@
...
+        for name in dirs:
+            if "checkpoint-" in name.lower():
+                post_process_vLLM_adapters_new_tokens(
+                    os.path.join(args.model_path, name),
+                    os.path.join(output_model_path, name),
+                    num_added_tokens,
+                )
+                found_checkpoints = 1
+        if found_checkpoints and output_model_path != args.model_path:
```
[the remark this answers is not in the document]
Ssukriti: moving it outside if, yes its needed as there can be other artifacts besides the checkpoint in the folder, since we are looping through checkpoint it will only copy files inside checkpoint otherwise. This ensures any files in root directory remain/or are copied

tuning/sft_trainer.py
```diff
@@ -418,8 +410,9 @@ def train(
         )
 
     trainer.train(resume_from_checkpoint)
-
-    return trainer
+    additional_metadata = {}
+    additional_metadata["added_tokens_info"] = added_tokens_dict
```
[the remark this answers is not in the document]
Ssukriti: ya in future we can add any information and also we may need 'embedding_size' to infer weight names as Fabian suggested later on

tuning/sft_trainer.py
```diff
@@ -693,6 +687,33 @@ def main():
...
+            if training_args.save_model_dir:
+                # Write number of added tokens to artifacts
+                with open(
+                    os.path.join(
+                        training_args.save_model_dir, "added_tokens_info.json"
+                    ),
+                    "w",
+                    encoding="utf-8",
+                ) as f:
+                    json.dump(additional_train_info["added_tokens_info"], f)
```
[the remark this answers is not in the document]
Ssukriti: no , writing it in all cases to differentiate between "tuning was not done on latest code / no tokens were added". The file will always exist, it will contain value 0 if no tokens added

tuning/data/tokenizer_data_utils.py
```diff
@@ -44,3 +52,4 @@ def tokenizer_and_embedding_resize(
 
         input_embeddings[-num_new_tokens:] = input_embeddings_avg
         output_embeddings[-num_new_tokens:] = output_embeddings_avg
+    return {"num_new_tokens": num_new_tokens, "new_embedding_size": embedding_size}
```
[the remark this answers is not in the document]
Ssukriti: its size of embeddings at end of tuning , whichever case it is. Will be helpful to decipher weight names based on size

tests/utils/test_merge_model_utils.py
```diff
@@ -0,0 +1,75 @@
...
+dir_path = os.path.dirname(os.path.realpath(__file__))
+DUMMY_TUNED_LLAMA_WITH_ADDED_TOKENS = os.path.join(
+    dir_path, "../artifacts/tuned_llama_with_added_tokens"
+)
+
+
+@pytest.mark.skipif(
+    not (torch.cuda.is_available()),
+    reason="Only runs if cuda is supported",
+)
```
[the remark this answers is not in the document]
Ssukriti: I think this was added for get_tensors() in the utility function . @willmj did the test not run on GithubActions and thats why it was skipped?

tuning/sft_trainer.py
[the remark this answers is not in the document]
Ssukriti: added a check in script to exit early if num_added_tokens == 0 though https://github.com/foundation-model-stack/fms-hf-tuning/pull/338/files#diff-443a0a5c8690e1a9251b2805f8e5850c97381d40b47cd90c5716eeaca84706d9R61

scripts/post_process_adapters_vLLM.py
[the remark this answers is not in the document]
anhuong: ahh thanks for the explanation, yes this is super important to copy over the training logs and any other files needed. I see that the copy method also only copies the files and not the directory so its not re-copying over the checkpoints

tests/utils/test_merge_model_utils.py
```diff
@@ -0,0 +1,75 @@
...
+def test_post_process_vLLM_adapters_new_tokens():
+    """Ensure that in post-process, we output the correct format supported by vLLM for added_tokens
```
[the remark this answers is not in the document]
anhuong: Now that i understand better, I think we should add a test where the added_tokens = 0 and show that post-processing doesn't occur

scripts/post_process_adapters_vLLM.py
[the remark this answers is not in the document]
anhuong: I was thinking this could be added to the README but I think it it makes sense that its provided in the script

Ssukriti:
@anhuong @willmj I have added sufficient unit tests pertaining to this change. Lets wait on @ashokponkumar to check the PR as well.

tuning/sft_trainer.py
```diff
@@ -242,23 +246,16 @@ def train(
...
+            special_tokens_dict["eos_token"] = "</s>"
+            special_tokens_dict["unk_token"] = "<unk>"
+            special_tokens_dict["pad_token"] = "<pad>"
         elif isinstance(tokenizer, (GPT2Tokenizer, GPTNeoXTokenizerFast)):
-            tokenizer.add_special_tokens(
-                {
-                    "pad_token": "<pad>",
-                }
-            )
+            special_tokens_dict["pad_token"] = "<pad>"
```
ashokponkumar, line 258: Sorry for my ignorance. 

But just wondering, why not just remove this logic that is specific to llama tokernizer or GPT2 tokenizer out of fms-hf-tuning, we just assume the base model just has it? This will remove the need for all these additional tokens. For example, I am not sure why `unk` token is being added. My undertanding was BPE based tokenizers never end up creating `unk` tokens.

Or we can make these tokens adding a externally flag, that allows adding custom tokens to any tokernizer. This would make fms-hf-tuning more generic.

tuning/sft_trainer.py
ashokponkumar: If we remove custom addition of tokens from fms-hf-tuning, we might not need to return this added tokens metadata.

ashokponkumar (commented):
Few clarifications..

tuning/utils/merge_model_utils.py
```diff
@@ -102,3 +105,95 @@ def fetch_base_model_from_checkpoint(checkpoint_model: str) -> str:
             "Base model adapter config exists, but has no base_model_name_or_path!"
         )
     return adapter_dict["base_model_name_or_path"]
+
+
+def copy_files_to_directory(src: str, dest: str, exclude_files: list[str] = None):
```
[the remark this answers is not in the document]
anhuong: We hit into an issue previously when using copytree ([internal link](https://github.ibm.com/ai-foundation/watson-fm-stack-tracker/issues/1089)) where the permissions and ownership of the directory was being copied which caused problems with error `Operation not permitted'`

tests/utils/test_merge_model_utils.py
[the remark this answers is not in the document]
anhuong: I see Angel's note about about skipping it in Github Actions but I was able to run the `get_tensors()` command locally on my machine and run the `test_merge_model_utils.py` unit tests on my machine (removing the cuda skipifs)

tests/utils/test_merge_model_utils.py
```diff
@@ -0,0 +1,139 @@
...
+        framework="pt",
+    ) as f:
+        for k in f.keys():
+            if "lm_head.weight" in k:
+                found_lm_head = True
+    assert found_lm_head
+
+    # do the post processing
+    with tempfile.TemporaryDirectory() as tempdir:
+        shutil.copytree(DUMMY_TUNED_LLAMA_WITH_ADDED_TOKENS, tempdir)
```
anhuong, line 123: When I ran this locally, it failed for me because tempdir already existed. I had to add `dirs_exist_ok=True` for this to pass

tuning/utils/merge_model_utils.py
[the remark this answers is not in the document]
anhuong: We do have a similar function [here](https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/build/utils.py#L27) that recursively copies all of the contents of a directory but it doesn't have the exclude and copies the whole directory instead of just the files which is what this method is doing

tests/utils/test_merge_model_utils.py
Ssukriti: Could you commit that please @anhuong ? I was planning to run the test myself today, but since you already ran it

tuning/sft_trainer.py
Ssukriti: I feel there is often some confusion here, this piece of code was taken from openinstruct https://github.com/allenai/open-instruct/blob/main/open_instruct/finetune.py#L635 , if you see they have mentioned that though the tokens are already in model, they are not added as special_tokens. This can have some implications in quality as models treat special tokens different from regular tokens - do not split it etc. Hence this change is just making those tokens special tokens instead and not adding any new tokens. 

Num_added_tokens will just eb the pad token

Ssukriti: pad token is always needed to be added for any model architecture - that is a open source requirement by SFT Trainer, without which training will not proceed. Hence we always add pad token for all architectures here https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/tuning/sft_trainer.py#L278 (except granite which already has pad token)

Hence minimum 1 token is always added and hence post processing is needed

Ssukriti:
@ashokponkumar as explained on slack, we cannot remove addition of pad token as it is a rwquirement and known in open source. Without which we will run into this error https://stackoverflow.com/questions/70544129/transformers-asking-to-pad-but-the-tokenizer-does-not-have-a-padding-token

Most models in open source do not have pad token - new llama3.1, llama3, llama, allam, mixtral, mistral . Hence we have to add minimal 1 token for all these architectures, which we do in generic manner 'if pad token is None, set it'

This PR is thus doing the post-processing needed to handle addition of any token for LoRA inferencing on vLLM. Without this PR, LoRA inference on vLLM does not work for any of above architectures. 

Even if we remove other tokens, like unk etc - **which can be done in following PR and issue** , the change is still needed for pad token. Hence better to keep the change generic and return number of added tokens from code

anhuong (approved):
LGTM 🚀

PR #341, opened 2024-09-13
ci: Add dev deps and eval scripts to Dockerfile

aluu317:
### Was the PR tested

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -149,10 +149,9 @@ RUN if [[ "${ENABLE_AIM}" == "true" ]]; then \
     fi
 
 RUN if [[ "${ENABLE_DEV}" == "true" ]]; then \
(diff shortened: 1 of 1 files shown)


anhuong:
Interesting strategy, this would allow us to build dev images by setting `ENABLE_DEV` although the common way of creating a dev image is to create a dev stage as Alex has done in this PR -- https://github.com/foundation-model-stack/fms-hf-tuning/pull/283. Probably better to have the dev stage so that it can also include dev tools like git and microdnf/yum

aluu317:
@anhuong Sure, I like the dev stage approach as well. Looks like that PR also installs vscode for debugging. I see you started a review there, I can close this one and we can see about getting that other PR merged in?

aluu317:
Looks like PR #283 will cover adding a dev image

PR #343, opened 2024-09-17
fix: Addition of default pad token in tokenizer when EOS and PAD token are equal

Abhishek-TAMU:
### Description of the change

For any model when tokenizer has PAD token equal to EOS token, then the `DataCollatorForCompletionOnlyLM` masks the EOS token from `labels` of dataset batch which leads to repetitive `predicted_target` during inference time thus decreasing f1-micro for models like `granite-34b-gptq` 

### Related issue number

[#1309 ](https://github.ibm.com/ai-foundation/watson-fm-stack-tracker/issues/1309)

### How to verify the PR

Running HF inference after QLoRA tuning of `granite-34b-gptq` model. 

### Was the PR tested

### DIFF
--- a/tuning/sft_trainer.py
+++ b/tuning/sft_trainer.py
@@ -262,6 +262,9 @@ def train(
         if tokenizer.unk_token is None:
             logger.warning("UNK token set to default, missing in tokenizer")
             special_tokens_dict["unk_token"] = configs.DEFAULT_UNK_TOKEN
+        if tokenizer.pad_token == tokenizer.eos_token:
+            logger.warning("PAD token set to default, different from eos token")
+            special_tokens_dict["pad_token"] = configs.DEFAULT_PAD_TOKEN
 
     # TODO: lower priority but understand if resizing impacts inference quality and why its needed.
     # It makes sense if we manipulate tokenizer that we also save it and provide it to inference.
--- a/tuning/sft_trainer.py
+++ b/tuning/sft_trainer.py
@@ -263,8 +263,13 @@ def train(
             logger.warning("UNK token set to default, missing in tokenizer")
             special_tokens_dict["unk_token"] = configs.DEFAULT_UNK_TOKEN
         if tokenizer.pad_token == tokenizer.eos_token:
-            logger.warning("PAD token set to default, different from eos token")
-            special_tokens_dict["pad_token"] = configs.DEFAULT_PAD_TOKEN
+            logger.warning(
+                "PAD token set to default, to make it different from eos token"
+            )
+            if tokenizer.eos_token != configs.DEFAULT_PAD_TOKEN:
+                tokenizer.pad_token = configs.DEFAULT_PAD_TOKEN
+            else:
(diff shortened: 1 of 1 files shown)


Ssukriti (approved):

Ssukriti:
Thank you @Abhishek-TAMU . Lets wait for another review by @kmehant since he was closely involved with the tokens update.

tuning/sft_trainer.py
```diff
@@ -262,6 +262,9 @@ def train(
         if tokenizer.unk_token is None:
             logger.warning("UNK token set to default, missing in tokenizer")
             special_tokens_dict["unk_token"] = configs.DEFAULT_UNK_TOKEN
+        if tokenizer.pad_token == tokenizer.eos_token:
+            logger.warning("PAD token set to default, different from eos token")
+            special_tokens_dict["pad_token"] = configs.DEFAULT_PAD_TOKEN
```
kmehant, line 267: @Ssukriti @Abhishek-TAMU what if the eos_token in tokenizer is actually the DEFAULT_PAD_TOKEN? Can we handle that case?

kmehant: example - 
tokenizer.eos_token is configs.DEFAULT_PAD_TOKEN
and 
tokenizer.pad_token is configs.DEFAULT_PAD_TOKEN

At this point, we are setting tokenizer.pad_token  to configs.DEFAULT_PAD_TOKEN this again. We might want to think about this case.

Abhishek-TAMU: Good point. In this case can we modify and assign `eos_token` as `configs.DEFAULT_EOS_TOKEN` ? @Ssukriti

Ssukriti: its very unlikely that EOS token will be <PAD> no?

tuning/sft_trainer.py
```diff
@@ -262,6 +262,9 @@ def train(
         if tokenizer.unk_token is None:
             logger.warning("UNK token set to default, missing in tokenizer")
             special_tokens_dict["unk_token"] = configs.DEFAULT_UNK_TOKEN
+        if tokenizer.pad_token == tokenizer.eos_token:
+            logger.warning("PAD token set to default, different from eos token")
```
Ssukriti, line 266:
```suggestion
            logger.warning("PAD token set to default, to make it different from eos token")
```

tuning/sft_trainer.py
Ssukriti: but ya if we want to be absolutely sure, we can do that

if tokenizer.pad_token == tokenizer.eos_token:
     if tokenizer.eos_token != configs.DEFAULT_PAD_TOKEN:
            tokenizer.pad_token  = configs.DEFAULT_PAD_TOKEN
     else:
           tokenizer.eos_token = configs.DEFAULT_EOS_TOKEN

Abhishek-TAMU:
Pushed the suggested changes. Thank you!
@kmehant @Ssukriti

kmehant (approved):

PR #348, opened 2024-09-25
refactor: move tokenizer_data_utils with the rest of utils, add further unit testing.

willmj:
### Description of the change

Move `tokenizer_data_utils.py` from `/data` to `/utils` with the rest of the utils.
Update imports so function calls change from `tokenizer_data_utils.tokenizer_and_embedding_resize` to `tokenizer_and_embedding_resize`.
Add 3 unit tests:

   - Ensure adding special tokens works correctly
   - Ensure not adding special tokens doesn't modify tokenizer
   - Ensure input and output embeddings are resized properly

### How to verify the PR
`tox -e py`

### Was the PR tested


aluu317:
Can you resolve the conflicts? I think it might need some code changes

tests/utils/test_embedding_resize.py
```diff
@@ -74,3 +75,65 @@ def test_output_unaltered_across_embedding_resizes():
...
+        model=model,
+        multiple_of=1,
+    )
+
+    assert "<SEP>" in tokenizer.get_vocab()
+    assert "<PAD>" in tokenizer.get_vocab()
+
+    output_tokenizer_len = len(tokenizer.get_vocab())
+
+    assert output_tokenizer_len == input_tokenizer_len + 2
```
aluu317, line 100: Because the new return of of `tokenizer_and_embedding_resize`, I think we can test it even more thorough here with the returned value which is `{"num_new_tokens": num_new_tokens, "new_embedding_size": embedding_size}`.
We can assign an output above:
```
resize_results = tokenizer_and_embedding_resize(
        special_tokens_dict=special_tokens,
        tokenizer=tokenizer,
        model=model,
        multiple_of=1,
    )
 ```
and then here, we can do:
```
assert output_tokenizer_len - input_tokenizer_len == resize_results["num_new_tokens"]
```

File: tests/utils/test_embedding_resize.py [hunk 2]
```diff
@@ -74,3 +75,65 @@ def test_output_unaltered_across_embedding_resizes():
...
+
+    input_tokenizer_len = len(tokenizer.get_vocab())
+
+    tokenizer_and_embedding_resize(
+        special_tokens_dict={}, tokenizer=tokenizer, model=model, multiple_of=1
+    )
+
+    output_tokenizer_len = len(tokenizer.get_vocab())
+
+    assert input_tokenizer_len == output_tokenizer_len
```
aluu317, line 121: in addition, we can do another assert here `resize_result[num_new_tokens] == 0` as well

tests/utils/test_embedding_resize.py
```diff
@@ -74,3 +75,65 @@ def test_output_unaltered_across_embedding_resizes():
...
+
+def test_resize_with_multiple_of():
+    tokenizer = AutoTokenizer.from_pretrained(MODEL_NAME)
+    model = AutoModelForCausalLM.from_pretrained(MODEL_NAME)
+
+    tokenizer_and_embedding_resize(
+        special_tokens_dict={}, tokenizer=tokenizer, model=model, multiple_of=8
+    )
+
+    assert model.get_input_embeddings().embedding_dim % 8 == 0
```
aluu317, line 138: do we need to make sure that the input embeddings has multiple of 8?

tests/utils/test_embedding_resize.py
```diff
@@ -74,3 +75,65 @@ def test_output_unaltered_across_embedding_resizes():
...
+    assert model.get_output_embeddings().out_features % 8 == 0
```
aluu317, line 139: this is interesting. Is `out_features` the same as embedding size?
We can test here as well that `resize_result["new_embedding_size"] %8 ==0` tho

tests/utils/test_embedding_resize.py
aluu317: Yes, we do. Pls ignore my silly question

aluu317 (approved):
LGTM! Thank you

aluu317 (approved):
LGTM

PR #352, opened 2024-09-26
fix: unable to find output_dir in multi-GPU during resume_from_checkpoint check

Abhishek-TAMU:
### Description of the change

Addition of code to create `output_dir` in `accelerate_launch.py` if it doesn't exist.

### Related issue number

[#1352](https://github.ibm.com/ai-foundation/watson-fm-stack-tracker/issues/1352)

### How to verify the PR

Run full fine tuning or LoRA tuning with multiple GPUs and check if issue exists.

### Was the PR tested

### DIFF
--- a/build/accelerate_launch.py
+++ b/build/accelerate_launch.py
@@ -98,7 +98,7 @@ def main():
     #
     ##########
     output_dir = job_config.get("output_dir")
-    if not os.path.exists(output_dir): 
+    if not os.path.exists(output_dir):
         os.makedirs(output_dir)
     try:
         # checkpoints outputted to tempdir, only final checkpoint copied to output dir
--- a/build/accelerate_launch.py
+++ b/build/accelerate_launch.py
@@ -98,6 +98,8 @@ def main():
     #
     ##########
     output_dir = job_config.get("output_dir")
+    if not os.path.exists(output_dir): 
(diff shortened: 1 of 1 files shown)


build/accelerate_launch.py
```diff
@@ -98,6 +98,8 @@ def main():
     #
     ##########
     output_dir = job_config.get("output_dir")
+    if not os.path.exists(output_dir):
```
anhuong, line 101: Thanks Abhishek! This looks great! I think this solves the main case with running multi-GPU training from the image. But this issue would still appear if someone ran `accelerate launch tuning/sft_trainer.py` on their own but unsure if creating the directory within sft_trainer.py makes sense or if the same issue would come up where the dir may not be created.... 

After talking to Abhishek about this, adding this in before the separate process are initiated ensures the output_dir exists. 

> If we want only one process to check the resume from checkpoint condition then we can maybe add the check if dist.rank() == 0 to process resume_from_checkpoint similar to this [case](https://ibm-research.slack.com/archives/C06MZPQ4ENL/p1726854004617029?thread_ts=1726792939.729799&cid=C06MZPQ4ENL)

But we want to check with others on this.

So for now, we are adding this improvement that fixes the bug and can look into the alternative as a future improvement.

anhuong (approved):

PR #360, opened 2024-10-01
fix: Revert changes from PR#326 with trl version change

willmj:
### Description of the change
Setting trl to version 0.9.6 causes the following error:
```
if not args.packing:
            # If we aren't skipping data preparation, then a dataset_text_field
            # or formatting_func must be provided.
            if (
                args.dataset_text_field is None
                and formatting_func is None
                and dataset_kwargs is not None
                and "skip_prepare_dataset" in dataset_kwargs
                and dataset_kwargs["skip_prepare_dataset"]
            ):
>               raise ValueError(
                    "You passed `packing=False` to the SFTTrainer/SFTConfig, but you didn't pass a `dataset_text_field` or `formatting_func` argument."
                )
E               ValueError: You passed `packing=False` to the SFTTrainer/SFTConfig, but you didn't pass a `dataset_text_field` or `formatting_func` argument.
```

This PR is to temporarily revert the changes in PR #326 until Fabian's qlora fix is in the latest version of trl

### Was the PR tested

### DIFF
--- a/tuning/sft_trainer.py
+++ b/tuning/sft_trainer.py
@@ -67,7 +67,6 @@ from tuning.utils.logging import set_log_level
 from tuning.utils.preprocessing_utils import (
     format_dataset,
     get_data_collator,
-    is_pretokenized_dataset,
     validate_data_args,
 )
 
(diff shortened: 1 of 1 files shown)


fabianlim:
@willmj what you pasted seems strange

The comments say that the raise is only when data prep is not skipped
```
# If we aren't skipping data preparation, then a dataset_text_field
 # or formatting_func must be provided.
```

But the code triggers when `dataset_kwargs["skip_prepare_dataset"]` is truthful, which is a contradiction because it should mean to skip the data prep

willmj:
@fabianlim misunderstood your original comment in #360, closing this PR

PR #361, opened 2024-10-01
ci: run unit tests, fmt, image build on release branch

anhuong:
### Description of the change

We want to run CI automation on PRs that go directly to release branch. This includes any patch fixes (currently) and the release time PRs. By not running unit tests, we miss differences between main and release branch as well as dependencies may be new and updated that should be tested. This vets the PR at release time. 

### Was the PR tested

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -16,9 +16,9 @@ name: Format
 
 on:
   push:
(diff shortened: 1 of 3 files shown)


willmj (approved):
LGTM!

PR #362, opened 2024-10-01
DO NOTE MERGE: test updating transformers

anhuong:
expect unit tests to fail, testing that CI automation runs as expected


anhuong:
Did not trigger the CI automation with PR along and github action update in main branch. After adding the github workflows change to this branch, it triggered the automation. So unit tests will run on the next release and then all future PRs into release branch after the github action change has been added to release branch

Issue #359, opened 2024-10-01
bug: crash when the `--output_dir` doesn't exist and multiple GPUs (processes) are used.


HarikrishnanBalagopal:
# Overview

Race condition? leading to a crash when multiple GPUs (processes) are used and the output directory doesn't exist.

## Steps to reproduce

Run a multiple GPU job with `torchrun` and a non-existent output directory.

```shell
torchrun --nnodes=1 --nproc-per-node=8 --rdzv_id=101 --rdzv_endpoint=127.0.0.1:12345 -m tuning.sft_trainer --fsdp full_shard auto_wrap --fsdp_config=fsdp_config.json --model_name_or_path=TinyLlama/TinyLlama-1.1B-step-50K-105b --training_data_path=TIGER-Lab/SKGInstruct-skg-only --output_dir=train_output --gradient_checkpointing=false --data_formatter_template=Question:\\n\\n\\nAnswer:\\n --response_template=\\n\\nAnswer:\\n --log_level=debug --tracker=aim --aim_repo=/aim/aimrepo --experiment=test-1 --max_steps=100 --logging_strategy=steps --logging_steps=1 --save_strategy=steps --save_steps=50 --learning_rate=1e-06
```

## Actual Output

```shell
  warnings.warn(
/proj/data-eng/haribala/conda/envs/platform/lib/python3.11/site-packages/transformers/training_args.py:2007: FutureWarning: `--push_to_hub_token` is deprecated and will be removed in version 5 of 🤗 Transformers. Use `--hub_token` instead.
  warnings.warn(
WARNING:sft_trainer.py:dataset kwargs None
/proj/data-eng/haribala/conda/envs/platform/lib/python3.11/site-packages/huggingface_hub/utils/_deprecation.py:100: FutureWarning: Deprecated argument(s) used in '__init__': dataset_text_field, max_seq_length. Will not be supported from version '1.0.0'.

Deprecated positional argument(s) used in SFTTrainer, please use the SFTConfig to set these arguments instead.
  warnings.warn(message, FutureWarning)
/proj/data-eng/haribala/conda/envs/platform/lib/python3.11/site-packages/trl/trainer/sft_trainer.py:280: UserWarning: You passed a `max_seq_length` argument to the SFTTrainer, the value you passed will override the one in the `SFTConfig`.
  warnings.warn(
[...]

PR #364, opened 2024-10-02
fix: crash when output directory doesn't exist

HarikrishnanBalagopal:
### Description of the change

Fixes crash in https://github.com/foundation-model-stack/fms-hf-tuning/issues/359

### Related issue number

https://github.com/foundation-model-stack/fms-hf-tuning/issues/359

### How to verify the PR

Run a multi GPU training with a non-existent output dir.

### Was the PR tested

### DIFF
--- a/tuning/sft_trainer.py
+++ b/tuning/sft_trainer.py
@@ -637,6 +637,9 @@ def main():
     combined_tracker_configs.file_logger_config = file_logger_config
     combined_tracker_configs.aim_config = aim_config
 
+    if training_args.output_dir:
+        os.makedirs(training_args.output_dir, exist_ok=True)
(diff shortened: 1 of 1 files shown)


kmehant:
FYA - @anhuong  https://github.com/foundation-model-stack/fms-hf-tuning/issues/359#issue-2558689388

Abhishek-TAMU:
If race condition is already tested as per [this](https://github.com/foundation-model-stack/fms-hf-tuning/pull/365#issuecomment-2391049004), then it looks good to me.

anhuong:
Abhishek additionally tested this in the image with the accelerate_launch.py script which worked nicely as well

anhuong (approved):

PR #374, opened 2024-10-17
docs: Data Preprocessor ADR

dushyantbehl:
### Description of the change

This PR adds an ADR for the newly proposed changes to the data pre procesor implementation in fms-hf-tuning.

### DIFF
--- a/architecture_records/004-datapreprocessor.md
+++ b/architecture_records/004-datapreprocessor.md
@@ -1,0 +1,421 @@
+# Data Pre Processor Design For fms-hf-tuning
+
+**Deciders(s)**: Sukriti Sharma (sukriti.sharma4@ibm.com), Will Johnson (Will.Johnson@ibm.com) , Abhishek Maurya (maurya.abhishek@ibm.com), Yu Chin Fabian Lim (flim@sg.ibm.com), Dushyant Behl (dushyantbehl@in.ibm.com), Ashok Pon Kumar (ashokponkumar@in.ibm.com)
+
+**Date (YYYY-MM-DD)**:  2024-03-06
+
+**Obsoletes ADRs**:  NA
+
+**Modified By ADRs**:  NA
+
+**Relevant Issues**: [1]
+
+- [Summary and Objective](#summary-and-objective)
+  - [Motivation](#motivation)
+  - [User Benefit](#user-benefit)
+- [Decision](#decision)
+  - [Alternatives Considered](#alternatives-considered)
+- [Consequences](#consequences)
+- [Detailed Design](#detailed-design)
+
+## Summary and Objective
+
+The primary objective of the `DataPreProcessor` design for fms-hf-tuning is to provide a unified yet powerful interface for handling diverse data formats and configurations.
+This interface should cater to various user expertise levels, enabling basic users to easily load and process data, while allowing advanced users to customize their data handling.
+
+### Key Goals:
+1. **Broad Data Format Support**: Allow datasets in formats such as Arrow, Parquet, and CSV.
+1. **Compatibility with Multiple Datasets and Files**: Enable multiple files per dataset and interleaving or mixing of datasets.
+1. **Support for Different Data Modalities**: Include images, audio, and text data, along with modality-specific preprocessing options.
+1. **User-Focused Configurations**: Provide simple data loading for regular users, while enabling advanced configurations for expert users.
+1. **Template-Based Preprocessing**: Support jinja template rendering, where necessary, for template-dependent preprocesing requirements.
+
+### Motivation
+
+The main motivation for this ADR stems from the fact that fms-hf-tuning is being used by many teams for a diverse set of use cases which are not currently supported in the library. To be precise, currently in the library for data preposssing we currently take two primary arguments `training_data_path` and `validataion_data_path` which take in a single file location for a dataset.
+A user can currently pass in
+1. a pretokenized json(l) dataset via 
+   ```
+   --training_data_path <pretokenized dataset>
+   ```
+1. a preprocessed json(l) formats with a single sequence and a specified `response_template` to use for masking on completion.
+    ```
+    --training_data_path <dataset.json> --dataset_text_field <e.g. 'input'> --response_template <e.g. '\n### Label:'>
+    ```
+1. a json(l) dataset and a `data_formatter_template` to use the formatting function on fly.
+    ```
+    --training_data_path <dataset.json> --data_formatter_template <'template'>
+    ```
+1. a json(l) dataset with `input` and `output` fields, names hardcoded and cannot be changed.
+    ```
+    --training_data_path <non-pretokenized dataset with input and output fields>.json
+    ```
+
+The first motivation for a change is the requirements from users asking for different data formats, current code only supports json while there are teams which are training using Parquet and Arrow format so they require additional data format support.
+
+Also use cases from teams require multiple datasets and even multiple data files in a dataset.
+
+Futher requirements from teams is to have a way to interleave datasets at run time by specifying static weights to mix different datasets which is also not supported by the code yet.
+
+Finally other requirements are to have preprocesing support for multiple modalities of data (starting with Image first) and have support for advanced preprocesing like jinja based template rendering of the dataset before consumption.
+
+All these requirements are new and are currently not supported by the library which motivated us to propose a change in the design of data preprocesing in this library to incorporate these and potentially any new changes in one go.
+
+### User Benefit
+
+Users will benefit from the additional argument which allows them to pass a single [`data_config`](#our-considerations-for-the-design) file specifying how to preprocess their dataset.
+Our data config file will extend users the capabilities to,
+1. Pass multiple data files and multiple datasets.
+1. Specify static weights in the configuration to interleave datasets.
+1. Define preprocessing routines to apply on the data and in which order
+
+This will make the process of handling custom datasets which might require rendering jinja template or processing image data way much easier.
+
+We do not require users to learn the specification of the additional `data_config` file, as the existing arguments to process dataset which are present in the code [`tuning.config.configs.DataArguments`](https://github.com/foundation-model-stack/fms-hf-tuning/blob/398c2a8fe26d734344240555585d95e05299faa8/tuning/config/configs.py#L67) will not be deprecated in this version and users can keep using the same data arguments for use cases being served by the library.
+
+At the very least a user not well versed with the `data_config` will be able to pass in for e.g. a pre-tokenized pyarrow dataset 
+
+```
+--training_data_path <pre-tokenized>.pyarrow
+```
+
+And at full length they can specify multi-file, multi dataset configuration which can process dataset according to the config specified like,
+
+
+```
+...
+  - name: dataset1
+    sampling:
+      ratio: 0.3
+    data_paths:
+      - /data/stackoverflow-kubectl_posts
+      - /data/stackoverflow-kubernetes_posts
+      - /data/stackoverflow-openshift_posts
+    data_handlers:
+      - name: render_template
+        arguments:
+          remove_columns: all
+          batched: false
+          fn_kwargs:
+            jinja_template: "{<jinja-template>}"
+```
+A small part of the `data-config` spec which is provided in detailed in the section below.
+
+## Decision
+
+Some terminology before we move ahead 
+
+<table>
+  <tr>
+    <th style="border: 1px solid black; padding: 5px;">User Persona</th>
+    <th style="border: 1px solid black; padding: 5px;">Description</th>
+  </tr>
+  <tr>
+    <td style="border: 1px solid black; padding: 5px;">Simple User</td>
+    <td style="border: 1px solid black; padding: 5px;">A user who uses this library to train models using a single dataset, passing it via a single command line argument.</td>
+  </tr>
+  <tr>
+    <td style="border: 1px solid black; padding: 5px;">Advanced User</td>
+    <td style="border: 1px solid black; padding: 5px;">A user with a deep understanding of datasets, who knows how to apply specific preprocessing and mixing techniques during training.</td>
+  </tr>
+  <tr>
+    <td style="border: 1px solid black; padding: 5px;">Intermediate User</td>
+    <td style="border: 1px solid black; padding: 5px;">A user who works with custom datasets but lacks full knowledge of data processing, relying on advanced users for guidance to fulfill their use case.</td>
+  </tr>
+</table>
+
+Please note that most of the users of product here would fall into the simple user category while advanced and intermediate users are researchers looking to use our library for diverse set of use cases.
+
+### Our considerations for the design
+
+1. Allow advanced users to use full power of the HuggingFace library as much as possible without recreating the same.
+1. Allow advanced users to specify custom data preprocessor pipeline in an easy way.
+1. Ensure the single design can handle these and many more use cases without major changes.
+1. Design for Advanced users while simplify for simple users.
+
+We propose to allow advanced users to specify a full spec which exposes data preprocessing API provided by the HF library directly to them to be able to fully utilise the interface. 
+
+The proposed input spec which user specifies as `data_config` on how to pass information for such preprocessing is
+
+```
+datapreprocessor:
+    type: default
+datasets:
+  - name: dataset1
+    sampling:
+      ratio: 0.3
+    data_paths:
+      - /data/stackoverflow-kubectl_posts
+      - /data/stackoverflow-kubernetes_posts
+      - /data/stackoverflow-openshift_posts
+    data_handlers:
+      - name: tokenize
+        arguments:
+          remove_columns: all
+          batched: false
+  - name: dataset2
+    sampling:
+      ratio: 0.4
+    data_paths:
+      - /data/stackoverflow-kubectl_posts
+      - /data/stackoverflow-kubernetes_posts
+      - /data/stackoverflow-openshift_posts
+    data_handlers:
+      - name: render_template
+        arguments:
+          remove_columns: all
+          batched: false
+          fn_kwargs:
+            jinja_template: "{<jinja-template>}"
+      - name: tokenize
+        arguments:
+          remove_columns: all
+          batched: false
+```
+
+To iterate again, here our goal is not to reimplement the functionality provided by HuggingFace but rather have a clean interface using a config where advanced users can use things like Iterable datasets or Interleaving datasets and perform custom preprocessing like applying jinja templates etc in an easy way.
+
+In this spec, at top level we have the `Dataprocessor` config which contains just one field `type` which is set to `default`. This is done to ensure any future top level `dataprocessor` configs will go into this block. Users need not touch or provide this as the `default` is automatically selected.
+
+The second block here is where users will list multiple `datasets` and each dataset will contain information on how to process it. We allow arguments like `sampling` for users to specify sampling ratios while [`interleaving datasets`](https://huggingface.co/docs/datasets/en/process#interleave) to use API like [`interleave_datasets`](https://huggingface.co/docs/datasets/v3.1.0/en/package_reference/main_classes#datasets.interleave_datasets) by HuggingFace.
+
+The most powerful feature of this block is `data_handlers`. Here we allow users to specify a list of routines to apply on the dataset at the time of preprocessing. A `data_handler` is a [`map`](https://huggingface.co/docs/datasets/en/process#map) operation performed on the dataset to which a user can further pass informational arguments. We expose the full set of arguments of HF [`Dataset.map`](https://huggingface.co/docs/datasets/v3.1.0/en/package_reference/main_classes#datasets.Dataset.map) operation here to the user as `kwargs` of a handler.
+
+As example in `dataset2` the data handler is requesting to apply a `render_template` function before tokenization on the dataset which processes the dataset and renders the `jinja template` specified as `fn_kwargs.jinja_template`, rest of the arguments like `remove_column` and `batched` are just HF Map API arguments.
+
+```
+- name: dataset2
+    sampling:
+      ratio: 0.4
+    data_paths:
+      - /data/stackoverflow-kubectl_posts
+      - /data/stackoverflow-kubernetes_posts
+      - /data/stackoverflow-openshift_posts
+    data_handlers:
+      - name: render_template
+        arguments:
+          remove_columns: all
+          batched: false
+          fn_kwargs:
+            jinja_template: "{<jinja-template>}"
+      - name: tokenize
+        arguments:
+          remove_columns: all
+          batched: false
+```
+
+By allowing the users to specify data handlers like this we allow them to use full Hugging Face API and at the same time specify preprocessing routines in a fixed order. The handlers list specify a [`DAG`](https://en.wikipedia.org/wiki/Directed_acyclic_graph) of operations to apply on the dataset and will be executed by the code in that order.
+
+Furthermore this design allows flexibility to be extended to any upcoming usecase because any operation to be executed on the dataset can be broken down into function execution implemented as data handlers.
+
+This makes our spec a complete solution for advanced users of the library allowing them to specify complete preprocessing operations to be applied to the dataset via a config file.
+
+Finally, with this spec we do not want to break the functionality for the simple users of the library. A simple user which wants to just use the library with a single dataset like today can pass the same dataset via `--training_data_path <file> --validataion_data_path <file>` arguments.
+
+Infact we do not change the behaviour currently supported by any of the `tuning.config.configs.DataArguments` arguments hence allowing the simple users of the library to continue using the library as is.
+
+### Performance Considerations
+
+Since this design allows complex preprocessing of the dataset on fly, the design should incorporate perfomrance measures to ensure that the system is not performing too slow or spending too much time while preprocessing the dataset to affect tuning/training time.
+
+The goal that we have here is to not be slower than the HuggingFace library which our whole design is based upon, in this sense we also imagine any performance improvements that we come across to be contributed back to HF library to keep our design simple and not reimplement stuff.
+
+-> Handling Large Dataset
+
+Our main reason for using HF [Map](https://huggingface.co/docs/datasets/en/process#map) heavily for data preprocessing is that for large datasets which are generally loaded as `IterableDatasets` the MAP API automatically performs [`lazy map operations`](https://huggingface.co/docs/datasets/en/about_mapstyle_vs_iterable#eager-data-processing-and-lazy-data-processing) and hence doesn't produce too much overhead while training.
+
+-> Caching intermediate dataset
+
+Hugging Face caches intermediate map operations which makes replay of our data preprocessor easier if same map parameters and operations are applied. If the file system is an issue we have two considerations,
+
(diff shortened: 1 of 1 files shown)


architecture_records/004-dataloader-v2.md
```diff
@@ -0,0 +1,314 @@
...
+1. Implementing data collators as map.
+    We explore this approach as all the processing of data by us is done in data handlers so to implement collators as data handlers 
+    makes sense but is a hard problem to solve. Data collation happens at the very last stage of batch creation and is done in an async
+    fashion because event the HF data collators like `DataCollatorForCompletionOnlyLM` or [`DataCollatorForSeq2Seq`](https://github.com/huggingface/transformers/blob/3f06f95ebe617b192251ef756518690f5bc7ff76/src/transformers/data/data_collator.py#L543) are passed to torch data loader by the HF APIs and is executed by the torch Dataloader.
+
+    So implementing data collators as map means we need to disable [torch data collation and batching](https://pytorch.org/docs/stable/data.html#disable-automatic-batching) and perform it in our code, which we do not intend to experiment with due to its complexity.
+
+1. Passing collators directly to SFTTrainer.
+
+    In the code collators are collected by `get_data_collator` functionality and passed to `SFTTrainer`. We can retain the same functionality
```
Ssukriti, line 271: yes data collators have to be inferred for the dataset and arguments passed . it is better to infer it in code for labelled datasets

architecture_records/004-dataloader-v2.md
```diff
@@ -0,0 +1,314 @@
...
+`Dataset.from_generator(<generator>)` functionality.
+
+The reason to make use of HF factory is to keep our code simple and free of duplication with preexisting library code.
+
+This also means that any implementation like `get_json_object` which load `json(l)` and then return a custom json dict
+can be implemented as data handlers over basic json dataloader from HF.
+
+### Splitting and Interleaving datasets
+
+Other argument such as `splitter_arguments` can be passed to HF [`datasets.test_train_split`](https://huggingface.co/docs/datasets/v3.0.1/en/package_reference/main_classes#datasets.Dataset.train_test_split) to create a test/train split of
```
Ssukriti, line 242: is this for creating validation dataset ?

architecture_records/004-dataloader-v2.md
```diff
@@ -0,0 +1,314 @@
...
+The abstract data loader class provides a way to register datahandler against a `name` which is a string.
+The data handler config `DataHandlerConfig` taken by `execute_data_handlers` represents a DAG of data handling
+routines which are to be executed on the data. 
+
+For standard HF API you can think of these as the HF Processing routines. Which could be Map/Filter/Select operations
+We implement most of the routines as map and because of this even the tokenisation of data which is done today
+in fms-hf-tuning via `tuning/utils/preprocessing_utils.py::get_preprocessed_dataset` can be retained as a data 
+handler which performs tokenization.
+
+The implementation is flexible enough for users to specify their own implementation of data handling routines
```
Ssukriti, line 213: users of product - i.e watsonX are not going to select what data handler to apply . Using tuning type and type of dataset , proper defaults have to be selected - watsonX is also not going to figure out , setting defaults is on us.

architecture_records/004-dataloader-v2.md
```diff
@@ -0,0 +1,314 @@
...
+### Advantages
+
+### Impact on performance
+
+## Detailed Design
+
+The input spec which user specifies on how to pass information to such data loader is this
+
+```
+dataloader:
```
fabianlim, line 68: Its not clear to me if this spec will be the default once this feature is merged in. If so, i urge to consider to maintain also the simpler earlier interface where you only need to pass a data path in. So users who do not need to have so complicated dataloader requirements do not need to go through the trouble to write the spec

architecture_records/004-dataloader-v2.md
```diff
@@ -0,0 +1,314 @@
...
+
+For example see this implementation - https://github.ibm.com/ai4code-wisdom/platform/blob/main/modelops/modelops/train.py#L251
+
+# Implementation of HuggingFace Data Loader.
+
+HF Dataloader or a dataloader which used Hugging Face API is implemented as an instance of the `Dataloader` class.
+
+When the dataloader goes through each `DataSetConfig`
+
+The HF dataloader implements functionality to process different type of files via its `load_dataset` factory.
```
fabianlim, line 231: It is not clear to me fi the plan is to implement a new `DataLoader` abstraction or what. I feel taht from a data standpoint, the abstraction interface should be the `DataSet` not the `DataLoader`. THis is because in `accelerate`, the It has its own `DataLoader` implementation. You dont want to touch that.

architecture_records/004-dataloader-v2.md
```diff
@@ -0,0 +1,314 @@
...
+HF provides same APIs like `datasets.interleave_datasets` over the `Iterable` datasets as well.
+
+Further important thing to note is in case of HF, if we use hugging face the `map` functionality which we use to implement data handling
+is handled in a lazy fashion meaning we don't need to handle the data handlers in a different way for streaming data. [More Information on HF Page.](https://huggingface.co/docs/datasets/en/about_mapstyle_vs_iterable#eager-data-processing-and-lazy-data-processing)
+
+## Handling data collators.
+
+Data collators specifically for TRL use cases like chat based interactions which apply chat templates and proper attention masking on the
+tokenized data like in the case of `DataCollatorForCompletionOnlyLM` handle a specific functionality on the data. In this design we consider two approaces for data collators.
+
```
fabianlim, line 261: similarly, I feel you do not need to touch the `DataCollators` since they are part of the `DataLoader` abstraction.

architecture_records/004-dataloader-v2.md
```diff
@@ -0,0 +1,314 @@
...
+This means the image and audio multi modal datasets will be compatible with our data handler routines. Once we have the data handling routines set we will allow users to train with multi modal datasets too.
+
+# Stateful Dataloader implementation.
+
+Stateful dataloader refers to the `fms-fsdp` implementation of the dataloader by our colleagues Davis and Linsong.
+
+Supporting stateful dataloader will mean refactoring the fms-fsdp implementation into our abstract `Dataloader` class.
+
+In brief, things to consider here will be,
+1. Data handler support needs to be added to the stateful data loader as we want lazy execution of handlers (as and when data is loaded).
```
fabianlim, line 296: its aI think accelerate v1.0 should already have stateful dataloader implemented if I recall correctly?

architecture_records/004-dataloader-v2.md
fabianlim: https://github.com/huggingface/transformers/issues/31441

architecture_records/004-dataloader-v2.md
dushyantbehl: @fabianlim the idea is to keep the simpler interface while also have a detailed spec which can handle complex use cases and variour other requirements.

architecture_records/004-dataloader-v2.md
dushyantbehl: This plan is not to superimpose on the data loader implementations of accelerate or others rather just to say that this internal abstraction will use HF/Accelerate API while other can implement data loaders which are experimental but not yet available in HF. 

Thanks for the pointer on the `StatefulDataloader` we can definitely evaluate that for our consideration but the idea to have a design like this is to allow flexibility of implementing custom dataloaders like the one in our `fms-fsdp` repo inside `fms-hf-tuning`.

architecture_records/004-dataloader-v2.md
dushyantbehl: Yes, in case users want to split the data set via ratio.

architecture_records/004-dataloader-v2.md
dushyantbehl: Agreed on that front, hence we also prefer the approach where we use preimplemented collators and no reimplement or break any api.

architecture_records/004-dataloader-v2.md
fabianlim: I believe accelerate stance on the data loader is that it shouldn’t be user implemented. I feel you can cover most use cases by customising dataset, data collator, sampler or batch sampler. Do you have a use case where these are not sufficient

architecture_records/004-dataloader-v2.md
ashokponkumar: @dushyantbehl Is my understanding right that we can have a set of inbuilt data configs which people can trigger with say some flag? 

For example, we can have:
--data-config-name pretokenized
--data-config-data-path <path>

So that we don't have to maintain multiple data manipulation paths.

Ssukriti: I think it would be even better if defaults could be selected automatically for simple file formats and use cases - users enter multimodal JSON and default handler is selected . only for users that want to modify it / have a complex use case, we also expose a way to do that

architecture_records/004-dataloader-v2.md
```diff
@@ -0,0 +1,314 @@
...
+        arguments:
+          remove_columns: all
+          batched: false
+          fn_kwargs:
+            jinja_template: "{<jinja-template>}"
+  - name: dataset2
+    sampling:
+      ratio: 0.3
+    data_handlers:
+      - name: apply_tokenizer_chat_template
```
Ssukriti, line 108: @dushyantbehl @ashokponkumar doesnt SFT trainer add chat template on their own? https://huggingface.co/docs/trl/en/sft_trainer#dataset-format-support if dataset is of certain format . Chat is already supported by SFT trainer, since majority of your data handler examples given by you are applying templates, trying to understand use case

architecture_records/004-dataloader-v2.md
Ssukriti: use case of custom dataloader is also not clear to me. Seems like from use cases you defined so far, we just need
1. preprocessing functions for selected data types like parquet,chat etc
2. code for sampling and thus mixing different formats of datasets in one training 

They do not warrant a new dataLoader. Also to mention a big reason of using standard HF APIs in this repo was simplicity and ease of maintenance. With the direction being - any new logic should be contributed to HF , so we can leverage standard APIs.

architecture_records/004-dataloader-v2.md
```diff
@@ -0,0 +1,314 @@
...
+
+## Handling Multi Modal Data.
+
+HF does provide support for handling [image datasets](https://huggingface.co/docs/datasets/en/image_process) and [audio datasets](https://huggingface.co/docs/datasets/en/audio_load) which can be utilised by us in our HF dataloader.
+
+The functionality listed by HF in implementing the use of image and audio datasets is `map` based functions to perform resize, encoding and other such operations on the dataset (see the link above).
+
+This means the image and audio multi modal datasets will be compatible with our data handler routines. Once we have the data handling routines set we will allow users to train with multi modal datasets too.
+
+# Stateful Dataloader implementation.
```
Ssukriti, line 289: this section needs to be removed as agreed, passing custom DataLoader adds complexity - if it is merged to accelerate , please use standard ones

architecture_records/004-dataloader-v2.md
Ssukriti: This section needs to be removed as agreed - no customization of DataLoaders themselves

architecture_records/004-dataloader-v2.md
```diff
@@ -0,0 +1,314 @@
...
+
+    1. Different formats of data → Arrow, Parquet etc.
+
+    1. Different modalities of data -> Images, Audio etc.
+
+    1. Streaming data.
+
+    1. Data Replay
+
+    1. Resume with different number of GPUs
```
Ssukriti, line 35: how is this going to happen with preprocessing?

architecture_records/004-dataloader-v2.md
```diff
@@ -0,0 +1,314 @@
...
+
+    1. Async data loading
+
+1. Data Preprocessing - 
+
+    1. Custom Attn Masking
```
fabianlim, line 41: What is custom attn masking? attn masks are generally very slow and memory intensive, this is the reason why we use padding free where attention masks are not needed. So why do we want to use attention masks again?

architecture_records/004-dataloader-v2.md
```diff
@@ -0,0 +1,314 @@
...
+
+Further important thing to note is in case of HF, if we use hugging face the `map` functionality which we use to implement data handling
+is handled in a lazy fashion meaning we don't need to handle the data handlers in a different way for streaming data. [More Information on HF Page.](https://huggingface.co/docs/datasets/en/about_mapstyle_vs_iterable#eager-data-processing-and-lazy-data-processing)
+
+## Handling data collators.
+
+Data collators specifically for TRL use cases like chat based interactions which apply chat templates and proper attention masking on the
+tokenized data like in the case of `DataCollatorForCompletionOnlyLM` handle a specific functionality on the data. In this design we consider two approaces for data collators.
+
+1. Implementing data collators as map.
```
fabianlim, line 262: I feel one simplification, is that if the dataset is preprocessed, then just use a datacollator that collates all the tokens like a `DataCollatorForSeq2Seq`. 

Once data is processed, you no longer need a fancy data collator anymore. All it needs to do is form the batch

architecture_records/004-dataloader-v2.md
```diff
@@ -0,0 +1,314 @@
...
+The input spec which user specifies on how to pass information to such data loader is this
+
+```
+dataloader:
+    type: default(hfdataloader)/stateful(fms-fsdp implementation)
+    streaming: true
+datasets:
+  - name: dataset1
+    sampling:
+      ratio: 0.3
```
fabianlim, line 74: @dushyantbehl @Ssukriti I feel the better design is not to expose so many fine grained configurations to the user. Consider the the typical use case for loading datasets

```python
from datasets import load_dataset
data = load_dataset("alpaca")
```

When users load a dataset, they just want to use it, they dont really want to care to set complicated flags like "streaming: True". 

In my experience, dataset preperation is very specialized skill. If you think about ILAB, the data is prepared  by the Cambridge team and we just take the tokens and train with it. As such, a recipe is prepared and given to regular users who just deal with the preprocessed data. Simple things like choosing the wrong chat template can severely degrade performance, so regular users wont even mess with that.

Thus in fms-hf-tuning, there is no way to pass such a reciepe, because users will have to set their own `response_template`, etc. I feel that this ADR should aim to provide these reciepes

However what this ADR is doing now is focusing a lot on these fine-detailed configs, which by my argument, a regular user will not mess with. I think we can take inspiration with how HF does these things with fine-grained configs 

```python
from transformers import AutoModelForCausaLMl
model = transformers.AutoModelForCausalLM("....")
```
this actually downloads a `config.json` with complicated model settings that users dont really look at, but is very important to the ensuing that the model is correctly built.

So then the analogue that we might want is

```python
from fms_datasets import load_datasets

# train_dataset is a torch.Dataset that can just be passed to trainer, it
# should be already processed (or can be processed on the fly)
train_dataset = load_datasets('....")

# if train_dataset is multimodal, then some things like
# tokenizer.apply_chat_template(train_dataset, tokenize=False) may not make sense anymore
# so this has to be considered in the ADR
```
where you point to a path with a fine-grained config, and it does the data prep completely for the user. 

Therefore with this abstraction, someone well-versed in dataset prep, will then 
- prepare and publish the fine-grained configs, and then pass them to regular users to be used
- the publisher of such configs, could be IBM, or could be power-users in the public 

Some thoughts:
- performance is key. The more configs you give to the user, the more you need to think about how to store the intermediate results. If a dataset is not tokenized, then the tokenizer needs to run every time training starts. HF datasets caches this automatically. However, if chat templates are providable, and dataset mixing is allowable, then how are all of these going to be cached?
- If a dataset is big, then it should be handled automatically for the user without a user needing to set `streaming:true`
- If its multimodal the mixing should also be handling for the user, maybe allowing the user to select some high-level parameters like mixing constants
- chat templates, response templates, etc, any preprocessing config should be just written out in the fine-grained config. This is along the lines of "code-as-documentation".
- we should search out other open source repos that does the simimilar thing, and discuss why we cant use them
- we should call out why a standard HF dataset is insufficient, and why we need to extend (to a `DataLoader` as you call it)
- the ADR should call out high-level goals of the design, like "there is a need to distributed preprocessing recipes. In the "Summary and Objectives" when things like "support Arrow formats" is listed out -> it shoudl be clarified that this is not because HF datasets does not support arrow formats, but t is because fms-hf-tuning forces users to provide json datasets (i know this is kind of implied very late in the ADR).
- I feel we should not invent new abstractions, if a `DataLoader`, as you call it, can be abstracted by a `DataSet`, then its better to go with the latter, since everybody is familiar with it. Less abstractions, the more unstable the code.
- the concepts of `DataHandlerConfig, DataConfig, DataLoaderConfig` are completely internal, and to some extent, is of lesser importance to the ADR, whereas the higher level concerns, like usability and architectural considerations, need to be first ironed out and clarified before going into internals.

Ssukriti: I agree we need to talk about design more. I have expressed my dissatisfaction with it in some issues as well. We can discuss on a call next week

architecture_records/004-dataloader-v2.md
dushyantbehl: The implementation of data loader in the `fms-fsdp` supports resuming with different number of GPUs so when we add support for that dataloader we can pass additional args to the dataloader as the gpu number.

architecture_records/004-dataloader-v2.md
dushyantbehl: By custom attention masking we mean support for multi turn chat based data support which means custom part of input data to be masked or not...more likely done inside of the collators but keeping the design so we accept inputs accordingly to support such a use case. 

cc @ashokponkumar

fabianlim: @dushyantbehl can I clarify if you actually mean masking (which is to create an attention mask to say this token can only attend to this other token), or you mean loss masking which means to replace the losses with -100?

architecture_records/004-dataloader-v2.md
Ssukriti: @ashokponkumar can we please move away from providing default configs to auto detection for standard use cases? Providing default configs is still very messy and confusing . 

- for advanced users it is fine to allow to customize, but for straightforward use cases , we need to shortlist what data formats we want to support in repo - **accept datasets as one argument -> call preprocessing and training as needed** 
- for users that want to pass multiple datasets and need sampling , they can use the config file, but again specifying data preprocessing function in the config need not be compulsory?

architecture_records/004-dataloader-v2.md
```diff
@@ -0,0 +1,249 @@
...
+Ths motivation for this design is to implement the data preprocessor in a flexible way and expose to the users a powerful API which can allow them custom processing of data however they want, and at the same time have backwards compatibility retaining the simple interface of the existing API.
+
+### User Benefit
+
+### Simple User Perspective
+
+For simple and base users of the code we want to retain the same functionality wherever possible i.e,
+allow users to pass in a single data file and perform simple preprocessing.
+
+This means retaining the data arguments to the library which exist currently and ensuring that the appropriate processing required on the data is handeled internally.
```
Ssukriti, line 55:
```suggestion
This means retaining the data arguments to the library which exist currently and ensuring that the appropriate processing required on the data is handled internally. This interface will be extended to support more data formats and the default preprocessing function will be selected internally. 

We will shortlist the data formats we want to provide implicit support for, based on product use cases. 
```

architecture_records/004-dataloader-v2.md
```diff
@@ -0,0 +1,249 @@
...
+
+### Advanced User Perspective
+
+For advanced users we want to open up an argument to our library `data_config_file` which will take as input a data preprocesing config file specifying what preprocessing to apply on the data and in what order.
```
Ssukriti, line 59:
```suggestion
For advanced users who have custom preprocessing needs, we want to open up an argument to our library `data_config_file` which will take as input a data preprocesing config file specifying what preprocessing to apply on the data and in what order.
```

architecture_records/004-datapreprocessor.md
```diff
@@ -0,0 +1,421 @@
...
+
+This makes our spec a complete solution for advanced users of the library allowing them to specify complete preprocessing operations to be applied to the dataset via a config file.
+
+Finally, with this spec we do not want to break the functionality for the simple users of the library. A simple user which wants to just use the library with a single dataset like today can pass the same dataset via `--training_data_path <file> --validataion_data_path <file>` arguments.
+
+Infact we do not change the behaviour currently supported by any of the `tuning.config.configs.DataArguments` arguments hence allowing the simple users of the library to continue using the library as is.
+
+### Performance Considerations
+
+Since this design allows complex preprocessing of the dataset on fly, the design should incorporate perfomrance measures to ensure that the system is not performing too slow or spending too much time while preprocessing the dataset to affect tuning/training time.
```
willmj, line 219: nit: 
```suggestion
Since this design allows complex preprocessing of the dataset on fly, the design should incorporate performance measures to ensure that the system is not performing too slow or spending too much time while preprocessing the dataset to affect tuning/training time.
```

architecture_records/004-datapreprocessor.md
```diff
@@ -0,0 +1,421 @@
...
+
+The goal that we have here is to not be slower than the HuggingFace library which our whole design is based upon, in this sense we also imagine any performance improvements that we come across to be contributed back to HF library to keep our design simple and not reimplement stuff.
+
+-> Handling Large Dataset
```
willmj, line 223: nit:
```suggestion
#### Handling Large Dataset
```

architecture_records/004-datapreprocessor.md
```diff
@@ -0,0 +1,421 @@
...
+
+Our main reason for using HF [Map](https://huggingface.co/docs/datasets/en/process#map) heavily for data preprocessing is that for large datasets which are generally loaded as `IterableDatasets` the MAP API automatically performs [`lazy map operations`](https://huggingface.co/docs/datasets/en/about_mapstyle_vs_iterable#eager-data-processing-and-lazy-data-processing) and hence doesn't produce too much overhead while training.
+
+-> Caching intermediate dataset
```
willmj, line 227: nit:
```suggestion
#### Caching intermediate dataset
```

architecture_records/004-datapreprocessor.md
```diff
@@ -0,0 +1,421 @@
...
+Leaving all users to write their own preprocessing logic can also lead to code duplication across many teams which is something we want to avoid.
+
+More importantly as stated in the motivation we are getting ever increased demand from users who want to use this library directly with their dataset and have quick roundtrip for testing. This design allows users to specify simple parameters in the config and test for complex usecases easily.
+
+### Passing all datasets we take to the Huggingface SFTTrainer api and let it handle them without preprocessing at our end.
+
+Another alternative we have is to take the `dataset` input to this library and pass it directly to the trainer `SFTrainer` in our case directly and let it handle loading and preprocessing the dataset.
+
+[SFTrainer](https://huggingface.co/docs/trl/v0.12.1/en/sft_trainer#trl.SFTTrainer) supports specifying the `train_dataset` and `eval_dataset` for both of which it supports iterable datasets along with normal datasets allowing us to pass a large dataset supported via streaming.
```
willmj, line 251: nit:
```suggestion
[SFTTrainer](https://huggingface.co/docs/trl/v0.12.1/en/sft_trainer#trl.SFTTrainer) supports specifying the `train_dataset` and `eval_dataset` for both of which it supports iterable datasets along with normal datasets allowing us to pass a large dataset supported via streaming. 
```

architecture_records/004-datapreprocessor.md
```diff
@@ -0,0 +1,421 @@
...
+Leaving all users to write their own preprocessing logic can also lead to code duplication across many teams which is something we want to avoid.
+
+More importantly as stated in the motivation we are getting ever increased demand from users who want to use this library directly with their dataset and have quick roundtrip for testing. This design allows users to specify simple parameters in the config and test for complex usecases easily.
+
+### Passing all datasets we take to the Huggingface SFTTrainer api and let it handle them without preprocessing at our end.
+
+Another alternative we have is to take the `dataset` input to this library and pass it directly to the trainer `SFTrainer` in our case directly and let it handle loading and preprocessing the dataset.
```
willmj, line 249: nit:
```suggestion
Another alternative we have is to take the `dataset` input to this library and pass it directly to the trainer `SFTTrainer` in our case directly and let it handle loading and preprocessing the dataset.
```

architecture_records/004-datapreprocessor.md
```diff
@@ -0,0 +1,421 @@
...
+
+[SFTrainer](https://huggingface.co/docs/trl/v0.12.1/en/sft_trainer#trl.SFTTrainer) supports specifying the `train_dataset` and `eval_dataset` for both of which it supports iterable datasets along with normal datasets allowing us to pass a large dataset supported via streaming. 
+
+Please not that even in this case users will need to tell us that the dataset is large and is to be loaded via `streaming=True` because the argument which tells HF to load the dataset in iterable mode or standard mode is passed to [`load_dataset`](https://huggingface.co/docs/datasets/v3.1.0/en/package_reference/loading_methods#datasets.load_dataset)
```
willmj, line 253: nit:
```suggestion
Please note that even in this case users will need to tell us that the dataset is large and is to be loaded via `streaming=True` because the argument which tells HF to load the dataset in iterable mode or standard mode is passed to [`load_dataset`](https://huggingface.co/docs/datasets/v3.1.0/en/package_reference/loading_methods#datasets.load_dataset)
```

architecture_records/004-datapreprocessor.md
```diff
@@ -0,0 +1,421 @@
...
+
+```
+from datasets import load_dataset
+train_ds = load_dataset('imdb', split='train', streaming=True)
+```
+
+Additionally, `SFTrainer` has support for [data formatting function](https://huggingface.co/docs/trl/en/sft_trainer#dataset-format-support). Users can pass a `formatting_function` directly to `SFTtrainer` which formats the dataset for them,
```
willmj, line 260: nit:
```suggestion
Additionally, `SFTTrainer` has support for [data formatting function](https://huggingface.co/docs/trl/en/sft_trainer#dataset-format-support). Users can pass a `formatting_function` directly to `SFTTrainer` which formats the dataset for them,
```

architecture_records/004-datapreprocessor.md
```diff
@@ -0,0 +1,421 @@
...
+Taken from [HuggingFace docs](https://huggingface.co/docs/trl/en/sft_trainer#dataset-format-support)
+
+As our library is a wrapper on top of HF we cannot direclty allow users to pass a custom formatting function and
+our `data_handler` design can also support formatting dataset in a similar way to `formatting function` where users specify just name of the handler and we apply formatting on our end. The design for `data_handler` that we have is a superset of this feature which is more flexible and can support many more use cases.
+
+## Consequences
+
+### Arguments Required
+In this design, apart from the `data_config` spec users will also need to pass the `--response_template` argument. This is because the `DataCollator` functionality of this library is not being touched by our design.
+Also users need to specify `--dataset_text_field` which is infered from the `DataArguments` for now to ensure the simple interface remains same.
```
willmj, line 288: This is saying even if `--data_config` is passed in the user will also still need to pass in `--response_template` and `--dataset_text_field`?

dushyantbehl: Yes @willmj I updated the comment to be a bit more clear.

dushyantbehl:
> This is very clear and gives great explanation of the use-cases, details, and design considerations. Have a couple of nitpicks and one question about required arguments below but it looks good to me!

Thanks @willmj fixed the `nits`.

willmj (approved):
Thanks Dushyant, looks good to me

Abhishek-TAMU (commented):
This looks good to me. Thanks Dushyant!

PR #375, opened 2024-10-18
build(deps): torch<2.5 due to FA2 error with new version

anhuong:
### Description of the change

Reported issue with building FA2 with latest torch 2.5 that came out yesterday morning. Thus will cap torch below 2.5.

### Was the PR tested


willmj (approved):
LGTM

PR #383, opened 2024-10-31
test: get highest checkpoint instead of hard coded path

anhuong:
### Description of the change

[Unit tests were failing](https://github.com/foundation-model-stack/fms-hf-tuning/actions/runs/11579915780/job/32237578921?pr=382) on transformers v4.46 with errors:
- Resume training tests failing with `assert 9.666666666666666 == (3.6666666666666665 + 5)` where the epochs being logged didn't match
- Other tests failing because the checkpoint path that was being used was hardcoded to `checkpoint-5` but the checkpoints were only saving 3 checkpoints, not matching the number of epochs.

Because of this I updated our trainingArguments parameters to use `gradient_accumulation_steps=1` and `logging_strategy="epoch"` so that the loss logs would print per each full epoch. When setting gradient_accumulation_steps>1 the loss values would be fractions of epochs. I suspect that because the logging was set to partial epochs that the checkpoints being saved may have also been off.

This is more consistent with `save_strategy="epoch"` that we have set and I removed `logging_steps=1` since we are logging based on epochs.

The loss logging and then a print statement I added that for `os.listdir(output_dir)` showed:
```sh
{'loss': 20.4898, 'grad_norm': 23.858930587768555, 'learning_rate': 1e-05, 'epoch': 1.67}
{'loss': 20.9744, 'grad_norm': 23.15357780456543, 'learning_rate': 8.535533905932739e-06, 'epoch': 3.67}
{'train_runtime': 1.649, 'train_samples_per_second': 30.322, 'train_steps_per_second': 3.032, 'train_tokens_per_second': 1825.376, 'train_loss': 25.917576789855957, 'epoch': 3.67}
['checkpoint-1', 'training_logs.jsonl', 'checkpoint-0', 'checkpoint-2']
```
As you can see only 3 checkpoints are saved including one for checkpoint-0, likely because of the partial epochs being logged.

With these changes, the loss logging and `os.listdir(output_dir)` shows:
```sh
{'loss': 10.5682, 'grad_norm': 41.685245513916016, 'learning_rate': 9.504844339512096e-06, 'epoch': 1.0}
{'loss': 10.3005, 'grad_norm': 43.35952377319336, 'learning_rate': 7.169418695587791e-06, 'epoch': 2.0}
{'loss': 9.8799, 'grad_norm': 35.97984313964844, 'learning_rate': 3.887395330218429e-06, 'epoch': 3.0}
[...]

PR #384, opened 2024-11-01
build(deps): set transformers below 4.46, waiting on fixes

anhuong:
### Description of the change

The triton fast_kernels are broken in transformers v4.46 due to the changes and the way the patching is done in fms-acceleration described [here](https://github.com/foundation-model-stack/fms-acceleration/issues/98). In addition, the  gradient_accumulation is deteriorated in v4.46 which is fixed in [transformers PR](https://github.com/huggingface/transformers/pull/34373) which has been merged but not released yet.

In addition, upgrading transformers broke unit tests and is being resolved/discussion in #383.

Thus we will set transformers below v4.46 while we wait for these fixes to go in.

### Was the PR tested

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,7 +28,7 @@ classifiers=[
 dependencies = [
 "numpy>=1.26.4,<2.0",
 "accelerate>=0.20.3,!=0.34,<1.1",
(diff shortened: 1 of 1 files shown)


aluu317 (approved):
LGTM!

PR #387, opened 2024-11-06
build(deps): set lower limit for transformers to 4.45 for granite 3.0

willmj:
### Description of the change
Set lower limit for transformers to 4.45 to make sure fixes for granite 3.0 are included.

### Was the PR tested

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,7 +28,7 @@ classifiers=[
 dependencies = [
 "numpy>=1.26.4,<2.0",
 "accelerate>=0.20.3,!=0.34,<1.1",
-"transformers>4.41,<4.46",
+"transformers>4.45,<4.46",
 "torch>=2.2.0,<2.5",
 "sentencepiece>=0.1.99,<0.3",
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -28,7 +28,7 @@ classifiers=[
 dependencies = [
 "numpy>=1.26.4,<2.0",
 "accelerate>=0.20.3,!=0.34,<1.1",
-"transformers>4.41,<4.46",
+"transformers>4.45,<4.46",
```
Ssukriti, line 31:
```suggestion
"transformers>=4.45,<4.46",
```

Ssukriti:
@ashokponkumar just wanted to confirm with you if restricting dependencies a bit would be alright for your use cases. Mainly because granite 3.0 needs minimal transformers 4.45 . By keeping a broad range, it is confusing other users who may not be aware of this restriction to upgrade

Ssukriti (approved):

ashokponkumar (approved):
Looks good. We need this for the framework to work with granite models.

PR #393, opened 2024-11-19
fix: enable use of pretokenized dataset with padding_free

HarikrishnanBalagopal:
### Was the PR tested


kmehant:
Cherry picked the fix

PR #397, opened 2024-11-27
feat: add a new arg for HFResourceScanner callback

aluu317:
### Description of the change

This PR adds a new argument to sft_trainer `add_scanner_callback` which is False by default.
When enabled, we add a callback to use this [HFResourceScanner](https://github.com/foundation-model-stack/hf-resource-scanner) to measure memory usage, tps, and time consumed for the training. This is actual usage measurement, not an estimation.

Q: I wonder if `target_step` should be 1 here to make scanner fire on all number of train epochs?

### Was the PR tested


pyproject.toml
```diff
@@ -38,6 +38,7 @@ dependencies = [
 "protobuf>=5.28.0,<6.0.0",
 "datasets>=2.15.0,<3.0",
 "simpleeval>=0.9.13,<1.0",
+"HFResourceScanner",
```
fabianlim, line 41: @aluu317 where is this `HFResourceScanner` residing?  HF Trainer already has some internal tools to measure resource, so want to know what is the delta this provides

aluu317: @fabianlim https://github.com/foundation-model-stack/hf-resource-scanner tthat's the scanner.

Could you send me link to the HF trainer internal tool link?

kmehant: @fabianlim this is more of a light weight solution (very little overhead in terms of run time) and also reports richer information than the internal trainer tool like memory breakup. Furthermore, there is more control through this repo to incrementally add new metrics as needed.

tuning/sft_trainer.py
```diff
@@ -636,10 +648,23 @@ def main():
...
     combined_tracker_configs.aim_config = aim_config
+    sc_callback = None
 
     if training_args.output_dir:
         os.makedirs(training_args.output_dir, exist_ok=True)
         logger.info("using the output directory at %s", training_args.output_dir)
+        if add_scanner_callback:
+            # Third Party
+            # pylint: disable=import-outside-toplevel
+            from HFResourceScanner import Scanner
```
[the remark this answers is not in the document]
fabianlim: yes 100%. it must be optional. We need to do a check like [this](https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/tuning/config/acceleration_configs/acceleration_framework_config.py#L29)

pyproject.toml
fabianlim: @aluu317 you can see it via this args for [time](https://github.com/huggingface/transformers/blob/main/src/transformers/training_args.py#L1485) and [TrainerMemoryTracker](https://huggingface.co/transformers/v4.7.0/_modules/transformers/trainer_utils.html). 

So hf-resource-scanner is only for speed right? @kmehant

kmehant: @fabianlim its for time/speed and memory as well. There is more fine-grained information like what is time taken for forward pass etc.

fabianlim: @kmehant yes i noticed after i posted my original message, but I have edited it to ask a differnet thing now, please see

ChanderG: @fabianlim Scanner is more general purpose (not just time and memory, but other things as needed) and lightweight in that it instruments for a single target step (not the whole run). 

In general, it is not targeted for prod (though it can be) - mainly for testing/debugging etc.

fabianlim: @ChanderG ok then all the more it should not be the default behavior (cc. @aluu317 )

Do you mind sharing some thoughs on my comment on [torch.cuda.](https://github.com/foundation-model-stack/fms-hf-tuning/pull/397#discussion_r1861521569)

tuning/sft_trainer.py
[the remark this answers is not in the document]
aluu317: I added a commit to make the HFResourceScanner a `.[scanner-dev]` optional package. Let me know if this is what you meant @kmehant @fabianlim

pyproject.toml
ChanderG: Not sure exactly what the HFTrainer version is doing, but I have benchmarked Scanner overheads and they are negligible.

fabianlim: @ChanderG it depends on model. for the larger ones that consume alot of memory in the GPU I have observed them to be not. In particular when doing `multi-gpu` training.

aluu317:
@kmehant @fabianlim @ChanderG Should we proceed with this PR or revisit another way to scan for memory and time from tuning? Please review if it's ok.

fabianlim:
@aluu317 i leave that to others to comment. I am not very clear what are the use cases for this resource scanner. Im particular why the current HF metrics do not suffice and why this is needed.

tuning/sft_trainer.py
```diff
@@ -607,10 +624,19 @@ def main():
...
     combined_tracker_configs.file_logger_config = file_logger_config
     combined_tracker_configs.aim_config = aim_config
+    sc_callback = None
 
     if training_args.output_dir:
         os.makedirs(training_args.output_dir, exist_ok=True)
         logger.info("using the output directory at %s", training_args.output_dir)
+        if add_scanner_callback:
+            output_fmt = os.path.join(training_args.output_dir, "scanner_output.json")
+            sc_callback = [Scanner(output_fmt=output_fmt)]
```
kmehant, line 634: @aluu317 
Can we check again if the package exists here? if exists we add the callback if not we can raise a ValueError or warn the user and fall back not adding the scanner. Either ways is fine. 

Thank you.

ashokponkumar:
> @aluu317 i leave that to others to comment. I am not very clear what are the use cases for this resource scanner. Im particular why the current HF metrics do not suffice and why this is needed. But if its optional I am fine.

There are certain metrics like memory at certain points and a few others that are required for improving the estimator. It would be good to keep this optional like it is now, so that we can enable for the environments where the data collection is required. 

I would suggest lets move go ahead with this and merge, if we are okay with this optional feature.

tuning/sft_trainer.py
aluu317: sounds good. I added another check. sorry for delay, i was OOO for a few days. Please review again

kmehant (dismissed):
LGTM! Thank you.

tuning/sft_trainer.py
```diff
@@ -607,10 +624,27 @@ def main():
...
     if training_args.output_dir:
         os.makedirs(training_args.output_dir, exist_ok=True)
         logger.info("using the output directory at %s", training_args.output_dir)
+        if add_scanner_callback:
+            if _is_package_available("HFResourceScanner"):
+                output_fmt = os.path.join(
+                    training_args.output_dir, "scanner_output.json"
+                )
+                sc_callback = [Scanner(output_fmt=output_fmt)]
+                logging.info(
```
kmehant, line 638: nit: can we use `logger` instead of `logging`?

aluu317: ahh my bad, I didnt see this. Fixed

dushyantbehl:
@aluu317 @kmehant @ashokponkumar just a thought...why not use the same `tracker` backend for `Scanner`?

You can disable the `track` and `set_params` API of the tracker object by making them `NOOP` or `pass` like they are
but this would mean all the initialisation, package finding etc will move to a separate file making the main code cleaner.

tuning/sft_trainer.py
kmehant: Thank you

kmehant:
https://github.com/foundation-model-stack/fms-hf-tuning/pull/397#issuecomment-2537828610

@dushyantbehl +1 to this approach. 

Apologies @aluu317  for multiple rounds of requests :( WDYT about @dushyantbehl design? Wish to know your thoughts, thank you.

dushyantbehl:
@aluu317 can we close this as we have already merged scanner tracking into main via #422 ?

kmehant:
Featured added in https://github.com/foundation-model-stack/fms-hf-tuning/pull/422 by @aluu317

PR #402, opened 2024-12-04
fix: crash when using iterable dataset

HarikrishnanBalagopal:
### Was the PR tested


kmehant:
@HarikrishnanBalagopal Best way is to resolve features first and then fail if that does not work. see - https://github.com/foundation-model-stack/fms-hf-tuning/commit/032d1a675833db9b3d6da2caeb7f9c4c3e9ef2b5

HarikrishnanBalagopal:
> @HarikrishnanBalagopal Best way is to resolve features first and then fail if that does not work. see - [032d1a6](https://github.com/foundation-model-stack/fms-hf-tuning/commit/032d1a675833db9b3d6da2caeb7f9c4c3e9ef2b5)

`_resolve_features` is a private function, is there a public documented wrapper for it?

PR #406, opened 2024-12-06
fix: Limit trl version to <0.12

Abhishek-TAMU:
### Description of the change

Limit trl version to <0.12 due to transformers dependency change in [TRL v0.12.2](https://github.com/huggingface/trl/releases/tag/v0.12.2)

### Was the PR tested


anhuong (approved):
For more detail, since transformers is currently set to <4.46, trl>=0.12 was not being picked up due to trl 0.12.0 and 0.12.1 relying on trl version>4.46. But since trl 0.12.2 was released today the dep limit was changed to rely on transformers<4.47 which allowed us to pick up new trl version. Although Abhishek had tested on previous trl version for the release which is why we are lowering the limit. In addition, we are hitting into error when upgrading trl as seen [here](https://github.com/foundation-model-stack/fms-hf-tuning/actions/runs/12201673960/job/34040593485)

@Abhishek-TAMU let's create an issue for upgrading trl to track this

PR #408, opened 2024-12-07
feat: Perform dataset mixing via sampling probabilities in data config

dushyantbehl:
### Was the PR tested

### DIFF
--- a/tuning/data/data_processors.py
+++ b/tuning/data/data_processors.py
@@ -93,12 +93,35 @@ class DataPreProcessor:
     def _process_dataset_configs(
         self, dataset_configs: List[DataSetConfig], **extra_kwargs
     ) -> Union[Dataset, IterableDataset]:
-        train_dataset = None
-        final_datasets = None
+
         splitName = "train"  # default
 
+        all_datasetdicts = []
+        sampling_probabilities = []
+
+        # quick check to see if we are sampling and if we need to throw error.
+        sampling_probabilities = [d.sampling for d in dataset_configs if d.sampling]
+
+        if len(sampling_probabilities) > 0:
+            if len(sampling_probabilities) != len(dataset_configs):
+                raise ValueError(
+                    "Sampling probabilities should be provided for all datasets"
+                )
+            if sum(p for p in sampling_probabilities) != 1:
+                raise ValueError("Sampling probabilities don't sum to 1")
+            sample_datasets = True
+            logging.info(
+                "Sampling ratios are specified; given datasets will be interleaved."
+            )
+        else:
+            logging.info(
+                "Sampling is not specified; if multiple datasets are provided,"
+                " the given datasets will be concatenated."
+            )
+            sample_datasets = False
+
         logging.info("Starting DataPreProcessor...")
-        # Iterate over the multiple datasets provided to us
+        # Now Iterate over the multiple datasets provided to us to process
         for d in dataset_configs:
             logging.info("Loading %s", d.name)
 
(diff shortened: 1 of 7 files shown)


dushyantbehl:
@ashokponkumar @willmj @Abhishek-TAMU the data mixing feature is ready for review now.

ashokponkumar:
@dushyantbehl How do we plan to support validation split and training split?

tuning/data/data_processors.py
```diff
@@ -153,19 +174,41 @@ def _process_dataset_configs(
...
+                    datasets=v,
+                    probabilities=sampling_probabilities,
+                    stopping_strategy=strategy,
+                    seed=seed,
+                )
+                final_datasets[k] = interleaved
+        else:
+            for k, v in final_datasets.items():
+                combined = datasets.concatenate_datasets(v)
+                final_datasets[k] = combined
```
Abhishek-TAMU, line 209: Nit: Does this make sense ? (To avoid concatenation when we just pass 1 dataset without sampling, which is our basic current implementation) ?
 
`final_datasets[k] = v[0] if len(v) == 1 else datasets.concatenate_datasets(v)`

Abhishek-TAMU (commented):
Thanks @dushyantbehl, for the PR. The code to add the feature of sampling from multiple datasets looks good to me. The unit tests also look good. Just two minor suggestions.

tuning/data/data_processors.py
```diff
@@ -93,12 +93,36 @@ def load_dataset(
...
+        if len(sampling_probabilities) > 0:
+            if len(sampling_probabilities) != len(dataset_configs):
+                raise ValueError(
+                    "Sampling probabilities should be provided for all datasets"
+                )
+            if sum(p for p in sampling_probabilities) != 1:
+                raise ValueError("Sampling probabilities don't sum to 1")
+            sample_datasets = True
+            logging.info(
+                "Sampling ratios are specified given datasets will be interleaved."
```
Abhishek-TAMU, line 116: Nit: `"Sampling ratios are specified; given datasets will be interleaved."`

tuning/data/data_processors.py
```diff
@@ -93,12 +93,36 @@ def load_dataset(
...
+            )
+        else:
+            logging.info(
+                "Sampling is not specified given datasets will be concatenated"
```
Abhishek-TAMU, line 120: If we choose to accept [this](https://github.com/foundation-model-stack/fms-hf-tuning/pull/408/files#r1876530327), then can make it as:

`Sampling is not specified; if multiple datasets are provided, the given datasets will be concatenated.`

willmj:
Made an image for testing: https://v3.travis.ibm.com/github/ai-foundation/sft-trainer-image/builds/31001438

tuning/data/data_processors.py
```diff
@@ -93,12 +93,36 @@ def load_dataset(
...
+
         splitName = "train"  # default
 
+        all_datasetdicts = []
+        sampling_probabilities = []
+
+        # quick check to see if we are sampling and if we need to throw error.
+        for d in dataset_configs:
+            if d.sampling:
+                sampling_probabilities.append(d.sampling)
```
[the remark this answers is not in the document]
dushyantbehl: Good suggestion. Applied

dushyantbehl:
> @dushyantbehl How do we plan to support validation split and training split?

@ashokponkumar as discussed offline this feature will be left to subsequent patches we apply so current patch is independent of this feature thanks.

Abhishek-TAMU (approved):
Changes look good to me!

PR #399, opened 2024-12-02
fix: Moved deprecated positional arguments from SFTTrainer to SFTConfig

Luka-D:
### Description of the change

This PR moves deprecated positional arguments from SFTTrainer to the SFTConfig declaration. This is due to some positional arguments, such as `dataset_kwargs`, being depricated with version '1.0.0'. 

`dataset_kwargs` was moved to `training_args` following the example by HF [here](https://huggingface.co/docs/trl/en/sft_trainer#training-the-vision-language-model). Additionally, `max_seq_length`, `dataset_text_field` and `packing` were also moved into `training_args`.  Following the class definition [here](https://huggingface.co/docs/trl/en/sft_trainer#trl.SFTConfig)

As a result, these warnings were fixed:
```
FutureWarning: Deprecated argument(s) used in '__init__': dataset_kwargs. Will not be supported from version '1.0.0'.

Deprecated positional argument(s) used in SFTTrainer, please use the SFTConfig to set these arguments instead.
  warnings.warn(message, FutureWarning)
```
### Related issue number

I began working off of Will's branch with [PR #317](https://github.com/foundation-model-stack/fms-hf-tuning/pull/317), however since some of his changes haven't been approved yet, I have only included changes to positional arguments.

### How to verify the PR

Run:
``` 
python3 tuning/sft_trainer.py \
--model_name_or_path Maykeye/TinyLLama-v0 \
--training_data_path tests/data/twitter_complaints_small.jsonl \
--output_dir outputs/lora-tuning \
--num_train_epochs 5 \
--per_device_train_batch_size 4 \
--gradient_accumulation_steps 4 \
--learning_rate 1e-5 \
--response_template "\n### Label:" \
--dataset_text_field "output" \
--use_flash_attn false \
--torch_dtype "float32" \
--peft_method "lora" \
--r 8 \
--lora_dropout 0.05 \
--lora_alpha 16  \
--log_level "error"
```

The logs used to show these warnings:
```
/Users/luka/Documents/GitHub/fms-hf-tuning/.venv/lib/python3.12/site-packages/huggingface_hub/utils/_deprecation.py:100: FutureWarning: Deprecated argument(s) used in '__init__': dataset_kwargs. Will not be supported from version '1.0.0'.

Deprecated positional argument(s) used in SFTTrainer, please use the SFTConfig to set these arguments instead.
  warnings.warn(message, FutureWarning)
[...]

PR #418, opened 2024-12-13
fix: update dataclass objects directly instead of creating new variables

kmehant:
There was spread of information in multiple variables, therefore confining the source of truth to a single variable, preferably with dataclass objects in this scenario.

cc: @dushyantbehl

### DIFF
--- a/tuning/sft_trainer.py
+++ b/tuning/sft_trainer.py
@@ -295,9 +295,9 @@ def train(
     (
         formatted_train_dataset,
         formatted_validation_dataset,
-        dataset_text_field,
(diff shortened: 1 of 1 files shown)


dushyantbehl:
LGTM.

Abhishek-TAMU (commented):
This looks good to me. Additionally PR https://github.com/foundation-model-stack/fms-hf-tuning/pull/399 also ignored using `max_seq_len` from function `process_dataargs`. Thanks @kmehant @dushyantbehl for the fix and cleanup.

ashokponkumar (approved):

PR #419, opened 2024-12-13
fix: Remove deprecated push_to_hub_token to resolve warning

Luka-D:
### Description of the change

Running tuning locally produces this warning message:
```
FutureWarning: `--push_to_hub_token` is deprecated and will be removed in version 5 of 🤗 Transformers. Use `--hub_token` instead.
```
When I print out the `transformer_kwargs` dictionary, I can see that both `hub_token` and `push_to_hub_token` are in the dictionary. Here is a snippet of it.
```
.....'hub_token': '<HUB_TOKEN>', 'hub_private_repo': False, 'hub_always_push': False, 'gradient_checkpointing': False, 'gradient_checkpointing_kwargs': None, 'include_inputs_for_metrics': False, 'eval_do_concat_batches': True, 'fp16_backend': 'auto', 'evaluation_strategy': None, 'push_to_hub_model_id': None, 'push_to_hub_organization': None, 'push_to_hub_token': '<PUSH_TO_HUB_TOKEN>',......
```

I made a temporary fix by simply popping `push_to_hub_token` out of the dictionary and now the warning is gone.
```
# Remove deprecated push_to_hub_token
transformer_kwargs.pop("push_to_hub_token", None)
```

I am wondering if anyone knows where `push_to_hub_token` is originating from or best way to remove it.

### Related issue number

Issue [#1205](https://github.ibm.com/ai-foundation/watson-fm-stack-tracker/issues/1205)

### How to verify the PR

Run tuning locally:
```
python3 tuning/sft_trainer.py \
--model_name_or_path Maykeye/TinyLLama-v0 \
--training_data_path tests/artifacts/testdata/jsonl/twitter_complaints_small.jsonl \
--output_dir outputs/lora-tuning \
--num_train_epochs 5 \
--per_device_train_batch_size 4 \
--gradient_accumulation_steps 4 \
--learning_rate 1e-5 \
--response_template "\n### Label:" \
--use_flash_attn false \
--torch_dtype "float32" \
--peft_method "lora" \
--r 8 \
--lora_dropout 0.05 \
--lora_alpha 16  \
--log_level "error"
```


Luka-D:
Summary of changes made with https://github.com/foundation-model-stack/fms-hf-tuning/pull/419/commits/9e567e2576c24f380207521fbd2b04befc2ba3cb

It seems that using .to_dict() method on `train_args` was returning the `hub_token` and `push_to_hub_token` fields, which had a default value like `<PUSH_TO_HUB_TOKEN>`.  Changing it to `train_args.__dict__.items()` has removed some of those default values.

Because of the change to `train_args.__dict__.items()`, I needed to change SFTConfig fields function to only return parameters that are accepted by SFTConfig. Previously it was also returning fields like `_n_gpu`, which is inherited from `TrainingArguments` class but is not accepted in SFTConfig and will raise an error. Using `SFTConfig.__dict__["__match_args__"]` will return the same parameters as those listed in the variables window on HF: https://huggingface.co/docs/trl/en/sft_trainer#trl.SFTConfig.

In summary, the changes to `transformer_kwargs` are:
- `hub_token` and `push_to_hub_token` are now both None, instead of `<hub_token>` and `<push_to_hub_token>`
- Some of the parameters have changed from strings into Object types. eg. `'lr_scheduler_type': 'linear'` has become `'lr_scheduler_type': <SchedulerType.LINEAR: 'linear'>`. `'save_strategy': 'epoch'` has become `'save_strategy': <IntervalStrategy.EPOCH: 'epoch'>` etc.
- I do not know if this change to Object types is the desired result, would like some feedback on this.

All code has passed `tox -e py` tests.

Please let me know your thoughts, thank you.

tuning/sft_trainer.py
```diff
@@ -346,10 +345,10 @@ def train(
     # from our object directly. In the future, we should consider renaming this class and / or
     # not adding things that are not directly used by the trainer instance to it.
 
-    transformer_train_arg_fields = [x.name for x in dataclasses.fields(SFTConfig)]
+    transformer_train_arg_fields = SFTConfig.__dict__["__match_args__"]
```
kmehant, line 348: Use of `__match_args__` requires Python>=3.10 isn't it? Correct me.

Luka-D: Yes, I believe it is Python 3.10 and greater. Although I'm a little confused on whether `__match_args__` are auto generated by Python when creating a class or manually set by those writing the class. As it is now, it is a key in the dictionary, so I wonder if the key would still exist in Python 3.9 and lower.

kmehant: I guess we should avoid usage of `__match_args__ ` in that case, WDYT?

Luka-D:
I made small change with https://github.com/foundation-model-stack/fms-hf-tuning/pull/419/commits/c7d540e43eb35314934a4e1b6fd9834954c5f6b7. I changed `__dict__` to `vars()`. The functionality is the same for both, `vars()` accesses the `__dict__` method, however using `vars()` is more readable and more pythonic in my opinion.

kmehant:
https://github.com/foundation-model-stack/fms-hf-tuning/pull/419#issuecomment-2612921428

Thank you @Luka-D , I will let @anhuong do a review.

tuning/sft_trainer.py
anhuong: We do support Python 3.9-3.12 currently

aluu317:
I've been following this thread and had some convos with Luka in Slack. Here's a summary to the best of my ability:

**Context**: There's an issue in HF with the `to_dict()` [function](https://github.com/huggingface/transformers/blob/main/src/transformers/training_args.py#L2568) of `TrainingArguments`. Or maybe it's by design:
This block in HF code:
```
def to_dict(self):
    // some other code
    for k, v in d.items():
           // some other stuff
            if k.endswith("_token"):
                d[k] = f"<{k.upper()}>"
```
This masks the values of any `_token` field in TrainingArguments. So even when we pass no tokens in training args, we call `train_args.to_dict()` and now we have a value for `push_to_hub_token`, and the value is the literal string "<PUSH_TO_HUB_TOKEN>". I'm thinking it could be by design that the token field is masked so it's not detectable whether the token was passed in or not. But this also changes the behavior of `to_dict()` to give the push_to_hub_token a bogus value even though it's supposed to be None.

Now, back to **our scenario**, we can avoid this by _not_ calling the culprit function `to_dict()`. That's why Luka opted in to use `.__dict__` instead. The problem with using `.__dict__` is that it's returning all fields on `TrainingArguments` including a field called `_n_gpu` that is a _non-init_ field for this class. So when we pass the arguments back into `SFTConfig` (which is a `TrainingArguments`), we get error that we are passing in `_n_gpu` which is a non-init field. 

[Ref](https://github.com/huggingface/transformers/blob/main/src/transformers/training_args.py#L1443):
```
_n_gpu: int = field(init=False, repr=False, default=-1)
```

So I think the solution here is to:
- use `.__dict__` or `vars()`, either of these will return same fields on `TrainingArguments` so we avoid getting some bogus value for `push_to_hub_token` even if it should be None.
- to remove the `_n_gpu` field from `transformer_train_arg_fields` which is what Luka is doing in the PR, the right thing to do here is to filter it with `if x.init`. So something like:
```
transformer_train_arg_fields = [x.name for x in dataclasses.fields(SFTConfig) if x.init]
``` 
and put in a note that we are filtering out fields that aren't part of the init fields, similar to this [note](https://github.com/huggingface/transformers/blob/main/src/transformers/training_args.py#L2560) in HF.

aluu317 (approved):
lgtm!

PR #421, opened 2024-12-16
feat: exit on unknown args

### LINKED ISSUE #420: Log and/or exit on unknown args

Sometimes we have typos in the input command arg, or use incorrect names of flags.

fms-hf-tuning to cleanly report unknown args and fail, instead of silently ignoring it and proceeding.

None.

## Additional context

None.

ChanderG:
### Description of the change

Logs and errors out in case of unknown input args.

### Related issue number

Closes #420 

### How to verify the PR

Simply run with some extra/flags with typos etc.

### Was the PR tested

Manually.

### DIFF
--- a/tuning/sft_trainer.py
+++ b/tuning/sft_trainer.py
@@ -516,11 +516,14 @@ def parse_arguments(parser, json_config=None):
             fusedops_kernels_config,
             attention_and_distributed_packing_config,
             additional,
-            _,
+            leftover,
         ) = parser.parse_args_into_dataclasses(return_remaining_strings=True)
 
         peft_method = additional.peft_method
         exp_metadata = additional.exp_metadata
+        if leftover:
+            logging.error("Extra un-recognized arguments found: %s", leftover)
+            sys.exit(USER_ERROR_EXIT_CODE)
 
     if peft_method == "lora":
         tune_config = lora_config


tuning/sft_trainer.py
```diff
@@ -516,11 +516,14 @@ def parse_arguments(parser, json_config=None):
...
             additional,
-            _,
+            leftover,
         ) = parser.parse_args_into_dataclasses(return_remaining_strings=True)
 
         peft_method = additional.peft_method
         exp_metadata = additional.exp_metadata
+        if leftover:
+            logging.error("Extra un-recognized arguments found: %s", leftover)
+            sys.exit(USER_ERROR_EXIT_CODE)
```
ashokponkumar, line 526: Wouldn't be better to log and continue?

anhuong: Sorry for the delay in reviewing this, there are two issues here -- to pass in accelerate launch arguments via JSON for kube deployments, the user must pass in a new field called `accelerate_launch_args` as described [here](https://github.com/foundation-model-stack/fms-hf-tuning/tree/main/build#configuration). This is interpreted in the accelerate_launch.py script and then ignored in SFTTrainer. This change would break this behavior. I can see how being warned about misspellings or additional arguments is helpful. We can start this off as warning to not break current behavior or this PR can address the issue with `accelerate_launch_args` by removing them in the accelerate_launch.py script

anhuong (changes requested):

PR #425, opened 2024-12-19
feat: Integrate MLflow tracker

dushyantbehl:
### Description of the change

This code change integrates MLflow tracker backed via the existing `tracker` api in `fms-hf-tuning`

### Related issue number

Closes https://github.ibm.com/ai-foundation/watson-fm-stack-tracker/issues/1495

### How to verify the PR

TBA

### Was the PR tested

### DIFF
--- a/README.md
+++ b/README.md
@@ -823,12 +823,13 @@ For details about how you can use set a custom stopping criteria and perform cus
 
 ## Experiment Tracking
 
-Experiment tracking in fms-hf-tuning allows users to track their experiments with known trackers like [Aimstack](https://aimstack.io/) or custom trackers built into the code like
+Experiment tracking in fms-hf-tuning allows users to track their experiments with known trackers like [Aimstack](https://aimstack.io/), [MLflow Tracking](https://mlflow.org/docs/latest/tracking.html) or custom trackers built into the code like
 [FileLoggingTracker](./tuning/trackers/filelogging_tracker.py)
 
 The code supports currently two trackers out of the box, 
 * `FileLoggingTracker` : A built in tracker which supports logging training loss to a file.
 * `Aimstack` : A popular opensource tracker which can be used to track any metrics or metadata from the experiments.
+* `MLflow Tracking` : Another popular opensource tracker which stores metrics, metadata or even artifacts from experiments.
 
 Further details on enabling and using the trackers mentioned above can be found [here](docs/experiment-tracking.md).  
 
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -19,8 +19,9 @@ ARG USER=tuning
 ARG USER_UID=1000
 ARG PYTHON_VERSION=3.11
 ARG WHEEL_VERSION=""
-## Enable Aimstack if requested via ENABLE_AIM set to "true"
+## Enable Aimstack or MLflow if requested via ENABLE_AIM/MLFLOW set to "true"
 ARG ENABLE_AIM=false
+ARG ENABLE_MLFLOW=false
 ARG ENABLE_FMS_ACCELERATION=true
 
 ## Base Layer ##################################################################
@@ -150,6 +151,10 @@ RUN if [[ "${ENABLE_FMS_ACCELERATION}" == "true" ]]; then \
 RUN if [[ "${ENABLE_AIM}" == "true" ]]; then \
         python -m pip install --user "$(head bdist_name)[aim]"; \
     fi
+
+RUN if [[ "${ENABLE_MLFLOW}" == "true" ]]; then \
+    python -m pip install --user "$(head bdist_name)[mlflow]"; \
+fi
 
     # Clean up the wheel module. It's only needed by flash-attn install
 RUN python -m pip uninstall wheel build -y && \
--- a/docs/experiment-tracking.md
+++ b/docs/experiment-tracking.md
@@ -115,6 +115,34 @@ sft_trainer.train(train_args=training_args, tracker_configs=tracker_configs,....
 The code expects either the `local` or `remote` repo to be specified and will result in a `ValueError` otherwise.
 See [AimConfig](https://github.com/foundation-model-stack/fms-hf-tuning/blob/a9b8ec8d1d50211873e63fa4641054f704be8712/tuning/config/tracker_configs.py#L25) for more details.
 
+## MLflow Tracker
+
+To enable [MLflow Tracking](https://mlflow.org/docs/latest/tracking.html) users need to pass `"mlflow"` as the requested tracker as part of the [training argument](https://github.com/foundation-model-stack/fms-hf-tuning/blob/a9b8ec8d1d50211873e63fa4641054f704be8712/tuning/config/configs.py#L131).
+
+
+When using MLflow, users need to specify additional arguments which specify [mlflow tracking uri](https://mlflow.org/docs/latest/tracking.html#common-setups) location where either a [mlflow supported database](https://mlflow.org/docs/latest/tracking/backend-stores.html#supported-store-types) or [mlflow remote tracking server](https://mlflow.org/docs/latest/tracking/server.html) is running.
+
+Example
+```
+from tuning import sft_trainer
+from tuning.config.tracker_configs import MLflowConfig, TrackerConfigFactory
+
+training_args = TrainingArguments(
+    ...,
+    trackers = ["mlflow"],
+)
+
+tracker_configs = TrackerConfigFactory(
+    mlflow_config=MLflowConfig(
+        mlflow_experiment="experiment-name",
+        mlflow_tracking_uri=<tracking uri>
+        )
+    )
+
+sft_trainer.train(train_args=training_args, tracker_configs=tracker_configs,....)
+```
+
+The code expects a valid uri to be specified and will result in a `ValueError` otherwise.
 
 ## Running the code via command line `tuning/sft_trainer::main` function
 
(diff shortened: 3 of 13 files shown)


README.md
```diff
@@ -823,12 +823,13 @@ For details about how you can use set a custom stopping criteria and perform cus
 
 ## Experiment Tracking
 
-Experiment tracking in fms-hf-tuning allows users to track their experiments with known trackers like [Aimstack](https://aimstack.io/) or custom trackers built into the code like
+Experiment tracking in fms-hf-tuning allows users to track their experiments with known trackers like [Aimstack](https://aimstack.io/), [MLflow](https://mlflow.org/) or custom trackers built into the code like
```
praveenj83, line 826:
```suggestion
Experiment tracking in fms-hf-tuning allows users to track their experiments with known trackers like [Aimstack](https://aimstack.io/), [MLflow Tracking](https://mlflow.org/docs/latest/tracking.html) or custom trackers built into the code like
```

README.md
```diff
@@ -823,12 +823,13 @@ For details about how you can use set a custom stopping criteria and perform cus
...
 [FileLoggingTracker](./tuning/trackers/filelogging_tracker.py)
 
 The code supports currently two trackers out of the box, 
 * `FileLoggingTracker` : A built in tracker which supports logging training loss to a file.
 * `Aimstack` : A popular opensource tracker which can be used to track any metrics or metadata from the experiments.
+* `MLflow` : Another popular opensource tracker which stores metrics, metadata or even artifacts from experiments.
```
praveenj83, line 832:
```suggestion
* `MLflow Tracking` : Another popular opensource tracker which stores metrics, metadata or even artifacts from experiments.
```

docs/experiment-tracking.md
```diff
@@ -115,6 +115,34 @@ sft_trainer.train(train_args=training_args, tracker_configs=tracker_configs,....
 The code expects either the `local` or `remote` repo to be specified and will result in a `ValueError` otherwise.
 See [AimConfig](https://github.com/foundation-model-stack/fms-hf-tuning/blob/a9b8ec8d1d50211873e63fa4641054f704be8712/tuning/config/tracker_configs.py#L25) for more details.
 
+## MLflow Tracker
+
+To enable [MLflow](https://mlflow.org/) users need to pass `"mlflow"` as the requested tracker as part of the [training argument](https://github.com/foundation-model-stack/fms-hf-tuning/blob/a9b8ec8d1d50211873e63fa4641054f704be8712/tuning/config/configs.py#L131).
```
praveenj83, line 120:
```suggestion
To enable [MLflow Tracking](https://mlflow.org/docs/latest/tracking.html) users need to pass `"mlflow"` as the requested tracker as part of the [training argument](https://github.com/foundation-model-stack/fms-hf-tuning/blob/a9b8ec8d1d50211873e63fa4641054f704be8712/tuning/config/configs.py#L131).
```

tuning/trackers/tracker_factory.py
```diff
@@ -26,21 +26,29 @@
...
+_is_mlflow_available = _is_package_available("mlflow")
 
 
 def _get_tracker_class(T, C):
     return {"tracker": T, "config": C}
 
 
+def _is_tracker_installed(name):
+    if name == "aim":
+        return _is_aim_available
```
ashokponkumar, line 48: Dont' we need to add support for mlflow too?

praveenj83:
@dushyantbehl Have you made a note that this requires updating the same image as we use for AIM for the time being?

dushyantbehl:
> @dushyantbehl Have you made a note that this requires updating the same image as we use for AIM for the time being?

Yes @praveenj83 I have added arguments in dockerfile to enable MLflow like we do for Aim. We can apply it to the same image.

tuning/trackers/tracker_factory.py
```diff
@@ -60,10 +68,23 @@ def _register_aim_tracker():
...
+        # Local
+        from .mlflow_tracker import MLflowTracker
+        from tuning.config.tracker_configs import MLflowConfig
+
+        mlflow_tracker = _get_tracker_class(MLflowTracker, MLflowConfig)
+
+        REGISTERED_TRACKERS[MLFLOW_TRACKER] = mlflow_tracker
+        logging.info("Registered mlflow tracker")
+    else:
+        logging.info(
```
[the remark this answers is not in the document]
dushyantbehl: Yes..I will change it to error.

tuning/trackers/mlflow_tracker.py
```diff
@@ -0,0 +1,204 @@
...
+from transformers.integrations import MLflowCallback  # pylint: disable=import-error
+
+# Local
+from .tracker import Tracker
+from tuning.config.tracker_configs import MLflowConfig
+
+MLFLOW_RUN_URI_EXPORT_FILENAME = "mlflow_tracker.json"
+
+
+class RunIDExporterMlFlowCallback(MLflowCallback):
```
ashokponkumar, line 30:
```suggestion
class RunURIExporterMlFlowCallback(MLflowCallback):
```

tuning/trackers/mlflow_tracker.py
```diff
@@ -0,0 +1,204 @@
...
+    """
+    Custom MlflowCallBack callback is used to export run id from Mlflow
```
ashokponkumar, line 32:
```suggestion
    Custom MlflowCallBack callback is used to export run uri from Mlflow
```

tuning/trackers/mlflow_tracker.py
```diff
@@ -0,0 +1,204 @@
...
+
+        active_run = self._ml_flow.active_run()
+        if not active_run:
+            return
+
+        active_run_info = active_run.info
+        if active_run_info:
+            experiment_id = active_run_info.experiment_id
+            experiment_url = f"<host_url>/#/experiments/{experiment_id}"
+            run_id = active_run_info.run_id
```
ashokponkumar, line 73: Is it called run_id in MLFlow?

dushyantbehl: Yes - https://github.com/mlflow/mlflow/blob/master/mlflow/entities/run_info.py#L44

tuning/trackers/tracker_factory.py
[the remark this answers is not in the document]
dushyantbehl: This shouldn't be an error because it happens at code load time when we are registering all available trackers.
The actual error is thrown later if user requests this tracker. In this case we just inform user that if she intends to use the tracker it should be installed.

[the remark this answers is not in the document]
ashokponkumar: then shouldtn't this be debug message?

ashokponkumar:
LGTM

ashokponkumar (dismissed):

tuning/config/tracker_configs.py
```diff
@@ -63,7 +63,27 @@ def __post_init__(self):
...
+    # If mlflow_run_uri_export_path is set the run uri will be output in a json format
+    # to the location pointed to by `mlflow_run_uri_export_path/mlflow_tracker.json`
+    # If this is not set then the default location where run uri will be exported
+    # is training_args.output_dir/mlflow_tracker.json
+    # Run uri is not exported if mlflow_run_uri_export_path variable is not set
+    # and output_dir is not specified.
+    mlflow_run_uri_export_path: str = None
+
+    def __post_init__(self):
+        if self.mlflow_experiment is None:
```
aluu317, line 81: why not just set the default value above (similar to `FileLoggingTrackerConfig`)?
```
mlflow_experiment: str = "fms-hf-tuning"
```

tuning/trackers/tracker_factory.py
```diff
@@ -81,6 +104,8 @@ def _register_trackers():
         _register_aim_tracker()
     if FILE_LOGGING_TRACKER not in REGISTERED_TRACKERS:
         _register_file_logging_tracker()
+    if MLFLOW_TRACKER not in REGISTERED_TRACKERS:
```
aluu317, line 107: For completeness, could you also add a comment below line 100?

build/Dockerfile
```diff
@@ -151,6 +152,10 @@ RUN if [[ "${ENABLE_AIM}" == "true" ]]; then \
         python -m pip install --user "$(head bdist_name)[aim]"; \
     fi
 
+RUN if [[ "${ENABLE_MLFLOW}" == "true" ]]; then \
```
aluu317, line 155: Do we want to only install this package when both `ENABLE_AIM` and `ENABLE_MLFLOW` are set to true? What happens if `ENABLE_AIM` is false and `ENABLE_MLFLOW` is true, we would install here, which is not what we want right?

aluu317: I think I understand now, they are independent. We don't need AIM to enable MLFLOW

tests/trackers/test_mlflow_tracker.py
```diff
@@ -0,0 +1,130 @@
...
+
+    # This should not mean file logger is not present.
+    # code will add it by default
+    # The below validate_training check will test for that too.
+    train_args.trackers = ["mlflow"]
+
+    tracker_configs = TrackerConfigFactory(
+        mlflow_config=MLflowConfig(
+            mlflow_experiment="unit_test",
+            mlflow_tracking_uri=os.path.join(tempdir, "mlflowdb.sqlite"),
```
aluu317, line 84: Nice. Should we add a another assert in the end of this test to verify the mlflow tracking worked? Something like:
```
assert os.path.exists(os.path.join(tempdir, "mlflowdb.sqlite") is True
```
and maybe that the content has the experiment name `unit_test`?

aluu317 (commented):
Overall looks good! I just had some nits

tuning/trackers/tracker_factory.py
dushyantbehl: sure!

build/Dockerfile
dushyantbehl: yes we do not need AIM..they are independent packages.

tests/trackers/test_mlflow_tracker.py
dushyantbehl: I have added this check but the experiment name is under a custom data format by mlflow so hard to find so for now `mlflow` db is present should suffice for us as the run won't succeed without it.

aluu317 (approved):
LGTM!

PR #431, opened 2025-01-02
feat: Allow hf dataset id to be passed via training_data_path

dushyantbehl:
### Description of the change

Recent commit https://github.com/foundation-model-stack/fms-hf-tuning/commit/d7f06f5fc898eb700a9e89f08793b2735d97889c introduced support for hf dataset id inside `data_config` but the code had a minor fix needed to support hf dataset id being passed via `training_data_path`.

With this patch code now supports hf dataset id via `training_data_path` argument.

### How to verify the PR

Tested with an hf dataset and added a unit test for the same.

### Was the PR tested

### DIFF
--- a/tuning/data/data_handlers.py
+++ b/tuning/data/data_handlers.py
@@ -130,7 +130,7 @@ def apply_custom_data_formatting_template(
         if index_object not in element:
             raise KeyError("Requested template string is not a valid key in dict")
 
-        return element[index_object]
+        return str(element[index_object])
 
     return {
         dataset_text_field: re.sub(r"{{([\s0-9a-zA-Z_\-\.]+)}}", replace_text, template)
--- a/README.md
+++ b/README.md
@@ -62,13 +62,13 @@ pip install fms-hf-tuning[aim]
 For more details on how to enable and use the trackers, Please see, [the experiment tracking section below](#experiment-tracking).
 
 ## Data Support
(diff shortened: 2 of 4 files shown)


dushyantbehl:
cc @anhuong @Abhishek-TAMU

tuning/data/data_handlers.py
```diff
@@ -130,7 +130,7 @@ def replace_text(match_obj):
         if index_object not in element:
             raise KeyError("Requested template string is not a valid key in dict")
 
-        return element[index_object]
+        return str(element[index_object])
```
dushyantbehl, line 133: The dataset which I have tested in my unit test is `lhoestq/demo1` which is used with the formatting template `"### Text:{{review}} \n\n### Stars: {{star}}"` 
Now if you see the dataset the field `star` is an `int` which would cause an error if this element was not a string.

so this change supports any dataset formatting with dataset row types other than string as well.

dushyantbehl:
changes look good to me @Abhishek-TAMU. Can you please also check why the image build failed post the change.

willmj (approved):
Looks good to me as well

willmj:
@dushyantbehl a new release of [fms-acceleration](https://github.com/foundation-model-stack/fms-acceleration/releases/tag/v0.5.0) just came out which may be causing this problem.

anhuong:
I also verified that this works with a different dataset to load from HF, thanks Dushyant!

Abhishek-TAMU (approved):
LGTM after the image build passes successfully!

PR #434, opened 2025-01-08
fix: function name from `requires_agumentation` to `requires_augmentation`

willmj:
Nit fix for function in fms-acceleration which needs to be changed in fms-hf-tuning as well.
[PR #118](https://github.com/foundation-model-stack/fms-acceleration/pull/118) in fms-acceleration should be merged at the same time.


fabianlim (dismissed):
LGTM

fabianlim:
@willmj there is a problem with the image [build](https://github.com/foundation-model-stack/fms-hf-tuning/actions/runs/12673008681/job/35318349236?pr=434) action. The command `docker rmi $(docker image ls -aq)` fails because there are no existing images. So need to guard against that

fabianlim (approved):

PR #436, opened 2025-01-09
fix: Add pad_token to special_tokens_dict when pad_token == eos_token

Abhishek-TAMU:
### Description of the change

Add `pad_token` to special_tokens_dict when `pad_token == eos_token`

### Related issue number

Issue:https://github.com/foundation-model-stack/fms-hf-tuning/pull/343
Issue 343 modifies pad_token of tokenizer but in this PR adds it to `special_tokens_dict` for model resizing. 

Reason of this case of `pad_token == eos_token`: 
When tokenizer has PAD token equal to EOS token, then the DataCollatorForCompletionOnlyLM masks the EOS token from labels of dataset batch and hence its not used in loss calculation and hence it leads to repetitive `predicted_target` during inference time.

### Was the PR tested

### DIFF
--- a/tuning/sft_trainer.py
+++ b/tuning/sft_trainer.py
@@ -290,8 +290,10 @@ def train(
             )
             if tokenizer.eos_token != configs.DEFAULT_PAD_TOKEN:
                 tokenizer.pad_token = configs.DEFAULT_PAD_TOKEN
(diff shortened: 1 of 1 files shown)


Ssukriti (commented):
thank you! 

for context for others: Without this fix the token values are not actually getting modified, to set a unique pad token and then EOS is masked

Ssukriti (approved):

PR #437, opened 2025-01-14
feat: Enable streaming in data preprocessor

willmj:
### Description of the change
These changes enable streaming and test streaming datasets.
Added:

- Add `streaming` as an arg in `DataSetConfig` similarly to `sampling`
- Add examples of DataSetConfig in tests/artifacts/predefined_data_configs/ for streaming
- Add unit tests
- Since `IterableDatasets` can't be indexed, use first example where column names are needed
- User must set `max_steps` instead of `num_train_epochs` if using streaming

### How to verify the PR

- [x] Run new unit tests which verify HF inference works and passing `streaming` in dataconfig returns and IterableDataset
- [x] Run on single GPU error
- [x] Run on multi GPU without error

### Was the PR tested


tuning/sft_trainer.py
```diff
@@ -347,6 +348,11 @@ def train(
     }
     training_args = SFTConfig(**transformer_kwargs, **additional_args)
 
+    if isinstance(formatted_train_dataset, IterableDataset):
+        # Which one? Should it be for user to decide or set?
+        # training_args.max_steps = training_args.num_train_epochs
+        training_args.max_steps = 1
```
willmj, line 354: Could do
```
        if training_args.num_train_epochs:
            training_args.max_steps = training_args.num_train_epochs
        else:
            training_args.max_steps = 1
```

willmj: Or if `max_steps = training_size/batch_size` could calculate it in the else statement like that

Abhishek-TAMU: Guess for `Iterable dataset`, we don't have fixed length of complete dataset so it seems number of epochs doesn't work with it and hence needs `max_steps`. 
So yea we could probably calculate `max_steps` using, `training_size, batch_size_per_device, GA, num_train_epochs`.

Would like to see what @dushyantbehl thinks...

Abhishek-TAMU (commented):
Thanks @willmj for integrating usage of Iterable datasets. Just some initial thoughts.

ashokponkumar:
Shouldn't streaming be a top level object instead of a per dataset object? Is it possible to mix streaming and non-streaming datasets using concat?

tuning/data/data_processors.py
```diff
@@ -271,8 +298,9 @@ def _process_dataset_configs(
                     if kwargs["remove_columns"] == "all":
                         kwargs["remove_columns"] = column_names
 
-                    if "num_proc" not in kwargs:
-                        kwargs["num_proc"] = os.cpu_count()
+                    if not d.streaming:
+                        if "num_proc" not in kwargs:
+                            kwargs["num_proc"] = os.cpu_count()
```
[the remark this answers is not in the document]
willmj: The `map` function for `IterableDatasetDict` does not take `num_proc` as an argument, so I'd assume yes.

seshapad:
@willmj Is this PR in a usable state? We need to run a EPT with large datasets. Without streaming the data processing is failing. We want the streaming feature to address this issue.

kmehant:
@willmj I would request your attention to this https://github.com/foundation-model-stack/fms-hf-tuning/blob/224f35b21777eeff8038e6a62b8ef7126adcdcd6/tuning/utils/preprocessing_utils.py#L49-L60. iterabledatasets often loose out column information, so its good to be defensive on retrieving columns wherever necessary.

tests/test_sft_trainer.py
```diff
@@ -791,6 +792,59 @@ def test_run_causallm_ft_pretokenized(dataset_path):
         assert "### Text: @NortonSupport Thanks much.\n\n### Label:" in output_inference
 
 
+@pytest.mark.parametrize(
+    "datafiles, datasetconfigname",
+    [
+        (
+            [TWITTER_COMPLAINTS_DATA_DIR_JSON],
+            DATA_CONFIG_TOKENIZE_AND_APPLY_INPUT_MASKING_YAML,
```
Abhishek-TAMU, line 800: I assume this yaml files is to be used for the this test case: `DATA_CONFIG_YAML_STREAMING`

willmj:
@seshapad I have now had a successful tuning job with streaming on multi GPU. You should be able to try it out, let me know if you run into any errors.

tuning/data/setup_dataprocessor.py
```diff
@@ -45,20 +45,43 @@
...
+
+    if isinstance(data, IterableDataset):
+        # For an IterableDataset, inspect the first element if possible
+        try:
+            first_example = next(iter(data))
+            return (
+                isinstance(first_example, dict)
+                and "input_ids" in first_example
+                and "labels" in first_example
+            )
```
willmj, line 73: @kmehant is this check sufficient to what you posted?

tuning/data/setup_dataprocessor.py
```diff
@@ -68,6 +91,17 @@ def _process_dataconfig_file(
...
+        if train_args.max_steps < 1:
+            logging.error(
+                "ValueError: `--max_steps` must be set when streaming is set in data \
+                            preprocessor config"
+            )
+            raise ValueError(
+                "`--max_steps` must be set when streaming is set in data preprocessor config"
+            )
+        if train_args.num_train_epochs:
+            logging.warning("`--num_train_epochs` will be overwritten by `--max_steps`")
```
kmehant, line 104: Since, this is the default behaviour of the HF transformers - https://github.com/huggingface/transformers/blob/ec7afad60909dd97d998c1f14681812d69a15728/src/transformers/trainer.py#L695 

We can avoid this warning all together.

tuning/data/setup_dataprocessor.py
kmehant: Looks good to me.

kmehant: @willmj 
Can we have a unit test case with pretokenized + packing + streaming (common EPT case to cover)? Thank you.

tuning/sft_trainer.py
```diff
@@ -332,6 +333,9 @@ def train(
         time.time() - data_preprocessing_time
     )
 
+    if isinstance(formatted_train_dataset, IterableDataset):
+        train_args.split_batches = True
```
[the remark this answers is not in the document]
kmehant: @willmj an important point to be captured in the log is that, since `split_batches` is being set to True the message should also capture that `per_device_train_batch_size` would now be `global batch size` and no more holds that meaning.

willmj:
Tuning + inference works well!
Config:
```
      {
          "model_name_or_path": "/llama3/hf/8b_pre_trained",
          "data_config_path": "/testing/tuning/input/apply-custom-template-streaming-data-config.yaml",
          "output_dir": "/testing/tuning/output/llama3-8b/ft/tone_20250129_1045-streaming-dataconfig",
          "save_model_dir": "/testing/tuning/output/llama3-8b/ft/tone_20250129_1045-streaming-dataconfig/save_model",
          "max_steps": 200,
          "per_device_train_batch_size": 4,
          "gradient_accumulation_steps": 1,
          "learning_rate": 1e-4,
          "response_template": "\n### Response:",
          "dataset_text_field": "output"
      }
```

Inference result on `"Text: @sho_help @showtime your arrive is terrible streaming is stop and start every couple mins. Get it together it's xmas\n\n### Label:"`:
```
{
  "responses": [
    {
      "generatedTokenCount": 2,
      "text": " polite\u003c|end_of_text|\u003e",
      "inputTokenCount": 34,
      "stopReason": "EOS_TOKEN",
      "stopSequence": "\u003c|end_of_text|\u003e"
    }
  ]
}

```

seshapad:
[error_log_streaming.txt](https://github.com/user-attachments/files/18599114/error_log_streaming.txt)
@willmj The `streaming` option crashes. I have attached the log for debugging. Here is the data config:
```
dataprocessor:
    type: default
    sampling_stopping_strategy: all_exhausted
    seed: 66
    streaming: true
datasets:
  - name: pleias
    sampling: 1.0
    data_paths:
      - "/cos1/sseshapad/data-mixing-experiments/greek_datasets/raw_data_pleias_greek/data/subset_default/"
    data_handlers:
      - name: apply_dataset_formatting
        arguments:
          remove_columns: ['source_directory', 'domain', 'document', 'subset', 'split', 'document_id', 'identifier', 'collection', 'license', '_meta_timestamp', '_meta_request_url', '_meta_final_url', '_meta_dataset', '_meta_job_id', '_meta_file_name', '_meta_json']
          fn_kwargs:
            dataset_text_field: "contents"
```
I can share the dataset with you if you wish to attempt reproducing this bug.

tuning/data/data_preprocessing_utils.py
```diff
@@ -96,3 +96,14 @@ def get_data_collator(
         raise ValueError(
             "Could not pick a data collator. Please refer to supported data formats"
         )
+
+    # Note that this automatically pads labels with -100
+    # TODO check if this is sufficient for preprocessed
+    # For EPT case where they need packing
+    if is_traindata_tokenized:
```
willmj, line 103: For EPT case where we want tokenized with packing

willmj:
Model trained on single + multi GPU, getting decent results:
```
      {
          "model_name_or_path": "/llama3/hf/8b_pre_trained",
          "data_config_path": "/testing/tuning/input/apply-custom-template-streaming-data-config-10-datasets.yaml",
          "output_dir": "/testing/tuning/output/llama3-8b/ft/tone_20250204_1505-streaming-dataconfig-10-datasets",
          "save_model_dir": "/testing/tuning/output/llama3-8b/ft/tone_20250204_1505-streaming-dataconfig-10-datasets/save_model",
          "max_steps": 2200,
          "per_device_train_batch_size": 4,
          "gradient_accumulation_steps": 1,
          "learning_rate": 1e-4,
          "response_template": "\n### Response:",
          "dataset_text_field": "output"
      }
```
```

**Comment by dushyantbehl** [comment]
Closes https://github.com/foundation-model-stack/fms-hf-tuning/pull/233

File: tuning/sft_trainer.py [hunk 8]
```diff
@@ -332,6 +333,10 @@ def train(
         time.time() - data_preprocessing_time
     )
 
+    if isinstance(formatted_train_dataset, IterableDataset):
+        train_args.accelerator_config = {"split_batches": True}
+        logger.info("Setting split_batches to true - splitting batches among devices")
```
kmehant, line 338: @willmj an important point to be captured in the log is that, since split_batches is being set to True the message should also capture that per_device_train_batch_size would now be global batch size and no more holds that meaning.

tuning/data/data_processors.py
```diff
@@ -233,20 +255,29 @@ def _process_dataset_configs(
             )
             sample_datasets = False
 
+        streaming = False
```
kmehant, line 258: Trying to understand, any reason for introducing this new variable instead of reusing `self.processor_config.streaming`?

willmj: I did it so the code is more readable - however I can see how it could add confusion. Happy to use `self.processor_config.streaming` if you think that would be better.

tuning/sft_trainer.py
willmj: Updated, let me know what you think

tuning/data/data_processors.py
kmehant: IMO, I would try to maintain only one source of information (variable) in a code snippet and only use another when its absolutely necessary (like shallow copy and then operate holding the original piece of information intact etc)

willmj: Yeah that definitely makes sense, thanks for the recommendation. Fixed!

tests/data/test_data_handlers.py
```diff
@@ -84,6 +84,35 @@ def test_apply_custom_formatting_jinja_template():
...
+    tokenizer = AutoTokenizer.from_pretrained(MODEL_NAME)
+    formatted_dataset_field = "formatted_data_field"
+    formatted_dataset = json_dataset.map(
+        apply_custom_data_formatting_template,
+        fn_kwargs={
+            "tokenizer": tokenizer,
+            "dataset_text_field": formatted_dataset_field,
+            "template": template,
+        },
+    )
```
dushyantbehl, line 101: could you add a check here that this is of type `IterableDataset`

tests/data/test_data_preprocessing_utils.py
```diff
@@ -156,7 +161,10 @@ def test_load_dataset_with_hf_dataset(hf_dataset, splitName):
         processor_config=DataPreProcessorConfig(), tokenizer=None
     )
     load_dataset = processor.load_dataset(
-        datasetconfig=datasetconfig, splitName=splitName, datafile=None
+        datasetconfig=datasetconfig,
+        streaming=processor.processor_config.streaming,
+        splitName=splitName,
+        datafile=None,
     )
     assert isinstance(load_dataset, Dataset)
```
dushyantbehl, line 169: could there be a case that this doesn't hold? 
can we make this assert more explicity 

```
assert processor.processor_config.streaming is False and isinstance(load_dataset, Dataset)
```

tests/data/test_data_preprocessing_utils.py
```diff
@@ -687,6 +724,115 @@ def test_process_data_args_throws_error_where_needed(data_args, packing):
...
+    "data_config_path, data_path",
+    [
+        (
+            DATA_CONFIG_YAML_STREAMING_INPUT_OUTPUT,
+            TWITTER_COMPLAINTS_DATA_INPUT_OUTPUT_JSON,
+        ),
+        (DATA_CONFIG_YAML_STREAMING_PRETOKENIZED, TWITTER_COMPLAINTS_TOKENIZED_JSON),
+    ],
+)
+def test_process_streaming_dataconfig_file(data_config_path, data_path):
```
dushyantbehl, line 737: nit: `test_process_dataconfig_file_with_streaming`

tests/data/test_data_preprocessing_utils.py
```diff
@@ -687,6 +724,115 @@ def test_process_data_args_throws_error_where_needed(data_args, packing):
...
+@pytest.mark.parametrize(
+    "data_config_path, data_path",
+    [
+        (
+            DATA_CONFIG_YAML_STREAMING_INPUT_OUTPUT,
+            TWITTER_COMPLAINTS_DATA_INPUT_OUTPUT_JSON,
+        ),
+    ],
+)
+def test_process_streaming_dataconfig_file_no_max_steps_raise_error(
```
dushyantbehl, line 794: nit: `test_process_dataconfig_file_with_streaming_no_max_steps_errors`

tuning/data/data_preprocessing_utils.py
dushyantbehl: I assume we will be dropping this check once https://github.com/foundation-model-stack/fms-hf-tuning/pull/448 is merged..can you make this more explicit in the comment if that is the case.

tuning/data/data_processors.py
```diff
@@ -164,22 +182,26 @@ def _try_load_dataset(dataset_path, dataset_builder):
...
+            return _try_load_dataset(datafile, None, streaming=False)
 
         data_paths = datasetconfig.data_paths
         builder = datasetconfig.builder
         all_datasets = []
 
         for data_path in data_paths:
-            dataset = _try_load_dataset(data_path, builder)
+            dataset = _try_load_dataset(data_path, builder, streaming)
+            if streaming:
```
dushyantbehl, line 203: could we base this check on the type of `Dataset` instead?

tuning/data/data_processors.py
```diff
@@ -263,16 +292,18 @@ def _process_dataset_configs(
                     column_names = raw_datasets[splitName].column_names
 
                     # remove __content__ from all processing
-                    if "__content__" in column_names:
-                        column_names.remove("__content__")
+                    if column_names:
+                        if "__content__" in column_names:
```
dushyantbehl, line 296: can we make this a single line using `and`?

tuning/data/data_processors.py
```diff
@@ -263,16 +292,18 @@ def _process_dataset_configs(
...
+                            column_names.remove("__content__")
 
                     if "remove_columns" not in kwargs:
                         kwargs["remove_columns"] = None
                     if kwargs["remove_columns"] == "all":
                         kwargs["remove_columns"] = column_names
 
-                    if "num_proc" not in kwargs:
-                        kwargs["num_proc"] = os.cpu_count()
+                    if not self.processor_config.streaming:
```
dushyantbehl, line 304: can we also base this check on type of dataset?

dushyantbehl: and if dataset is `Iterable` remove the key `num_proc` with a warning if it exists.

tests/data/test_data_handlers.py
willmj: Made a check for `IterableDatasetDict`

tuning/data/setup_dataprocessor.py
```diff
@@ -45,12 +45,18 @@
...
     if isinstance(data, str):
         # Create a data processor with default processor config
         processor = get_datapreprocessor(
             processor_config=DataPreProcessorConfig(), tokenizer=None
         )
-        data = processor.load_dataset(None, splitName="train[:1]", datafile=data)
+        data = processor.load_dataset(
+            None,
+            streaming=processor.processor_config.streaming,
+            splitName="train[:1]",
```
dushyantbehl, line 57: do we still need to specify streaming if all we do is load just first line of the `train` split?
Can you please check what does HF docs say about this.

tuning/data/data_processors.py
```diff
@@ -322,6 +353,8 @@ def _process_dataset_configs(
                 )
 
         train_dataset = final_datasets.get("train", None)
+        if self.processor_config.streaming:
+            train_dataset = resolve_iterable_dataset_features(train_dataset)
```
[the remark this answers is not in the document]
willmj: Yes, any time after the data is transformed it can lose columns

tuning/data/setup_dataprocessor.py
```diff
@@ -68,6 +75,15 @@ def _process_dataconfig_file(
         tokenizer=tokenizer,
         additional_data_handlers=additional_data_handlers,
     )
+    if processor.processor_config.streaming:
+        if train_args.max_steps < 1:
```
dushyantbehl, line 79: is this always `0` or can this be `None`

tuning/sft_trainer.py
```diff
@@ -333,6 +334,14 @@ def train(
         time.time() - data_preprocessing_time
     )
 
+    if isinstance(formatted_train_dataset, IterableDataset):
+        train_args.accelerator_config = {"split_batches": True}
+        logger.info(
+            "Setting `split_batches` to true - splitting batches among devices \
+                    `per_device_train_batch_size` is now the global batch size, and \
+                    should be treated as such."
+        )
```
dushyantbehl, line 343: While I can live with this check for now but feel like we should be more clear in this and must not waste a run where user's run fails and they scramble through the logs to find this..

Is there a suggestion to make this explicit?

Also...please move this inside `process_dataargs` itself...we do have `train_args` available so can do this inside that function...and do we need to set `accelerator_config` to this dict ... do we not need to append it?

dushyantbehl:
@seshapad @HarikrishnanBalagopal can we take this branch now and shoot an EPT run which ended in error before? 

@willmj has made all the code corrrectness changes so request you to do a sanity check before we go for merge.

seshapad:
@HarikrishnanBalagopal  please provide image for this branch. I will start an ept.

tuning/utils/utils.py
```diff
@@ -51,6 +52,22 @@ def load_yaml_or_json(file_path: str) -> dict:
     return None
 
 
+def resolve_iterable_dataset_features(data: IterableDataset):
```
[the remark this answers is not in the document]
willmj: Since I can imagine it might need to be called outside of the class in future cases I think I will leave it as a util function.

tuning/data/data_preprocessing_utils.py
willmj: Can you elaborate more on why we might not need this after that PR is merged in?

tuning/data/setup_dataprocessor.py
willmj: Per [HF documentation](https://huggingface.co/docs/transformers/en/main_classes/trainer#transformers.TrainingArguments.max_steps) it seems it defaults to `-1` so I would think it can't be `None`. Could still add a check `if train_args.max_steps < 1 or train_args.max_steps is None:`, let me know what you think.

tuning/data/data_processors.py
willmj: I am using `del` to remove it the key

tuning/data/setup_dataprocessor.py
willmj: For checking the columns, it seems fine in unit tests to not pass streaming - however it does load the example as a `Dataset` instead of an `IterableDataset`. If this is okay with you we can either pass in `streaming` through `kwargs` of `load_dataset`, default `streaming` to false in `load_dataset`, or just set it to `false` for loading this. Let me know what you think will work best.

tuning/sft_trainer.py
willmj: Yes per Mehant's [suggestion](https://github.com/foundation-model-stack/fms-hf-tuning/pull/437#discussion_r1933227645) we set accelerator_config to this dict. You bring up a good point that documentation should be added for this PR, I will add it soon.

willmj: I added documentation in `advanced-data-preprocessing.md`, and made the warning more explicit.

tuning/data/data_preprocessing_utils.py
dushyantbehl: My point was if packing support is present for tokenized data in TRL we do not need any collator there I assume....we just need to feed tokenized dataset with packing to the trainer for EPT was my assumption..like we do for non tokenized datasets.

tuning/data/setup_dataprocessor.py
dushyantbehl: If for checking columns a single sample can be loaded without streaming you can choose that route and force disable streaming in this call..I would be fine with it

What my question was to ask if a single sample can be loaded in all cases without performance considerations even for large datasets...so I wanted to ask if 1) HF load the only 1 sample from disk? 2) HF loads all samples and then drops all but one
In (2) the performance can take a hit.

tuning/data/setup_dataprocessor.py
dushyantbehl: if the default is `-1` we can leave it like this.

tuning/data/setup_dataprocessor.py
willmj: It seems from HF documentation on [slice splits](https://huggingface.co/docs/datasets/v3.2.0/loading#slice-splits) that HF load dataset goes for number 1.

tuning/data/data_preprocessing_utils.py
dushyantbehl: @kmehant can you check if this understanding is correct?

kmehant: @willmj tokenized with packing is added in [v0.13.0](https://github.com/huggingface/trl/releases/tag/v0.13.0) release of TRL. So we should not be be supporting this until we have https://github.com/foundation-model-stack/fms-hf-tuning/pull/448 merged. So it would anyway error out here https://github.com/foundation-model-stack/fms-hf-tuning/blob/f88f031a09604cc86c529b468792f12699f72a2a/tuning/data/setup_dataprocessor.py#L99-L103 even before coming to this piece of code.

kmehant: So we can remove this whole chunk of code on this file since it would any wont be executed.

willmj:
According to [this comment](https://github.com/foundation-model-stack/fms-hf-tuning/pull/437#discussion_r1948579043) once trl is upgraded the case highlighted shouldn't be needed so I have removed it, which may cause some tests to fail while trl is not merged. This means this PR is waiting on [upgrading TRL](https://github.com/foundation-model-stack/fms-hf-tuning/pull/448).

willmj:
Removing the following test case:
```
@pytest.mark.parametrize(
    "datafiles, datasetconfigname",
    [
        (
            [TWITTER_COMPLAINTS_TOKENIZED_JSON],
            DATA_CONFIG_YAML_STREAMING_PRETOKENIZED,
        ),
    ],
)
def test_run_causallm_ft_and_inference_streaming_ept(datasetconfigname, datafiles):
    """Check if we can finetune causallm models using multiple datasets with multiple files"""
    with tempfile.TemporaryDirectory() as tempdir:
        data_formatting_args = copy.deepcopy(DATA_ARGS)

        # set training_data_path and response_template to none
        data_formatting_args.response_template = None
        data_formatting_args.training_data_path = None

        # add data_paths in data_config file
        with tempfile.NamedTemporaryFile(
            "w", delete=False, suffix=".yaml"
        ) as temp_yaml_file:
            with open(datasetconfigname, "r", encoding="utf-8") as f:
                data = yaml.safe_load(f)
                datasets = data["datasets"]
                for _, d in enumerate(datasets):
                    d["data_paths"] = datafiles
                yaml.dump(data, temp_yaml_file)
                data_formatting_args.data_config_path = temp_yaml_file.name

        train_args = copy.deepcopy(TRAIN_ARGS)
        train_args.output_dir = tempdir
        train_args.max_steps = 1
        train_args.packing = True

        sft_trainer.train(MODEL_ARGS, data_formatting_args, train_args)

        # validate full ft configs
        _validate_training(tempdir)
        _, checkpoint_path = _get_latest_checkpoint_trainer_state(tempdir)

        # Load the model
        loaded_model = TunedCausalLM.load(checkpoint_path, MODEL_NAME)

        # Run inference on the text
        output_inference = loaded_model.run(
            "### Text: @NortonSupport Thanks much.\n\n### Label:", max_new_tokens=50
        )
        assert len(output_inference) > 0
        assert "### Text: @NortonSupport Thanks much.\n\n### Label:" in output_inference
```
as it will be resolved by #468

docs/advanced-data-preprocessing.md
```diff
@@ -251,6 +252,17 @@ We also allow users to pass a [`seed`](https://huggingface.co/docs/datasets/v3.2
...
+
+When using streaming, `split_batches` in the `TrainingArguments` will automatically be set to `True`, by doing so, the main process will fetch a full batch and slice it into `num_processes` batches for each process. This means that `num_processes` must be divisible by `batch_size`. This will replace the global batch size.
+
+When using streaming, the user must set `max_steps` in the `TrainingArguments` instead of `num_train_epochs`. Since iterable datasets are loaded chunk-by-chunk, data cannot run through epochs in a typical fashion as the preprocessor can not know length of the dataset as it is being passed through. If both `max_steps` and `num_train_epochs` are given in a training config, `max_steps` will overwrite `num_train_epochs`.
```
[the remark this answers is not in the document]
dushyantbehl: Also can we have a configuration right here in the document for example say..use this type of data config and set these arguments for the training to work.
You can also link this to the ept document that we have.

docs/advanced-data-preprocessing.md
```diff
@@ -251,6 +252,24 @@ We also allow users to pass a [`seed`](https://huggingface.co/docs/datasets/v3.2
...
+In a data config this looks like (see [ept document](./ept.md#large-non-tokenized-dataset) for a more in-depth example):
+```
+dataprocessor:
+    type: default
+    streaming: true
+```
+
+When using streaming, `split_batches` in the `TrainingArguments` will automatically be set to `True`, by doing so, the main process will fetch a full batch and slice it into `num_processes` batches for each process. This means that `num_processes` must be divisible by `batch_size`. This will replace the global batch size.
+
+**When using streaming, the user must set `max_steps` in the `TrainingArguments` instead of `num_train_epochs`.** Since iterable datasets are loaded chunk-by-chunk, data cannot run through epochs in a typical fashion as the preprocessor can not know length of the dataset as it is being passed through. If both `max_steps` and `num_train_epochs` are given in a training config, `max_steps` will overwrite `num_train_epochs` since `max_steps` directly specifies the total number of optimization steps, which is needed when dataset length cannot be known.
```
Abhishek-TAMU, line 269: nit:

Current:
`data cannot run through epochs in a typical fashion as the preprocessor can not know length of the dataset as it is being passed through.`

Fix:
`data cannot run through epochs in a typical fashion as the **Trainer** can not know length of the dataset as it is being passed through.`

docs/advanced-data-preprocessing.md
```diff
@@ -251,6 +252,24 @@ We also allow users to pass a [`seed`](https://huggingface.co/docs/datasets/v3.2
...
+dataprocessor:
+    type: default
+    streaming: true
+```
+
+When using streaming, `split_batches` in the `TrainingArguments` will automatically be set to `True`, by doing so, the main process will fetch a full batch and slice it into `num_processes` batches for each process. This means that `num_processes` must be divisible by `batch_size`. This will replace the global batch size.
+
+**When using streaming, the user must set `max_steps` in the `TrainingArguments` instead of `num_train_epochs`.** Since iterable datasets are loaded chunk-by-chunk, data cannot run through epochs in a typical fashion as the preprocessor can not know length of the dataset as it is being passed through. If both `max_steps` and `num_train_epochs` are given in a training config, `max_steps` will overwrite `num_train_epochs` since `max_steps` directly specifies the total number of optimization steps, which is needed when dataset length cannot be known. 
+
+`max_steps` can be found by dividing the number of samples in a dataset by the batch size.
```
Abhishek-TAMU, line 271: nit:
`If the dataset size is known to the user, max_steps can be calculated as the total number of samples divided by the batch size.`

docs/ept.md
```diff
@@ -107,6 +107,35 @@ Here also the command line arguments would be
 
 The code again would add `EOS_TOKEN` to the non tokenized data before using it and also note that the `dataset_text_field` is assumed to be same across all datasets for now.
 
+### Large Non-Tokenized Dataset
+Lets say you have a large `JSONL` data file which cannot all fit into memory at once that you want to perform EPT on, you can create a `data_config` for the dataset in this manner,
```
Abhishek-TAMU, line 111: nit: Probably something like this ?

`Let's say you have a large JSONL data file that cannot all fit into memory at once and you want to perform EPT on it, you can use the streaming feature to efficiently load and process data in chunks. To enable streaming, you can define a data_config as follows:`

docs/ept.md
```diff
@@ -107,6 +107,35 @@ Here also the command line arguments would be
...
+        data_handlers:
+        - name: add_tokenizer_eos_token
+            arguments:
+            remove_columns: all
+            batched: false
+            fn_kwargs:
+                dataset_text_field: "dataset_text_field"
+```
+
+And the commandline passed to the library should include following.
```
Abhishek-TAMU, line 131: nit: `The command-line arguments passed to the library should include the following:`

Abhishek-TAMU (commented):
Looks good to me. Just some documentation nits :)

tuning/data/data_processors.py
```diff
@@ -283,16 +312,28 @@ def _process_dataset_configs(
...
                         column_names.remove("__content__")
 
                     if "remove_columns" not in kwargs:
                         kwargs["remove_columns"] = None
                     if kwargs["remove_columns"] == "all":
                         kwargs["remove_columns"] = column_names
 
-                    if "num_proc" not in kwargs:
-                        kwargs["num_proc"] = os.cpu_count()
+                    if not isinstance(raw_datasets, IterableDatasetDict):
```
dushyantbehl, line 327: minor nit: 
Could we just reverse this check for easier readablity?

```suggestion
                    if isinstance(raw_datasets, IterableDatasetDict):
                        if "num_proc" in kwargs:
                            del kwargs["num_proc"]
                            logger.warning(
                                "num_proc is not applicable for \
                                            IterableDatasets and has been removed."
                            )
                    else:
                        if "num_proc" not in kwargs:
                            kwargs["num_proc"] = os.cpu_count()
                            logger.info(f"setting num_proc to {os.cpu_count()}")
```

dushyantbehl:
LGTM

dushyantbehl (approved):

dushyantbehl:
@Abhishek-TAMU @ashokponkumar please review and merge

PR #438, opened 2025-01-15
feat: Add support for jinja based template rendering of the dataset

Abhishek-TAMU:
### Description of the change

Added a handler `apply_custom_data_formatting_jinja_template` which does jinja based template rendering of the dataset. 

Handling of edge case:
`Example template:"### Input: {{Tweet text}} \n\n ### Response: {{text_label}}"`
Jinja2 by default, does not support placeholders variable names with spaces (e.g., {{Tweet text}}), which will raise an error. 

Hence additional preprocessing check (function: `transform_placeholders`) has been done. This checks if there is space between the placeholder variable and then process it accordingly (by modifying variable by `{{element["Tweet text"]}}`.

### Related issue number

Issue: https://github.ibm.com/ai-foundation/watson-fm-stack-tracker/issues/1470

### How to verify the PR

Verify added test cases.

### Was the PR tested


tuning/data/data_handlers.py
```diff
@@ -137,6 +138,65 @@ def replace_text(match_obj):
...
+    template = transform_placeholders(template)
+    env = Environment(undefined=StrictUndefined)
+    jinja_template = env.from_string(template)
+
+    try:
+        rendered_text = jinja_template.render(element=element, **element)
+    except Exception as e:
+        raise KeyError(f"Dataset does not contain field in template. {e}") from e
+
+    rendered_text += tokenizer.eos_token
```
Abhishek-TAMU, line 168: @ashokponkumar Wanted to just confirm the removal of `eos_token` from dataset samples in this handler. In other handlers we add `eos_token` and don't expect users to add it. Hence in this handler where user passes Jinja template are we expecting user to pass `eos_token` too? I guess in case of non-pretokenized dataset not using `eos_token` when using `DataCollatorForCompletionOnlyLM` might affect F1 score on tuned models ?


2-  @dushyantbehl Can I ask how Jinja templating could be used with pre-tokenized dataset (Having `input_ids` and `labels` as columns) ?

dushyantbehl: 1. I think we need a proper documentation for now and a patch where we let users choose if they want an `eos_token` with the data handlers or not via one argument e.g. add a `kwarg` to the data handlers like `add_eos_token` this way we can let them choose what they want inside a data config. 
 for a data config we should not assume things like what should we do while users want to do. 
 for our data args we can have this added inside our code at the last data handler whatever we choose so that our data args usecases remain same. 
 
 if you feel can you take this up with this patch? to add the kwarg for `eos_token` to clean up the interface with users? else we can park this to a next patch. 
 
 2. For pre tokenised datasets we can ignore the jinja template imo this should be applied only to non tokeniser data sets . 
 We can add all these things to documentation and I request you to please add documentation with this patch.

tests/data/test_data_handlers.py
```diff
@@ -76,6 +104,25 @@ def test_apply_custom_formatting_template_gives_error_with_wrong_keys():
...
 
+def test_apply_custom_formatting_jinja_template_gives_error_with_wrong_keys():
+    """Tests that the jinja formatting function will throw error if wrong keys are passed to template"""
+    json_dataset = datasets.load_dataset(
+        "json", data_files=TWITTER_COMPLAINTS_DATA_JSONL
+    )
+    template = "### Input: {{not found}} \n\n ### Response: {{text_label}}"
+    formatted_dataset_field = "formatted_data_field"
+    tokenizer = AutoTokenizer.from_pretrained(MODEL_NAME)
+    with pytest.raises((KeyError, TemplateSyntaxError)):
```
dushyantbehl, line 115: can we catch this error inside our code and give users a simple text error?

tuning/data/data_handlers.py
```diff
@@ -137,6 +138,65 @@ def replace_text(match_obj):
...
+
+    return {dataset_text_field: rendered_text}
+
+
+def transform_placeholders(template: str) -> str:
```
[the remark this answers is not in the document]
dushyantbehl: I think we were only handling non nested datasets apart from chat templates...can we test things out with this patch if our code works for nested datasets then can we have a change of the argument type here?

tuning/data/data_handlers.py
Abhishek-TAMU: As per offline discussion, addition of kwarg `add_eos_token` would be done part of this [issue](https://github.ibm.com/ai-foundation/watson-fm-stack-tracker/issues/1471) and hence documentation of the same would also be taken care.

Though handler documentation is added in this PR.

tuning/data/data_handlers.py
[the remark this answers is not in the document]
Abhishek-TAMU: As per offline discussion, handling of nested dataset would be checked and done for all handlers as part of this [issue](https://github.ibm.com/ai-foundation/watson-fm-stack-tracker/issues/1471).

> Also if you can move to utils as we discussed in our last call. Thanks.

Done

tests/data/test_data_handlers.py
Abhishek-TAMU: `TemplateSyntaxError` is not needed anymore as this error comes when there is a space between placeholder variable in the template, and we are handling the space now with `transform_placeholders` utils function.

For `KeyError` the text error for user is mentioned in the handler `apply_custom_data_formatting_jinja_template`.

tests/data/test_data_handlers.py
```diff
@@ -16,6 +16,7 @@
 # https://spdx.dev/learn/handling-license-info/
 
 # Third Party
+from jinja2.exceptions import TemplateSyntaxError
```
dushyantbehl, line 19: Can you remove this import as its not used

tuning/data/data_handlers.py
```diff
@@ -137,6 +141,37 @@ def replace_text(match_obj):
...
+    tokenizer: AutoTokenizer,
+    dataset_text_field: str,
+    template: str,
+    **kwargs,
+):
+    """Function to format datasets with jinja templates.
+       Expects to be run as a HF Map API function.
+    Args:
+        element: the HF Dataset element loaded from a JSON or DatasetDict object.
+        dataset_text_field: formatted_dataset_field.
```
dushyantbehl, line 155: Please add `tokenizer` to the args here.

dushyantbehl: Also I know this is not on you but can you please fix the doc string for line 104 as well.

tuning/utils/config_utils.py
```diff
@@ -135,3 +136,34 @@ def txt_to_obj(txt):
     except UnicodeDecodeError:
         # Otherwise the bytes are a pickled python dictionary
         return pickle.loads(message_bytes)
+
+
+def transform_placeholders(template: str) -> str:
```
dushyantbehl, line 141: could we please rename this function to be more descriptive? 

sanitise jinja placeholders?

dushyantbehl:
Suggested minor changes and barring those LGTM. 
@willmj please share your review post that you are good to merge @Abhishek-TAMU

Abhishek-TAMU:
Made the suggested changes. @willmj Feel free to give a final review.

willmj (approved):
LGTM! Thanks Abhishek

dushyantbehl:
@ashokponkumar please review and merge. Thanks

tuning/data/data_handlers.py
```diff
@@ -137,6 +143,39 @@ def replace_text(match_obj):
...
+        template: Template to format data with. Features of Dataset
+            should be referred to by {{key}}.
+    Returns:
+        Formatted HF Dataset
+    """
+
+    template += tokenizer.eos_token
+    template = process_jinja_placeholders(template)
+    env = Environment(undefined=StrictUndefined)
```
kmehant, line 168: Can we use `SandboxedEnvironment` instead? We should avoid Environment as much as possible.

kmehant (changes requested):

tuning/data/data_handlers.py
Abhishek-TAMU: Thanks for the suggestion @kmehant. `SandboxedEnvironment` definitely seems to be good in terms of security, though the time taken for Jinja rendering (or `.map()` of Handler) is slightly more in this case due to internal checks, but not very significant difference.

@ashokponkumar @dushyantbehl @willmj If it sounds good to go ahead with this, I created a [draft PR](https://github.com/foundation-model-stack/fms-hf-tuning/pull/456) to include usage of `SandboxedEnvironment` with additional error handling.

PR #447, opened 2025-01-23
docs: add note to note that file extension is required in training data path

willmj:
### Description of the change
With changes to data preprocessor, file extension is now required for data files because of the variety of datasets we support (as opposed to just supporting JSON, which made it so file extension wasn't required because it could be assumed).

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,7 +80,7 @@ ARROW       |   ✅
 
 As iterated above, we also support passing a HF dataset ID directly via `--training_data_path` argument.
 
-Because of the variety of data formats and file types supported, if `--training_data_path` is a file, it requires a file extension in the path name. If `--training_data_path` does not have a file extension, the data preprocessor will check if it is a valid folder. If `--training_data_path` is not a folder and does not have a file extension, it is considered an HF dataset ID.
+**NOTE**: Due to the variety of supported data formats and file types, when --training_data_path points to a file, it must include a file extension (e.g., .json, .csv). If the path does not include a file extension, it will be treated either as a folder or as an HF dataset ID.
 
 ## Use cases supported with `training_data_path` argument
 
--- a/README.md
+++ b/README.md
@@ -80,7 +80,10 @@ ARROW       |   ✅
 
 As iterated above, we also support passing a HF dataset ID directly via `--training_data_path` argument.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -80,6 +80,8 @@ ARROW       |   ✅
 
 As iterated above, we also support passing a HF dataset ID directly via `--training_data_path` argument.
 
+Because of the variety of data formats and file types supported, if `--training_data_path` is a file, it requires a file extension in the path name. If `--training_data_path` does not have a file extension, the data preprocessor will check if it is a valid folder. If `--training_data_path` is not a folder and does not have a file extension, it is considered an HF dataset ID.
```
Abhishek-TAMU, line 83: Thanks Will @willmj for the quick PR. 
Just a suggestion that we can add this info as a `Note`. Also below is slightly reworded version as I am not sure if the flow of code checks we are doing needed to be mentioned in the documentation. 

`NOTE: Due to the variety of supported data formats and file types, when --training_data_path points to a file, it must include a file extension (e.g., .json, .csv). If the path does not include a file extension, it will be treated either as a folder or as an HF dataset ID. `

willmj: Thanks for the suggestion and reword, updated!

README.md
```diff
@@ -80,6 +80,8 @@ ARROW       |   ✅
 
 As iterated above, we also support passing a HF dataset ID directly via `--training_data_path` argument.
 
+**NOTE**: Due to the variety of supported data formats and file types, when --training_data_path points to a file, it must include a file extension (e.g., .json, .csv). If the path does not include a file extension, it will be treated either as a folder or as an HF dataset ID.
```
dushyantbehl, line 83: Can you change the wording according to our code

- if a valid folder, it will be treated as a folder.
- if it has a valid extention it will be treated like a file.
- else we try to load it as a HF dataset id.

Abhishek-TAMU:
```
**NOTE**: Due to the variety of supported data formats and file types, --training_data_path is handled as follows:

- If it points to a file with a valid extension (e.g., .json, .csv), it is treated as a file.
- If it points to a valid folder, it is treated as a folder.
- else we try to load it as a Hugging Face (HF) dataset ID.
```


I think this looks good right ? @dushyantbehl

willmj: I think that looks good, I added a bit more clarifying wording

Abhishek-TAMU (approved):
LGTM! Feel free to update the branch with base.

PR #448, opened 2025-01-23
chore(deps): upgrade trl and transformers

willmj:
### Description of the change
[requirements.txt](https://github.com/huggingface/trl/blob/v0.13.0/requirements.txt) in trl v0.13.0 no longer restricts transformers version.

### Related issue number
[More context](https://github.ibm.com/ai-foundation/watson-fm-stack-tracker/issues/1489)

### Was the PR tested


pyproject.toml
```diff
@@ -33,7 +33,7 @@ dependencies = [
 "sentencepiece>=0.1.99,<0.3",
 "tokenizers>=0.13.3,<1.0",
 "tqdm>=4.66.2,<5.0",
-"trl>=0.9.3,<0.12",
+"trl>=0.9.3,<1.0",
```
dushyantbehl, line 36: @willmj do we want to probably lock the version to higher than one which supports packing for pretokenized datasets which we enabled along with this patch.

willmj: @dushyantbehl I thought the latest version was 0.13.0, so this would include that. Feel free to make a suggested change if I'm misunderstanding.

willmj:
Dependent on [fms-acceleration #98](https://github.com/foundation-model-stack/fms-acceleration/issues/98)

dushyantbehl:
> I think if we want TRL version to be greater than `>=0.12 ` then I guess we need to first ensure `fms-hf-tuning` is compatible with transformers version `>=4.46` which I guess currently is not, as per this [issue](https://github.ibm.com/ai-foundation/watson-fm-stack-tracker/issues/1489), because I can see TRL latest version has minimum version of transformers as 4.46 [requirements.txt](https://github.com/huggingface/trl/blob/v0.13.0/requirements.txt)

Correct @Abhishek-TAMU that's what I wanted to highlight too thanks

dushyantbehl:
> Upgrade of transformers >=4.46 is dependent on [fms-acceleration #98](https://github.com/foundation-model-stack/fms-acceleration/issues/98), also have to update SFTConfig correctly

I see good to note both here and we can update this later post we get working our repo with the new transformers..thanks

pyproject.toml
ashokponkumar:
```suggestion
"trl>=0.13.0,<1.0",
```

@dushyantbehl is this your suggestion?

dushyantbehl:
@anhuong thanks for quick fix on https://github.com/foundation-model-stack/fms-acceleration/pull/123

Does this mean we can go ahead and move forward with this change? cc @willmj @fabianlim

Abhishek-TAMU:
> I think if we want TRL version to be greater than `>=0.12 ` then I guess we need to first ensure `fms-hf-tuning` is compatible with transformers version `>=4.46` which I guess currently is not, as per this [issue](https://github.ibm.com/ai-foundation/watson-fm-stack-tracker/issues/1489), because I can see TRL latest version has minimum version of transformers as 4.46 [requirements.txt](https://github.com/huggingface/trl/blob/v0.13.0/requirements.txt)

> Does this mean we can go ahead and move forward with this change?

As both `transformers` and `trl` version are being updated, I guess it would also be good to run 1 automated e2e test (for all models) to once confirm the successful working with the upgraded version of trl and transformers. @willmj

Abhishek-TAMU:
@willmj 

All the failed tests (41 failed tests) would be completely resolved by these 3 changes:

1- Setting `gradient_accumulation_steps = 1` here in [test_sft_trainer.py](https://github.com/foundation-model-stack/fms-hf-tuning/blob/381fdd55d42d2fffbf8ef0ed53d55bfe83373924/tests/test_sft_trainer.py#L90).

As the size of training dataset is 10 and batch_size is 4, hence due to `gradient_accumulation_steps` = 4 not all optimizer steps is run and hence its not running for entire 5 epochs and hence not even completing 1 epoch, hence no training_log json file is not generated (In unit tests with not found errors).

2- Same reasoning, to change `gradient_accumulation_steps = 1` in [test_launch_script.py](https://github.com/foundation-model-stack/fms-hf-tuning/blob/381fdd55d42d2fffbf8ef0ed53d55bfe83373924/tests/build/test_launch_script.py#L49) 

3- Use [_get_checkpoint_path](https://github.com/foundation-model-stack/fms-hf-tuning/blob/381fdd55d42d2fffbf8ef0ed53d55bfe83373924/tests/test_sft_trainer.py#L1144) to get last checkpoint instead of `checkpoint-5` as not every tuning has this checkpoint. Something like this:

```
def _get_checkpoint_path(dir_path):
    checkpoint_dirs = [
        d for d in os.listdir(dir_path)
        if os.path.isdir(os.path.join(dir_path, d)) and re.match(r'^checkpoint-\d+$', d)
    ]
    checkpoint_dirs.sort(key=lambda name: int(name.split('-')[-1]))
    return os.path.join(dir_path, checkpoint_dirs[-1])
```

willmj:
New [release](https://github.com/foundation-model-stack/fms-acceleration/releases/tag/v0.6.1) of fms-acceleration unblocking transformer 4.46. Should we go for merging?
@anhuong @dushyantbehl @Abhishek-TAMU

dushyantbehl:
@willmj I applied the fixes to test cases suggested by @Abhishek-TAMU above and have opened a PR on reverting the packing for pretokenized dataset restirctions https://github.com/foundation-model-stack/fms-hf-tuning/pull/468
This is based on your PR.

anhuong:
Running a few tests of tuning with the updated versions, but the changes look good to me. Once my tests run, I will note in this PR

anhuong:
Full fine-tuning and LoRA tuning ran fine with both on minimal example. We will have to test to see how these changes affect train_runtime and quality in the next release though @Abhishek-TAMU

anhuong (approved):

anhuong:
Hmm but the changes from main broke unit tests so will investigate...

anhuong:
New version of trl v0.15.0 was released 3 hours ago and there are not yet [release docs](https://github.com/huggingface/trl/releases/tag/v0.15.0). This upgrade broke the unit tests, setting upper limit to below 0.15.0 for now

Abhishek-TAMU (approved):
Changes LGTM! Thanks for the PR @willmj 
Thanks @anhuong for testing it with minimal example.

PR #449, opened 2025-01-24
Bump python version

dushyantbehl:
### Was the PR tested

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,15 +15,12 @@ authors = [
 ]
 license = {text = "Apache-2.0"}
 readme = "README.md"
-requires-python = "~=3.9"
+requires-python = ">=3.12"
 keywords = ['fms-hf-tuning', 'python', 'tuning']
 classifiers=[
     "License :: OSI Approved :: Apache Software License",
     "Development Status :: 4 - Beta",
-    "Programming Language :: Python :: 3",
-    "Programming Language :: Python :: 3.9",
-    "Programming Language :: Python :: 3.10",
-    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12"
 ]
 dependencies = [
 "numpy>=1.26.4,<2.0",
--- a/.github/workflows/build-and-publish.yaml
+++ b/.github/workflows/build-and-publish.yaml
@@ -24,8 +24,8 @@ jobs:
     strategy:
       matrix:
         python-version:
-          - setup: "3.11"
-            tox: "py311"
+          - setup: "3.12"
+            tox: "py312"
 
     environment:
       name: pypi
--- a/.github/workflows/coverage.yaml
+++ b/.github/workflows/coverage.yaml
@@ -10,10 +10,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - name: Set up Python 3.11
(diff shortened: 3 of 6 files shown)


willmj (commented):
Thanks for the quick fix and testing Dushyant! Apologies for the lack of understanding but could you clarify what prevents us from keeping a minimum version of 3.9 for users who may want to use it?

dushyantbehl:
> Thanks for the quick fix and testing Dushyant! Apologies for the lack of understanding but could you clarify what prevents us from keeping a minimum version of 3.9 for users who may want to use it?

@willmj  Anything less than v3.12 has a FIPS security compliance issue which stems from underlying python version...if security is not a concern we can keep multiple versions one internal to the product where we go with the compliant 3.12+ python else we keep it open to 3.9 users like we had already.

willmj (dismissed):
That makes sense, thanks! This change seems good to me for now, but let's wait for @anhuong's approval before merging.

willmj:
Testing results on python 3.12 version:
|                     | 3.12 Baseline runtime | 3.12 Runtime with flags set | 3.12 Baseline MF1 Score | 3.12 MF1 score with flags set |
|---------------------|-----------------------|-----------------------------|-------------------------|-------------------------------|
| Granite 3.1 8b FT   | 3257                  | 3094                        | 0.590                   | 0.641                         |
| Granite 3.1 8b LoRA | 1033                  | 1043                        | 0.610                   | 0.605                         |
| Llama 3.1 8b FT     | 3482                  | 2956                        | 0.596                   | 0.621                         |
| Llama 3.1 8b LoRA   | 1091                  | 1092                        | 0.577                   | 0.568                         |
| PowerLM 3b QLoRA    | 1027                  | 630                         | 0.607                   | 0.589                         |

for comparison, this is what we saw on python 3.11
|                     | Baseline runtime | Runtime with flags set | Baseline MF1 Score | MF1 score with flags set |
|---------------------|------------------|------------------------|--------------------|--------------------------|
| Granite 3.1 8b FT   | 4476             | **3520**                   | 0.589              | **0.561**                    |
| Granite 3.1 8b LoRA | 1102             | **1094**                   | 0.599              | **0.601**                    |
| Llama 3.1 8b FT     | 4066             | **3391**                   | 0.596              | **0.645**                    |
| Llama 3.1 8b LoRA   | 1110             | **991**                    | 0.587              | **0.588**                    |

pyproject.toml
```diff
@@ -15,7 +15,7 @@ authors = [
 ]
 license = {text = "Apache-2.0"}
 readme = "README.md"
-requires-python = "~=3.9"
+requires-python = ">=3.12"
```
willmj, line 18: Should we change this back as well?
```suggestion
requires-python = "~=3.9"
```

willmj (changes requested):

pyproject.toml
anhuong: thanks for catching this!

willmj (approved):
LGTM!

aluu317 (approved):

anhuong (approved):
Updating python to 3.12 for fips compliance, image and library will be built with python 3.12 but support will remain for python 3.9-3.12

praveenj83:
Thanks everyone for the quick turnaround on this!

PR #451, opened 2025-01-26
feat: support loading vision model

anhuong:
### Was the PR tested


pyproject.toml
```diff
@@ -28,16 +28,17 @@ classifiers=[
...
 "numpy>=1.26.4,<2.0",
 "accelerate>=0.20.3,!=0.34,<1.1",
-"transformers>=4.45,<4.46",
+"transformers",
 "torch>=2.2.0,<2.5",
 "sentencepiece>=0.1.99,<0.3",
 "tokenizers>=0.13.3,<1.0",
 "tqdm>=4.66.2,<5.0",
-"trl>=0.9.3,<0.12",
+"trl==0.13",
```
anhuong, line 36: Note that both of these changes are waiting on a fix in fms-acceleration, otherwise the cross entropy fused ops will break with newer versions of transformers past 4.46. Also should update to give transformers upper bound version

tuning/config/configs.py
```diff
@@ -61,6 +61,10 @@ class ModelArguments:
                 tokenizer classes."
         },
     )
+    multimodal: bool = field(
```
anhuong, line 64: Is there a better way to tell if a user is loading a vision model and to then load the processor and load the model with `AutoModelForVision2Seq`? Should update this to vision specific flag

tuning/config/configs.py
```diff
@@ -122,6 +126,20 @@ class DataArguments:
...
         },
     )
+    text_field_name: str = field(
+        default=None,
+        metadata={
+            "help": "Required for running with vision models. \
+            The column name of the text data in the multi-modal dataset."
+        },
+    )
+    image_field_name: str = field(
```
anhuong, line 136: Will comment on these fields in the data collator section -- questions around how to load a multimodal dataset that contains images and text and how to preprocess the text portions

tuning/data/data_preprocessing_utils.py
```diff
@@ -96,3 +117,48 @@ def get_data_collator(
...
+        returns a batch of processed data
+        """
+        # Get the texts and images, and apply the chat template
+        texts = [
+            self.processor.apply_chat_template(
+                example[self.text_field_name], tokenize=False
+            )
+            for example in examples
+        ]
+        images = [example[self.image_field_name] for example in examples]
```
anhuong, line 141: @dushyantbehl This is the key part I need your input on -- this works well for a dataset like https://huggingface.co/datasets/HuggingFaceH4/llava-instruct-mix-vsft where there is a single column for images and for text. However, how do we want to preprocess datasets with multiple text fields that need to be concatenated together like https://huggingface.co/datasets/NSTiwari/DocumentIDEFICS_QA where the query and answer are separate columns, the data collator should look more like
```py
        for example in examples:
            image = example["image"]
            question = example["query"]['en']
            answer = random.choice(example["answers"])
            messages = [
                {
                    "role": "user",
                    "content": [
                        {"type": "text", "text": "Answer briefly."},
                        {"type": "image"},
                        {"type": "text", "text": question}
                    ]
                },
                {
                    "role": "assistant",
                    "content": [
                        {"type": "text", "text": answer}
                    ]
                }
            ]
            text = processor.apply_chat_template(messages, add_generation_prompt=False)
```

but how do we accommodate both forms and not knowing what the column field names are?

Or even further https://huggingface.co/datasets/philschmid/amazon-product-descriptions-vlm where there are multiple which contains many fields regarding amazon product information should look like
```py
    return {"messages": [
                {
                    "role": "system",
                    "content": [{"type": "text", "text": system_message}],
                },
                {
                    "role": "user",
                    "content": [
                        {
                            "type": "text",
                            "text": prompt.format(product_name=sample["Product Name"], category=sample["Category"]),
                        },{
                            "type": "image",
                            "image": sample["image"],
                        }
                    ],
                },
                {
                    "role": "assistant",
                    "content": [{"type": "text", "text": sample["description"]}],
                },
            ],
        }
```
where only some fields are taken -- is this something the user would have to provide their own collator for?

tuning/data/setup_dataprocessor.py
```diff
@@ -289,11 +290,18 @@ def _process_raw_data_args(
         eval_dataset_config.data_handlers = handlers
 
     # And let processor handle the logic
-    train_dataset = data_processor.process_dataset_configs([train_dataset_config])
+    train_dataset = None
+    if processor:
+        train_dataset = load_dataset(data_args.training_data_path, split="train")
+    else:
+        train_dataset = data_processor.process_dataset_configs([train_dataset_config])
```
anhuong, line 297: @dushyantbehl please help in this area for verifying how this data preprocessing would look if the multimodal dataset went through the `process_data_configs`

anhuong (commented):
Tested this code loading llama 3.2-11b vision model as well as llava 1.6-mistral-7b vision model and they both were able to be lora tuned successfully with dataset https://huggingface.co/datasets/HuggingFaceH4/llava-instruct-mix-vsft

Note that when loading llava model with FSDP, need to provide extra field `fsdp_transformer_layer_cls_to_wrap: "LlamaDecoderLayer"` for llava 1.5 and `fsdp_transformer_layer_cls_to_wrap: "MistralDecoderLayer"` for llava 1.6-mistral

Ran with configuration:
```json
{
  "model_name_or_path": "llava-hf/llava-v1.6-mistral-7b-hf", 
  "training_data_path": "HuggingFaceH4/llava-instruct-mix-vsft",
  "output_dir": "/fmaas-integration-tests/tuning/output/anhuong/llava1.6-mistral-7b-vision_llava-dataset_lora",
  "num_train_epochs": 3.0,
  "per_device_train_batch_size": 4,
  "gradient_accumulation_steps": 1,
  "learning_rate": 1e-4,
  "response_template": "\n### Response:",       <--- FIX: this field is not used
  "dataset_text_field": "output",       <--- FIX: this field is not used
  "bf16": true,
  "torch_dtype": "bfloat16",
  "use_flash_attn": false,
  "remove_unused_columns": false,
  "dataset_kwargs": {"skip_prepare_dataset": true},
  "multimodal": true,
  "peft_method": "lora",
  "r": 8,
  "lora_dropout": 0.05,
  "lora_alpha": 16,
  "target_modules": ["all-linear"],
  "lora_post_process_for_vllm": true,
  "gradient_checkpointing": true,
  "text_field_name": "messages",
  "image_field_name": "images"
}
```

tuning/data/data_preprocessing_utils.py
```diff
@@ -29,6 +32,9 @@ def get_data_collator(
     is_traindata_tokenized: bool,
     max_seq_length: int,
     instruction_template: Optional[str],
+    text_field_name: Optional[str],
+    image_field_name: Optional[str],
+    processor=None,
```
dushyantbehl, line 37: I am assuming with the term `processor` you mean [this](https://huggingface.co/docs/transformers/en/main_classes/processors) which is the new name for both image processor and tokenizers in the new HF code. if so can you please update to remove the term `tokenizer`? if we would prefer to have a single argument in that regard.

tuning/data/data_preprocessing_utils.py
```diff
@@ -49,12 +55,27 @@ def get_data_collator(
             str representing the human response in a chat template
         is_padding_free: bool
             if padding free plugin is used or not
+        text_field_name: str
```
dushyantbehl, line 58: can we call this `dataset_text_field` if the name is not related to some terminology of vision language models. 
We have been using `dataset_text_field` inside our code at many places. 

similarly maybe update the name for `image_field_name`

tuning/data/data_preprocessing_utils.py
```diff
@@ -96,3 +117,48 @@ def get_data_collator(
         raise ValueError(
             "Could not pick a data collator. Please refer to supported data formats"
         )
+
+class VisionDataCollator:
```
dushyantbehl, line 121: could you move the collator to a separate file for keeping functionality separate?

anhuong: THanks fo the feedback @dushyantbehl all of the changes you suggest look good, I'm really interseted in the questions I listed above and tagged you on if you can take a look

tuning/data/data_preprocessing_utils.py
dushyantbehl: @anhuong few things which you are doing in this data collator can be done as `data handlers` for example, 
1. applying tokenizer (or processor) chat template to the dataset can be handled via a data handler (we already have a [data handler](https://github.com/foundation-model-stack/fms-hf-tuning/blob/496daf98eb5cf88d695f29683e5e0f2485b3545c/tuning/data/data_handlers.py#L179) which applies tokenizer chat template and this can be extended to apply processor chat template and based on context you can pass tokenizer or processor. 
1. Tokenizing text and applying processor on images - Again this can be another data handler which performs these actions and passes a tokenized dataset with both text and images processed to the collator. 

This way we keep the data manipulation open and outside of collator which can later be customised as per user preference. 
This will also solve your problem of handling many types of datasets as we can potentially have different handlers for different datasets. That being said the datasets and usecases which you shared above is something I feel could be handled using `jinja template rendering`.

We have just added a [data handler](https://github.com/foundation-model-stack/fms-hf-tuning/blob/496daf98eb5cf88d695f29683e5e0f2485b3545c/tuning/data/data_handlers.py#L146) which can apply a `jinja` template on the dataset before processing and this can be used to handle all the use cases you showed above. 

In my opinion our first priority should be to identify most used data formats for vision model training and then have support for those and rest can be modelled using jinja template rendering where a user who brings in a dataset like 
https://huggingface.co/datasets/philschmid/amazon-product-descriptions-vlm are also suppose to bring a template on how its to be rendered. 

Finally I would be interested to know more about the masking being done by the data collator and if possibly some data collator already exists in HF which we can utilise for the same purpose so as to not reimplement things on our end. 

With these changes the code can flow through `process_dataset_configs` and `process_data_args` in same way and we can think of providing a few sample data configs for people to perform vision model training.

tuning/data/setup_dataprocessor.py
Abhishek-TAMU: @anhuong Yea for multi-modal case we can still use `process_data_configs` and we could make some changes in current data preprocessor code. But I guess the first step of change could to be adding `processor` as attribute of Class `DataPreProcessor` [here](https://github.com/foundation-model-stack/fms-hf-tuning/blob/f88f031a09604cc86c529b468792f12699f72a2a/tuning/data/data_processors.py#L35) so that we can use it as `self.processor` in `process_dataset_configs` instead of passing `processor` to function `process_dataset_configs` and other functions.

tuning/data/data_preprocessing_utils.py
anhuong: Appreciate the thoughts @dushyantbehl , I like what you said about the data handlers to manipulate the text portion of the datasets and then the text and image can be passed to the collator to combine them, tokenize, and mask as needed. 

> identify most used data formats for vision model training and then have support for those and rest can be modelled using jinja template rendering

This sounds good, agreed, from my search 1) single text field and single image field with conversational data 2) question and answer text fields and single image field 3) many text fields and single image field — where the image is a list of 1+ images. In which for tuning we could support the first two cases and the last one the user would have to provide a template or preprocess into one of the designated formats. 

There is no data collator that currently exists for this as far as a I know, all of the examples I looked at used cusom collators.

anhuong: Some more details on data formats for vision modes from Alex (many thanks!):
Data formats:
- 1+ images & 1 string that already has image placeholders
   - chats have explicitly marked where the images are with the model's specific image tokens
   - something like:
   ```
   {
       "messages": [
           {"role": "user", "content": "<image> what color is the background in this image?"}
       ]
   }
   ```
   - For granite, the tokenize chat template, which handles string data and mostly does the role formatting: https://huggingface.co/ibm-granite/granite-vision-3.1-2b-preview/blob/main/tokenizer_config.json#L196
- OpenAI's conversational data format
   - chat has list data for the content, and it needs to take care of placing the model's image tags
   - something like:
   ```
   {
       "messages": [
           {"role": "user", "content": [
               {"type": "image"},
               {"type": "text", "text": "what color is the background in this image?"}
           ]
       ]
   }
   ```
   - For granite, the chat_template.json handles, and basically will look for images and place the model specific placeholder at the front of the text, which is used by the Processor: https://huggingface.co/ibm-granite/granite-vision-3.1-2b-preview/blob/main/chat_template.json

I would recommend starting with the second one, because that follows OpenAI's standard spec, and as long as the chat template makes sense, it should be able to handle all 3 of those cases; this is also consistent with vLLMs server inference API and transformers, so it is also used as the example in TRL docs ([here](https://huggingface.co/docs/trl/en/sft_trainer#preparing-the-data)) - most multimodal models should have a chat template that can handle this, but if they can't, I think it's fair to ask them to provide one since this is standard in a lot of places

Abhishek-TAMU: > applying tokenizer (or processor) chat template to the dataset can be handled via a data handler (we already have a [data handler](https://github.com/foundation-model-stack/fms-hf-tuning/blob/496daf98eb5cf88d695f29683e5e0f2485b3545c/tuning/data/data_handlers.py#L179) which applies tokenizer chat template and this can be extended to apply processor chat template and based on context you can pass tokenizer or processor.
Tokenizing text and applying processor on images - Again this can be another data handler which performs these actions and passes a tokenized dataset with both text and images processed to the collator.

@dushyantbehl I want to comment on the 2nd handler that you mentioned we can go for:

If we are not using `batched=True` in `dataset.map(handlers, kwargs)` then any handler that tokenizes or apply processor to the dataset would tokenize it individually and hence `padding_tokens` won't be added, as `.map()` doesn't know the largest sequence length in dataset so that it might not apply `padding_token_id` accordingly to every data element `input_ids`.

Abhishek-TAMU: I have added `batched=True` for Data handler which uses processor in this [PR](https://github.com/anhuong/fms-hf-tuning/pull/4). Later we can generalize/optimize this for Data handlers wherever this is necessary.

tuning/data/data_preprocessing_utils.py
anhuong: Would having `dataset_text_field` used for different purposes depending on the model type cause confusion? I think as long as we document it correctly then it should be fine since for text datasets using `dataset_text_field` is for the field with single formatted sequence whereas for text+image dataset it will be for the text field vs image field that seems good

tuning/data/setup_dataprocessor.py
```diff
@@ -283,11 +338,18 @@ def _process_raw_data_args(
...
 
     eval_dataset = None
     if is_eval_dataset_present:
-        eval_dataset = data_processor.process_dataset_configs([eval_dataset_config])
+        if processor:
+            eval_dataset = load_dataset(data_args.training_data_path)
+        else:
+            eval_dataset = data_processor.process_dataset_configs([eval_dataset_config])
```
Abhishek-TAMU, line 352: @anhuong We can keep the original code and remove `if/else condition` here:
`eval_dataset = data_processor.process_dataset_configs([eval_dataset_config])`

Abhishek-TAMU:
@dushyantbehl PR for fix : https://github.com/anhuong/fms-hf-tuning/pull/5
More info [here](https://github.ibm.com/ai-foundation/watson-fm-stack-tracker/issues/1611#issuecomment-105540088)

tuning/data/data_preprocessing_utils.py
anhuong: @dushyantbehl I am not adding this change currently since although `AutoProcessor.from_pretrained()` can load both tokenizers and processors, it's hard to distinguish between which one is loaded and then if the processor with tokenizer is loaded then all times where tokenizer is used needs to be replaced with `processor.tokenizer`. Thus since tokenizer and processor are distinguished apart from each other I am keeping the tokenizer term for now

anhuong:
@dushyantbehl please give this another review

With the changes we have been doing additional testing and have hit into some snag:
- Llama transformers 4.49 
   - Running with llama model runs into error `RuntimeError: shape '[-1, 128256]' is invalid for input of size 635133616` which may be related to this issue: https://github.com/huggingface/transformers/issues/36040 since we are also running with FSDP and gradient_checkpointing. Although that issue notes transofmrers version 4.48 whereas we are able to successfully run tuning with llama vision model on transformers 4.48.
- Error when using larger datasets of 500+ samples 
   - `pyarrow.lib.ArrowInvalid: offset overflow while concatenating arrays, consider casting input from list<item: list<item: list<item: list<item: float>>>> to list<item: list<item: list<item: large_list<item: float>>>> first.`  --> we think that the streaming PR will address this issue
- Able to LoRA tune granite, llava, and llama models, recently ran into issues with full fine tuning which previously succeeded. May be due to configuration changes or changes in the library, investigating....

tuning/data/data_handlers.py
```diff
@@ -247,16 +252,80 @@ def apply_tokenizer_chat_template(
...
         )
+    if chat_template_key and chat_template_key in element:
+        element = element[chat_template_key]
+
     return {
         f"{dataset_text_field}": tokenizer.apply_chat_template(element, tokenize=False)
     }
 
 
+def apply_processor_multimodal_data(
```
dushyantbehl, line 272: nit: apply_multimodal_data_processor

tuning/data/data_handlers.py
```diff
@@ -247,16 +252,80 @@ def apply_tokenizer_chat_template(
         Formatted HF Dataset element by formatting dataset with tokenizer's chat template
         Saves the result to dataset_text_field argument.
     """
+    chat_template_key = kwargs.get("chat_template_key", None)
```
dushyantbehl, line 255: could we rename this to something else? possibly `chat_data_key` 

also if this is an important field I'd suggest making this a natural argument instead of hiding in kwargs.

tuning/data/data_handlers.py
```diff
@@ -247,16 +252,80 @@ def apply_tokenizer_chat_template(
...
+    processor = kwargs.get("processor", None)
```
dushyantbehl, line 256: Also can we not replace `tokenizer` with `processor` that way if its multi modal processor we use processor else we use tokenizer. 

in the end the call is same I guess which is `processor.apply_chat_template`?

tuning/data/data_handlers.py
```diff
@@ -247,16 +252,80 @@ def apply_tokenizer_chat_template(
...
+    if text is None or image is None:
+        raise ValueError("Missing text or image data in element.")
+
+    # Handler is used with batch=True where image is `List[List[PIL.Image], List[PIL.Image]]`
+    # We need to convert it to `List[PIL.Image]` for LlavaProcessor
+    if isinstance(processor, LlavaProcessor):
+        if isinstance(image, list) and image and isinstance(image[0], list):
+            image = [img[0] for img in image]
+        else:
+            raise ValueError(
```
dushyantbehl, line 306: Are we enforcing that this handler should be used always in batched mode?

tuning/data/data_preprocessing_utils.py
```diff
@@ -30,6 +39,7 @@ def get_data_collator(
     max_seq_length: int,
     instruction_template: Optional[str],
     is_padding_free: bool = False,
+    processor: Optional[Union[AutoProcessor, LlavaProcessor]] = None,
```
dushyantbehl, line 42: do we need to pass processor and tokenizer spearately? I thought the term processor overloads tokenizer and image processors in HF.

tuning/data/setup_dataprocessor.py
```diff
@@ -255,7 +302,15 @@ def _process_raw_data_args(
 
     handlers = None
     dataset_text_field = None
-    if is_traindata_tokenized:
+
+    # ToDo: Think about a better way to handle this if condition
```
dushyantbehl, line 306: nit: TODO

tuning/data/setup_dataprocessor.py
```diff
@@ -352,13 +411,24 @@ def process_dataargs(
...
+    if processor and not (
+        data_args.dataset_text_field or data_args.dataset_image_field
+    ):
+        logger.error(
+            "When running a vision model you must provide the dataset_text_field and \
+            dataset_image_field for the columns in the dataset. Values should be from \
+            column names: %s",
+            train_dataset.column_names,
+        )
+
```
dushyantbehl, line 431: Why are we not raising a `ValueError` here?

tuning/sft_trainer.py
```diff
@@ -213,28 +215,61 @@ def train(
...
-            model_args.tokenizer_name_or_path
-            if model_args.tokenizer_name_or_path
-            else model_args.model_name_or_path
-        ),
-        cache_dir=train_args.cache_dir,
-        use_fast=True,
-        legacy=True,
-    )
+        processor = AutoProcessor.from_pretrained(model_args.model_name_or_path)
+        tokenizer = processor.tokenizer
```
dushyantbehl, line 239: we can just use the term processor or whatever is standard.

tuning/sft_trainer.py
```diff
@@ -213,28 +215,61 @@ def train(
         fusedops_kernels_config,
     ).get_framework()
 
-    model_loader = AutoModelForCausalLM.from_pretrained
-    if framework is not None and framework.requires_custom_loading:
-        model_loader = framework.model_loader  # drop-in new loader
+    # processor = None
```
dushyantbehl, line 218: Can we remove the deadcode?

tuning/utils/collators.py
```diff
@@ -0,0 +1,56 @@
...
+
+        Returns:
+            batch (Dict[str, torch.Tensor]): A dict where each key corresponds to a batched tensor
+            created from the respective feature key in `features`. A new key `"labels"` is added,
+            which is a clone of `"input_ids"` with padding and image tokens masked as `-100`.
+        """
+
+        # The labels are the input_ids, and we mask the padding tokens in the loss computation
+        # TOOD: should we be ensuring EOS tokens is set?
+        # As chat template is applied so it should be set.
```
dushyantbehl, line 39: how do we ensure if chat template has EOS already?

pyproject.toml
```diff
@@ -29,7 +29,7 @@ classifiers=[
 dependencies = [
 "numpy>=1.26.4,<2.0",
 "accelerate>=0.20.3,!=0.34,<1.1",
-"transformers>=4.46,<4.48.2",
+"transformers>=4.49,<5.0",
```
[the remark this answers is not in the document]
anhuong: Good idea, the errors do seem not related to our changes and I did correct one unit test due to the upgrade already

tests/test_sft_trainer.py
```diff
@@ -318,14 +318,6 @@ def _get_training_logs_by_epoch(dir_path: str, epoch: int = None):
     return data_list
 
 
-def test_run_train_requires_output_dir():
```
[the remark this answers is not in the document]
anhuong: Correct this test was failing in newer transformers version since output_dir is no longer required and thus I removed teh test case since its no longer relevant

pyproject.toml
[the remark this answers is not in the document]
anhuong: Seeing how unit tests run here - https://github.com/foundation-model-stack/fms-hf-tuning/pull/485

tuning/data/data_handlers.py
anhuong: This would work well to just run processor.apply_chat_template but did you see my [previous comment](https://github.com/foundation-model-stack/fms-hf-tuning/pull/451#discussion_r1976966107) of how best to load the tokenizer since if the processor is loaded you need to pass the tokenizer via `processor.tokenizer` in all of the methods that need the tokenizer. They aren't interchangeable although some methods like `apply_chat_template()` are available on both the processor and tokenizer

Abhishek-TAMU: @dushyantbehl If you are specifically referring to replacing `tokenizer` with `processor` in [apply_tokenizer_chat_template](https://github.com/anhuong/fms-hf-tuning/blob/50cb0bd896059d68e011929ffcbffba0d95aece9/tuning/data/data_handlers.py#L239) then changes would be required [here too](https://github.com/anhuong/fms-hf-tuning/blob/50cb0bd896059d68e011929ffcbffba0d95aece9/tuning/data/data_processors.py#L300), and again for this change it would require changes in `tuning/data/setup_dataprocessor.py` on multiple places on how we receive `tokenizer` and `processor` both which comes back to the point by Anh [above](https://github.com/foundation-model-stack/fms-hf-tuning/pull/451#discussion_r1979578564).

tuning/utils/collators.py
Abhishek-TAMU: I don't think we should check this as `chat_template` of Llama and Granite vision model adds EOS when applying chat template.

tuning/data/data_preprocessing_utils.py
Abhishek-TAMU: Probably we can conclude on [this comment first](https://github.com/foundation-model-stack/fms-hf-tuning/pull/451/files#r1979405768) on the usage of `tokenizer` and `processor` separately and then could decide this.

tuning/sft_trainer.py
Abhishek-TAMU: You mean not to use `tokenizer = processor.tokenizer` ?

tuning/data/data_handlers.py
Abhishek-TAMU: Mentioned [here](https://github.com/foundation-model-stack/fms-hf-tuning/pull/451#discussion_r1957010946) on why I am using this handler in batched mode.

1- If we are not using `batched=True` in `dataset.map(handlers, kwargs)` then any handler that tokenizes or apply processor to the dataset would tokenize it individually and hence `padding_tokens` won't be added, as `.map()` doesn't know the largest sequence length in dataset so it doesn't apply padding_token_id accordingly to any of data element `input_ids`.

2- In this handler when applying processor (`element = processor(text=text, images=image, **processor_kwargs)`), element type return is `torch`. 
But when `.map()` is applied to this handler we get back `List` instead of `torch` and when `batch=False` then it adds one more dimension to the `List` which makes it incorrect. This is avoided by `batch=True`

tuning/data/data_handlers.py
```diff
@@ -243,20 +249,84 @@ def apply_tokenizer_chat_template(
...
+                "Expected image to be a non-empty list of lists \
+                with a single image for LlavaProcessor."
+            ) from e
```
Abhishek-TAMU, line 310: @anhuong I missed this in my PR :) Thank you.
```suggestion
            )
```

tuning/data/data_handlers.py
```diff
@@ -243,20 +249,84 @@ def apply_tokenizer_chat_template(
...
+        if isinstance(image, list) and image and isinstance(image[0], list):
+            image = [
+                img[0].convert("RGB") if img[0].mode != "RGB" else img[0]
+                for img in image
+            ]
+        else:
+            raise ValueError(
+                "Expected image to be a non-empty list of lists \
+                with a single image for LlavaNextProcessor."
+            ) from e
```
Abhishek-TAMU, line 323:
```suggestion
            )
```

tuning/utils/collators.py
anhuong: i don't think we have to ensure, i removed this comment, i think we should trust the chat template that its setup properly

tuning/sft_trainer.py
```diff
@@ -213,28 +215,54 @@ def train(
...
-    )
+    processor = None
+    try:
+        # try to load vision model
+        model_loader = AutoModelForVision2Seq.from_pretrained
+        model = model_loader(
+            model_args.model_name_or_path,
+            cache_dir=train_args.cache_dir,
+            torch_dtype=get_torch_dtype(model_args.torch_dtype),
+            attn_implementation="flash_attention_2"
```
Ssukriti, line 228: is attn_implementation="flash_attention_2" working for you @anhuong with the llama vision model?

when I try the standalone SFT script you gave me , it is says not supported for mllama architecture, also in https://github.com/huggingface/transformers/issues/35634

anhuong: It doesn't work for me either for llama thus I set `"use_flash_attn": false,` for llama but flash attention does work for granite and llava models. Writing up docs explaining usage

anhuong:
With the changes in this PR, we are able to full fine tune and LoRA tune Granite 3.2, Llama 3.2, and Llava-Next vision models on image+text dataset where the dataset has a single text field in conversational data format and the a single image field. More details on what is supported and how to run the models are in the README.
We have added a hack to resize the vocab size of Mllama models if embeddings are resized to work around a bug in transformers. 

Some exmaple configurations that have run through successfully,

Granite full fine tuning:
```json
{
  "model_name_or_path": "ibm-granite/granite-vision-3.1-2b-preview", 
  "training_data_path": "HuggingFaceH4/llava-instruct-mix-vsft",
  "output_dir": "/app/test_granite",
  "num_train_epochs": 1.0,
  "per_device_train_batch_size": 8,
  "gradient_accumulation_steps": 2,
  "learning_rate": 1e-4,
  "response_template": "\n### Response:",
  "bf16": true,
  "torch_dtype": "bfloat16",
  "use_flash_attn": true,
  "remove_unused_columns": false,
  "dataset_kwargs": {"skip_prepare_dataset": true},
  "gradient_checkpointing": true,
  "dataset_text_field": "messages",
  "dataset_image_field": "images",
  "gradient_checkpointing_kwargs": {"use_reentrant": false},
  "accelerate_launch_args": { "fsdp_transformer_layer_cls_to_wrap": "GraniteDecoderLayer"}
}
```

Llama LoRA tuning:
```json
{
  "model_name_or_path": "meta-llama/Llama-3.2-11B-Vision-Instruct", 
  "training_data_path": "HuggingFaceH4/llava-instruct-mix-vsft",
  "output_dir": "/app/test_image_output",
  "num_train_epochs": 1.0,
  "per_device_train_batch_size": 8,
  "gradient_accumulation_steps": 2,
  "learning_rate": 1e-4,
  "response_template": "\n### Response:",
  "bf16": true,
  "torch_dtype": "bfloat16",
  "use_flash_attn": true,
  "remove_unused_columns": false,
  "dataset_kwargs": {"skip_prepare_dataset": true},
  "gradient_checkpointing": true,
  "dataset_text_field": "messages",
  "dataset_image_field": "images",
  "use_flash_attn": false
}
```

Llava (also works with llava-hf/llava-v1.6-34b-hf but ran into OOM)
```json
{
  "model_name_or_path": "llava-hf/llava-v1.6-mistral-7b-hf", 
  "training_data_path": "HuggingFaceH4/llava-instruct-mix-vsft",
  "output_dir": "/app/test_llava_mistral_hack_fix",
  "num_train_epochs": 1.0,
  "per_device_train_batch_size": 8,
  "gradient_accumulation_steps": 2,
  "learning_rate": 1e-4,
  "response_template": "\n### Response:",
  "bf16": true,
  "torch_dtype": "bfloat16",
  "use_flash_attn": true,
  "remove_unused_columns": false,
  "dataset_kwargs": {"skip_prepare_dataset": true},
  "gradient_checkpointing": true,
  "dataset_text_field": "messages",
  "dataset_image_field": "images",
  "gradient_checkpointing_kwargs": {"use_reentrant": false},
  "accelerate_launch_args": { "fsdp_transformer_layer_cls_to_wrap": "MistralDecoderLayer"}
}
```

Abhishek-TAMU:
[Sharing PR](https://github.com/anhuong/fms-hf-tuning/pull/6) for tuning with Entire Vision Dataset.

README.md
```diff
@@ -846,6 +847,101 @@ The `fms_acceleration.cli` can do more to search for all available configs, plug
...
+We recommend running with `gradient_checkpointing` set as this will greatly reduce the memory needed to load and run the model. When running with gradient checkpointing, for the Llava and Granite models, you will need to also set `"gradient_checkpointing_kwargs": {"use_reentrant": false}` as seen in the example above in order to not use the activation checkpoint variant that requires reentrant autograd. Without this set tuning will lead to error:
+
+```
+RuntimeError: mat2 must be a matrix, got 1-D tensor
+RuntimeError: Expected weight to be of same shape as normalized_shape, but got weight of shape [0] and normalized_shape = [1152]
+```
+
+We also recommend running with `"remove_unused_columns": false` and `"dataset_kwargs": {"skip_prepare_dataset": true}` as you see above to avoid the default processing of the dataset and to ensure that the data is not processed as text-only.
+
+When running LoRA tuning, you must specify the `target_modules` since there are no defaults set for the vision models.
```
Abhishek-TAMU, line 943: Suggestion: See rendered text at last;

``````
### Tuning Considerations for vision models

Flash Attention 2.0 is not supported by `MllamaForConditionalGeneration` models, thus when running tuning with the `Llama 3.2 Vision Models` set:
```
use_flash_attn=False
```
#### Multi-GPU Tuning with FSDP:
When running `multi-GPU` tuning with `FSDP`, you need to wrap specific transformer layers. Use the following setting in FSDP config based on your model:

`Granite 3.2 Vision Models:`
```
"accelerate_launch_args": { "fsdp_transformer_layer_cls_to_wrap": "GraniteDecoderLayer" }
```

`Llava-Next and Llava-1.5 Models:`
```
"accelerate_launch_args": { "fsdp_transformer_layer_cls_to_wrap": "LlamaDecoderLayer" }
```

`Llava-1.6-Mistral Model:`
```
"accelerate_launch_args": { "fsdp_transformer_layer_cls_to_wrap": "MistralDecoderLayer" }
```

`Llama 3.2 Vision Models: ` No additional configuration is required.

#### Gradient Checkpointing:

- We recommend running with argument `gradient_checkpointing = True` as enabling this will greatly reduce the memory needed to load and run the model.

- When running with gradient checkpointing for the `Llava` and `Granite` vision models, you will need to also set `gradient_checkpointing_kwargs` to not use the activation checkpoint variant that requires reentrant autograd.  
```
"gradient_checkpointing_kwargs": {"use_reentrant": false}
```
Without setting this, tuning will lead to error:

```
RuntimeError: mat2 must be a matrix, got 1-D tensor
RuntimeError: Expected weight to be of same shape as normalized_shape, but got weight of shape [0] and normalized_shape = [1152]
```

#### Other arguments:
- To prevent default text-only processing and ensure proper handling of multimodal data, we recommend setting:
```
"remove_unused_columns": false
"dataset_kwargs": {"skip_prepare_dataset": true}
```
- When performing LoRA tuning on vision models, you must specify the `target_modules` explicitly, as no defaults are provided.
``````


**RENDERED TEXT:**

### Tuning Considerations for vision models

Flash Attention 2.0 is not supported by `MllamaForConditionalGeneration` models, thus when running tuning with the `Llama 3.2 Vision Models` set:
```
use_flash_attn=False
```
#### Multi-GPU Tuning with FSDP:
When running `multi-GPU` tuning with `FSDP`, you need to wrap specific transformer layers. Use the following setting in FSDP config based on your model:

`Granite 3.2 Vision Models:`
```
"accelerate_launch_args": { "fsdp_transformer_layer_cls_to_wrap": "GraniteDecoderLayer" }
```

`Llava-Next and Llava-1.5 Models:`
```
"accelerate_launch_args": { "fsdp_transformer_layer_cls_to_wrap": "LlamaDecoderLayer" }
```

`Llava-1.6-Mistral Model:`
```
"accelerate_launch_args": { "fsdp_transformer_layer_cls_to_wrap": "MistralDecoderLayer" }
```

`Llama 3.2 Vision Models: ` No additional configuration is required.

#### Gradient Checkpointing:

- We recommend running with argument `gradient_checkpointing = True` as enabling this will greatly reduce the memory needed to load and run the model.

- When running with gradient checkpointing for the `Llava` and `Granite` vision models, you will need to also set `gradient_checkpointing_kwargs` to not use the activation checkpoint variant that requires reentrant autograd.  
```
"gradient_checkpointing_kwargs": {"use_reentrant": false}
```
Without setting this, tuning will lead to error:

```
RuntimeError: mat2 must be a matrix, got 1-D tensor
RuntimeError: Expected weight to be of same shape as normalized_shape, but got weight of shape [0] and normalized_shape = [1152]
```

#### Other arguments:
- To prevent default text-only processing and ensure proper handling of multimodal data, we recommend setting:
```
"remove_unused_columns": false
"dataset_kwargs": {"skip_prepare_dataset": true}
```
- When performing LoRA tuning on vision models, you must specify the `target_modules` explicitly, as no defaults are provided.

**Review by Abhishek-TAMU — COMMENTED** [review]
Thanks @anhuong for adding this documentation content. I just have few suggestions which could improve the presentation. 

@dushyantbehl Feel free to let us know your thoughts on documentation.

File: README.md [hunk 26]
```diff
@@ -846,6 +847,101 @@ The `fms_acceleration.cli` can do more to search for all available configs, plug
 
 We also have support for extended pre training where users might wanna pretrain a model with large number of samples. Please refer our separate doc on [EPT Use Cases](./docs/ept.md)
 
+## Vision Language Models
+
+We also support full fine tuning and LoRA tuning for vision language models - Granite 3.2 Vision, Llama 3.2 Vision, and LLaVa-Next.
```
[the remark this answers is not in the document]
anhuong: I don't think we need to specify the sizes because we should support all of the different size models. And then I didn't explicitly say the mistral llava model since i was able to run tuning on the `llava-v1.6-34b-hf` model as well but hit into OOM issues since that model is much larger 34b instead of 7b. But I think ideally we would target supporting `llava-v1.6-34b-hf` over the mistral specific version

README.md
```diff
@@ -846,6 +847,101 @@ The `fms_acceleration.cli` can do more to search for all available configs, plug
...
+  "use_flash_attn": true,
+  "remove_unused_columns": false,
+  "dataset_kwargs": {"skip_prepare_dataset": true},
+  "gradient_checkpointing": true,
+  "gradient_checkpointing_kwargs": {"use_reentrant": false},
+  "accelerate_launch_args": { "fsdp_transformer_layer_cls_to_wrap": "GraniteDecoderLayer"}
+}
+```
+
+This can similarly be run as:
```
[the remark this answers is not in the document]
anhuong: both ways are calling the trainer via command-line so adjusting wording here

README.md
anhuong: very nice improvements, tweaked it a bit

README.md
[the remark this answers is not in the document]
anhuong: I agree this is a great place to add additional templating recommendations

README.md
```diff
@@ -846,6 +847,160 @@ The `fms_acceleration.cli` can do more to search for all available configs, plug
...
+  "dataset_kwargs": {"skip_prepare_dataset": true},
+  "gradient_checkpointing": true,
+  "gradient_checkpointing_kwargs": {"use_reentrant": false},
+  "accelerate_launch_args": { "fsdp_transformer_layer_cls_to_wrap": "GraniteDecoderLayer"}
+}
+```
+
+### Running the Trainer
+
+You can also run training by calling the trainer directly using the command line. For example:
```
Ssukriti, line 932:
```suggestion
You can also run training by calling the trainer directly using the command line. You can use `python` for single GPU or `accelerate launch` command for multi GPU. 
For example:
```

tuning/data/data_handlers.py
```diff
@@ -268,20 +274,74 @@ def apply_tokenizer_chat_template(
...
+    # We need to convert it to `List[PIL.Image]` for LlavaProcessor
+    if isinstance(processor, LlavaProcessor):
+        if isinstance(image, list) and image and isinstance(image[0], list):
+            image = [img[0] for img in image]
+
+    # If LlavaNextProcessor then convert mode of image to RGB. Process of Granite-3.2-Vision Model
+    elif isinstance(processor, LlavaNextProcessor):
+        if isinstance(image, list) and image and isinstance(image[0], list):
+            image = [
+                img[0].convert("RGB") if img[0].mode != "RGB" else img[0]
```
alex-jw-brooks, line 336: I think this can probably be collapsed into the block above it so that the LLava/LlavaNext processor are handling things the same way, i.e.,like this

```python
  if isinstance(processor, (LlavaProcessor, LlavaNextProcessor)):
         if isinstance(image, list) and image and isinstance(image[0], list):
             image = [
                 img[0].convert("RGB") if img[0].mode != "RGB" else img[0]
                 for img in image
             ]
```

in general, not having RGB images will cause problems for most models 🙂

Abhishek-TAMU: Thanks @alex-jw-brooks for the input :) 

> in general, not having RGB images will cause problems for most models 🙂

Do you mean/recommend that we convert all images to RGB for every model and not just `LlavaProcessor, LlavaNextProcessor` ?

alex-jw-brooks: Most models have utils to support doing it in their ImageProcessors (which are contained in the Processor, along with the tokenizer), so you probably don't actually need to do this, but it also probably doesn't really hurt.

Most of the time, the image processors classes have a kwarg called `do_convert_rgb`, which will handle the RGB conversion by default, but sometimes they might vary a little bit, particularly in how they handle alpha channels. As an example, the kwarg in the llava/llava next image processors map to this [common implementation](https://github.com/huggingface/transformers/blob/main/src/transformers/image_transforms.py#L793), while the same kwarg in mllama maps to [this implementation](https://github.com/huggingface/transformers/blob/179d02ffb8f3e0f0f3330dd2762286e83cbaa65b/src/transformers/models/mllama/image_processing_mllama.py#L496).

So I think I'd suggest checking what the image processor in transformers is doing for the new architecture being supported, but most of the time it's probably fine :)

Abhishek-TAMU: Thanks for the info @alex-jw-brooks! This is helpful.

dushyantbehl (changes requested):
Thanks @Abhishek-TAMU and @anhuong 

Just left some comments and requesting few changes and movement of code blocks for integration with data preprocessing more neatly.

tuning/data/data_handlers.py
```diff
@@ -260,6 +265,7 @@ def apply_tokenizer_chat_template(
     element: Dict[str, str],
     tokenizer: AutoTokenizer,
     dataset_text_field: str,
+    chat_data_key: str = None,
```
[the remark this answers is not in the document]
Abhishek-TAMU: Yes pull changes from main would fix it.

tuning/data/setup_dataprocessor.py
Abhishek-TAMU: As this check is for checking if `dataset_text_field` and `dataset_image_field` exists with the `processor`, it won't go inside data handler `_get_vision_dataset_handlers`  itself if error is not raise before. Hence moved its position up.

README.md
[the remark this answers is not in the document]
Abhishek-TAMU: Will let @anhuong  populate the table.

README.md
```diff
@@ -846,6 +847,160 @@ The `fms_acceleration.cli` can do more to search for all available configs, plug
 
 We also have support for extended pre training where users might wanna pretrain a model with large number of samples. Please refer our separate doc on [EPT Use Cases](./docs/ept.md)
 
+## Vision Language Models
+
+We also support full fine tuning and LoRA tuning for vision language models - Granite 3.2 Vision, Llama 3.2 Vision, and LLaVa-Next. 
+
+We support tuning an `image+text` dataset that includes:
```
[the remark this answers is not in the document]
Abhishek-TAMU: Sure, renamed section to `Tuning Vision Language Models`. [PR commit.](https://github.com/anhuong/fms-hf-tuning/pull/6/commits/04a7b32d0d3be03a6f8dff66ba90e21cbd0b042e#diff-e6943cd619cc097fa859831602e0c7716f5b55f702c002a13963dbf62eb7069f)

README.md
```diff
@@ -846,6 +847,160 @@ The `fms_acceleration.cli` can do more to search for all available configs, plug
...
+        "content": [
+            {"type": "text", "text": "He is the 44th President of the United States."}
+        ]
+    }
+]
+```
+
+### Processing of dataset
+
+First, each dataset sample is processed by applying the chat template to the raw text, which formats the conversation as required. Then, the model’s `processor` takes the formatted text and the corresponding image(s) and converts them into the final input representation (e.g., input_ids, attention masks, etc.) that the model uses for training.
```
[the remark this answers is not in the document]
Abhishek-TAMU: Added these HF links. 
https://huggingface.co/docs/transformers/main/en/chat_templating
https://huggingface.co/docs/transformers/main/en/processors

[PR commit.](https://github.com/anhuong/fms-hf-tuning/pull/6/commits/04a7b32d0d3be03a6f8dff66ba90e21cbd0b042e#diff-e6943cd619cc097fa859831602e0c7716f5b55f702c002a13963dbf62eb7069f)

README.md
```diff
@@ -846,6 +847,160 @@ The `fms_acceleration.cli` can do more to search for all available configs, plug
...
+--output_dir $OUTPUT_PATH  \
+--num_train_epochs 5  \
+--per_device_train_batch_size 4  \
+--gradient_accumulation_steps 4  \
+--learning_rate 1e-5  \
+--dataset_text_field "messages" \
+--dataset_image_field "images"
+```
+
+### Tuning Considerations for vision models
```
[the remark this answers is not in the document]
Abhishek-TAMU: I was thinking the same. Have separated a md file. [PR commit.](https://github.com/anhuong/fms-hf-tuning/pull/6/commits/04a7b32d0d3be03a6f8dff66ba90e21cbd0b042e#diff-e6943cd619cc097fa859831602e0c7716f5b55f702c002a13963dbf62eb7069f)

tuning/data/setup_dataprocessor.py
```diff
@@ -282,7 +329,14 @@ def _process_raw_data_args(
 
     handlers = None
     dataset_text_field = None
-    if is_traindata_tokenized:
+
+    # TODO: Better way to handle vision if condition
+    if data_args.dataset_text_field and data_args.dataset_image_field:
```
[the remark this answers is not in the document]
Abhishek-TAMU: [PR commit.](https://github.com/anhuong/fms-hf-tuning/pull/6/commits/04a7b32d0d3be03a6f8dff66ba90e21cbd0b042e#diff-e6943cd619cc097fa859831602e0c7716f5b55f702c002a13963dbf62eb7069f)

tuning/data/setup_dataprocessor.py
```diff
@@ -331,6 +385,7 @@ def process_dataargs(
     train_args: TrainingArguments,
     additional_data_handlers: Dict[str, DataHandler] = None,
     is_padding_free: bool = False,
+    processor=None,
```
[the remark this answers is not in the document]
Abhishek-TAMU: [PR commit.](https://github.com/anhuong/fms-hf-tuning/pull/6/commits/04a7b32d0d3be03a6f8dff66ba90e21cbd0b042e#diff-e6943cd619cc097fa859831602e0c7716f5b55f702c002a13963dbf62eb7069f)

tuning/data/setup_dataprocessor.py
```diff
@@ -238,6 +284,7 @@ def _process_raw_data_args(
     max_seq_length: int,
     additional_data_handlers: Dict[str, DataHandler] = None,
     is_padding_free: bool = False,
+    processor=None,
```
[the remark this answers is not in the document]
Abhishek-TAMU: [PR commit.](https://github.com/anhuong/fms-hf-tuning/pull/6/commits/04a7b32d0d3be03a6f8dff66ba90e21cbd0b042e#diff-e6943cd619cc097fa859831602e0c7716f5b55f702c002a13963dbf62eb7069f).

README.md
```diff
@@ -846,6 +847,160 @@ The `fms_acceleration.cli` can do more to search for all available configs, plug
...
+        ]
+    }
+]
+```
+
+### Processing of dataset
+
+First, each dataset sample is processed by applying the chat template to the raw text, which formats the conversation as required. Then, the model’s `processor` takes the formatted text and the corresponding image(s) and converts them into the final input representation (e.g., input_ids, attention masks, etc.) that the model uses for training.
+
+**Note**: `Granite 3.2` and `Llava-1.5` Vision models expect a single image for each dataset sample. If a list of images is provided, only the first image will be used.
```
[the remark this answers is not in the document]
anhuong: true we can add warnings if the user passes in a list of multiple images, i don't think we should fail on it in case the user wants to reuse the same dataset for diff models

tuning/data/setup_dataprocessor.py
```diff
@@ -394,10 +462,11 @@ def process_dataargs(
         max_seq_length,
         data_args.instruction_template,
         is_padding_free=is_padding_free,
+        processor=processor,
     )
 
     dataset_kwargs = {}
-    if is_tokenized_dataset:
+    if is_tokenized_dataset or processor is not None:
```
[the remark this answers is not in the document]
anhuong: note that currently we are only loading the `processor` / utilizing the processor variable if the model is able to be loaded with `AutoModelForVision2Seq` which is how we are ensuring vision model training. This is somewhat obscure especially if we refactor in the future processor to be used for tokenizers as well. In that same block that loads the vision model, we can create a variable called vision_tuning or something to then pass to data preprocessor and be more clear. But is not something the user sets but still something we infer.

tuning/sft_trainer.py
```diff
@@ -215,28 +217,54 @@ def train(
...
-    model = model_loader(
-        model_args.model_name_or_path,
-        cache_dir=train_args.cache_dir,
-        torch_dtype=get_torch_dtype(model_args.torch_dtype),
-        attn_implementation="flash_attention_2" if model_args.use_flash_attn else None,
-    )
+    processor = None
+    try:
+        # try to load vision model
+        model_loader = AutoModelForVision2Seq.from_pretrained
```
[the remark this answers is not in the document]
anhuong: Qhat do you mean by "this could be moved to one line"? I understand the `model_loader = framework.model_loader` is used for fms-acceleration but that is only in the else case for loading language models and not in the loading vision model section. In order to not have users need to specify if the model they are using is a multimodal vision model, I try loading the model as a vision model first. If this fails as it does for language models, it will load the model as it currently does.

tuning/sft_trainer.py
```diff
@@ -377,7 +406,7 @@ def train(
 
     trainer = SFTTrainer(
         model=model,
-        tokenizer=tokenizer,
+        processing_class=tokenizer,
```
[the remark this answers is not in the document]
anhuong: SFTTrainer renamed `tokenizer` field to `processing_class` with expected type of `PreTrainedTokenizerBase`  and in SFT examples, you still need to pass in the tokenizer into processing_class as seen in docs here https://huggingface.co/docs/trl/en/sft_trainer#training-the-vision-language-model

README.md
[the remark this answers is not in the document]
Abhishek-TAMU: Added warnings. [PR commit.](https://github.com/anhuong/fms-hf-tuning/pull/6/commits/8cfd14063432b539b5dc2e39493d2e4d75ce98e1)

tuning/data/setup_dataprocessor.py
```diff
@@ -218,6 +218,52 @@ def _get_default_dataset_handlers(data_args, tokenizer_kwargs):
...
+        "return_tensors": "pt",
+        "padding": True,
+    }
+    fn_kwargs2 = {
+        "processor": processor,
+        "fields_name": {
+            "dataset_text_field": data_args.dataset_text_field,
+            "dataset_image_field": data_args.dataset_image_field,
+        },
+        "processor_kwargs": processor_kwargs,
```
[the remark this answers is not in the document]
Abhishek-TAMU: Done. [PR commit.](https://github.com/anhuong/fms-hf-tuning/pull/6/commits/8cfd14063432b539b5dc2e39493d2e4d75ce98e1)

tuning/utils/collators.py
```diff
@@ -0,0 +1,55 @@
...
+
+        Returns:
+            batch (Dict[str, torch.Tensor]): A dict where each key corresponds to a batched tensor
+            created from the respective feature key in `features`. A new key `"labels"` is added,
+            which is a clone of `"input_ids"` with padding and image tokens masked as `-100`.
+        """
+
+        # The labels are the input_ids, and we mask the padding tokens in the loss computation
+        # As chat template is applied so it should be set.
+        batch = {}
```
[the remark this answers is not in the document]
Abhishek-TAMU: Added. [PR commit](https://github.com/anhuong/fms-hf-tuning/pull/6/commits/7cc8490f2b0e19ebcc1682782cfc6a472c866493)

tuning/data/data_handlers.py
```diff
@@ -268,20 +274,74 @@ def apply_tokenizer_chat_template(
...
         tokenizer: Tokenizer to be used.
         dataset_text_field: formatted_dataset_field.
+        chat_data_key: dataset field where chat template will be applied.
     Returns:
         Formatted HF Dataset element by formatting dataset with tokenizer's chat template
         Saves the result to dataset_text_field argument.
     """
+    processor = kwargs.get("processor", None)
+    if processor is not None:
+        tokenizer = processor
```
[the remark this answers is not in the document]
Abhishek-TAMU: I see. Before doing this change I believe, we need to finalize on what different variable we will use for already used `processor` in function `get_datapreprocessor` as it is same as image `processor` what I have passed here. 

My thought was to replace `processor` with `preprocessor` in function `get_datapreprocessor` and continue use Image `processor` as `processor` itself.
image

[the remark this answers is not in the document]
Abhishek-TAMU: This is done.

tuning/utils/tokenizer_data_utils.py
```diff
@@ -52,4 +81,10 @@ def tokenizer_and_embedding_resize(
 
         input_embeddings[-num_new_tokens:] = input_embeddings_avg
         output_embeddings[-num_new_tokens:] = output_embeddings_avg
+
+        # resize vocab size when embeddings updated for Mllama models
+        # needed to fix bug while waiting for https://github.com/huggingface/transformers/pull/36591
+        if isinstance(model, MllamaForConditionalGeneration):
+            model.language_model.vocab_size = embedding_size
```
Ssukriti, line 88: @Abhishek-TAMU with latest transformers in main, the loss calculation is moved to `MllamaForConditionalGeneration` , hence vocab size of `MllamaForConditionalGeneration` needs to be modified and language model is not needed.      https://github.com/huggingface/transformers/blob/786d9c5ed920a099573ea7b6dbf265f1aeb32fc0/src/transformers/models/mllama/modeling_mllama.py#L2173

if you want to support both old and new versions, you can set vocab size of both as there is no harm, but for new version `self.config.get_text_config().vocab_size = vocab_size` from https://github.com/huggingface/transformers/blob/main/src/transformers/modeling_utils.py#L2691 is needed as that is what is used in loss function

tuning/utils/tokenizer_data_utils.py
```diff
@@ -38,8 +40,35 @@ def tokenizer_and_embedding_resize(
     num_new_tokens = tokenizer.add_special_tokens(special_tokens_dict)
     embedding_size = int(multiple_of * math.ceil(len(tokenizer) / multiple_of))
     num_new_tokens = num_new_tokens + embedding_size - len(tokenizer)
-    model.resize_token_embeddings(embedding_size)
```
Ssukriti, line 41: this was needed here as this function is used for optimization as well, when there are no new tokens , but to resize embeddings to multiple of 4 or so which was found to give performance gains. Hence IRL team had added ability to resize even without addition of new tokens

tuning/utils/tokenizer_data_utils.py
```diff
@@ -38,8 +40,35 @@ def tokenizer_and_embedding_resize(
...
+
     if num_new_tokens > 0:
+        if isinstance(model, MllamaForConditionalGeneration):
```
Ssukriti, line 45: related to comment above , if general "resize to multiple of " is done without addition of new tokens, it should be handled separately for this class too . Input embedding size and output embedding size will be different. Code may be simplified to handle both cases

tuning/utils/tokenizer_data_utils.py
Ssukriti: there should be a way to consolidate this if condition and one line with above "if" condition and put them in one block

tuning/utils/tokenizer_data_utils.py
```diff
@@ -38,8 +40,35 @@ def tokenizer_and_embedding_resize(
...
+            # Get new input embedding size
+            current_input_embeddings = model.get_input_embeddings()
+            current_output_embeddings = model.get_output_embeddings()
+            input_embedding_size = current_input_embeddings.weight.shape[0] + (
+                embedding_size - current_output_embeddings.weight.shape[0]
+            )
+
+            # Save current input embedding
```
Ssukriti, line 53: unit test for this class would be great for this function

tuning/utils/tokenizer_data_utils.py
Abhishek-TAMU: Sure, let me confirm with Mehant.

tuning/utils/tokenizer_data_utils.py
Abhishek-TAMU: That's a good point. Once I get confirmation from Mehant on the above comment on moving `model.resize_token_embeddings(embedding_size)`, I will consider this point too. Thanks.

tuning/utils/tokenizer_data_utils.py
Abhishek-TAMU: Yea, I was planning to remove this if condition when your PR would release, but now when I think of your statement of supporting both old and new version, I guess I can make [this change](https://github.com/foundation-model-stack/fms-hf-tuning/pull/451/commits/4f91fb3bc5d8e9cb2bc152ee5977e63af76aaa2b). Thanks!

tuning/utils/tokenizer_data_utils.py
Abhishek-TAMU: I can definitely add a test case, something like below. But I was wondering do we use bigger models for test as Llama vision model is a 11b model (`meta-llama/Llama-3.2-11B-Vision-Instruct`) and downloading and loading it every time in unit test run would be a lot of time just for this 1 test case. Any comments ?

```
def test_resize_llama_vision_model():
    model = AutoModelForVision2Seq.from_pretrained(LLAMA_VISION_MODEL_NAME)
    processor = AutoProcessor.from_pretrained(LLAMA_VISION_MODEL_NAME)
    tokenizer = processor.tokenizer

    current_input_embeddings = model.get_input_embeddings()
    current_output_embeddings = model.get_output_embeddings()
    input_tokenizer_len = len(tokenizer.get_vocab())

    resize_result = tokenizer_and_embedding_resize(
        special_tokens_dict={"unk_token": "<unk>"}, tokenizer=tokenizer, model=model, multiple_of=1
    )

    resized_input_embeddings = model.get_input_embeddings()
    resized_output_embeddings = model.get_output_embeddings()
    output_tokenizer_len = len(tokenizer.get_vocab())

    assert input_tokenizer_len == output_tokenizer_len + 1
    assert "<unk>" in tokenizer.get_vocab()
    assert resize_result["num_new_tokens"] == 1

    assert resized_output_embeddings.weight.shape[0] == current_output_embeddings.weight.shape[0] + 1
    assert resized_input_embeddings.weight.shape[0] == current_input_embeddings.weight.shape[0] + 1
```

tuning/utils/tokenizer_data_utils.py
Abhishek-TAMU: Confirmed the change with Mehant and we can move `model.resize_token_embeddings(embedding_size)` into `if num_new_tokens > 0:` and It will be applicable even when `multiple_of` is > 1

tuning/utils/tokenizer_data_utils.py
Abhishek-TAMU: As per [this](https://github.com/foundation-model-stack/fms-hf-tuning/pull/451#discussion_r2025084688) comment we see that we might not need resize `if num_new_tokens = 0`.

tuning/utils/tokenizer_data_utils.py
Ssukriti: no we will have to find a dummy vision model for unit tests , or we ll have to create one ourselves with dummy files and trimmed down vocab and embeddings

tuning/utils/tokenizer_data_utils.py
Ssukriti: thank you for checking :)

tuning/utils/tokenizer_data_utils.py
Ssukriti: unit test otherwise looks good to me. I didnt understand why  2 tokens are added though, if we add one UNK  `# 2 new tokens were added: <unk> and <image>`. Does a new image token get added for every language token?

let us know in comment once you  assess feasibility of unit test. We should ideally have unit tests for some vision model

Abhishek-TAMU: > unit test otherwise looks good to me. I didnt understand why 2 tokens are added though, if we add one UNK  # 2 new tokens were added: <unk> and <image>. Does a new image token get added for every language token?

`<image>` token is present in Llama vision tokenizer and tokens length: `len(tokenizer) = 128257`
But `model.get_output_embeddings().weight.shape` is torch.Size([**128256**, 4096]). 

Hence when`special_tokens_dict` just has `<unk>`, then `num_new_tokens = num_new_tokens + embedding_size - len(tokenizer)` makes` embedding size = 128258` and hence model resizes from 128256 to 128258 (as it takes additional `<image>` token also in calculation)

Abhishek-TAMU: For `Granite`, `Llava` Vision model, 
`len(tokenizer)` == `model.get_output_embeddings().weight.shape[0]`

and as mentioned above for `Llama` vision model:
`len(tokenizer)` == `model.get_output_embeddings().weight.shape[0]` + 1

docs/vision-language-model-tuning.md
```diff
@@ -0,0 +1,174 @@
+# Tuning Vision Language Models
+Our library also supports full fine tuning and LoRA tuning for vision language models.
+
+## Supported Dataset Format
+We support tuning an `image+text` dataset that includes:
```
dushyantbehl, line 5:
```suggestion
We support tuning with an `image+text` dataset that includes:
```

tuning/utils/tokenizer_data_utils.py
Abhishek-TAMU: Upon checking failing unit test cases, I find out that even though `special_tokens_dict` is empty, and this `num_new_tokens = tokenizer.add_special_tokens(special_tokens_dict)` gives `num_new_tokens=0`, 
`TinyLlama` needs model resizing as model embedding size (32000) is lesser than Vocab size (32004) and tokenizer already have all the 32004 tokens which avoids the `if block`.

So `model.resize_token_embeddings(embedding_size)` is needed out of if block `if num_new_tokens > 0:`. I have made the change [here](https://github.com/foundation-model-stack/fms-hf-tuning/pull/451/commits/14afb72e74f2160f330a06b66a9e54d5b0002d14).

tuning/data/setup_dataprocessor.py
```diff
@@ -231,6 +233,45 @@ def _get_default_dataset_handlers(data_args, tokenizer_kwargs):
...
+    fn_kwargs2 = {
+        "fields_name": {
+            "dataset_text_field": data_args.dataset_text_field,
+            "dataset_image_field": data_args.dataset_image_field,
+        },
+        "processor_kwargs": processor_kwargs,
+    }
+    kwargs2 = {
+        "fn_kwargs": fn_kwargs2,
+        "batched": False,
```
dushyantbehl, line 265: `batched` would be default `False` right?

Also do we need `num_proc` to be `None`?

tuning/data/setup_dataprocessor.py
```diff
@@ -231,6 +233,45 @@ def _get_default_dataset_handlers(data_args, tokenizer_kwargs):
     return [handler], data_args.dataset_text_field
 
 
+### Vsion Data Format
+def _get_vision_dataset_handlers(data_args, processor_kwargs):
+
+    handlers = []
+
+    # First data handler configuration
+    fn_kwargs1 = {
```
dushyantbehl, line 242: minor nit.
`fn_kwargs1` to `handler_fn_kwargs1`
`kwargs1` to `handler_kwargs1`

tuning/data/data_processors.py
```diff
@@ -43,14 +43,19 @@
 class DataPreProcessor:
 
     tokenizer = None
+    image_processor = None
```
dushyantbehl, line 46: See my comment below about renaming this to `processor` for consistency

dushyantbehl (changes requested):
Few comments and request to go through the merge again due to inconsistency in code post rebasing.

tuning/sft_trainer.py
```diff
@@ -219,30 +225,55 @@ def train(
...
-    )
+    processor = None
+    try:
+        # try to load vision model
+        model_loader = AutoModelForVision2Seq.from_pretrained
+        model = model_loader(
+            model_args.model_name_or_path,
+            cache_dir=train_args.cache_dir,
+            torch_dtype=get_torch_dtype(model_args.torch_dtype),
+            attn_implementation="flash_attention_2"
```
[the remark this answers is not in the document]
Abhishek-TAMU: Tox formatting wraps it.

tuning/data/setup_dataprocessor.py
```diff
@@ -410,8 +484,13 @@ def process_dataargs(
         max_seq_length,
         data_args.instruction_template,
         is_padding_free=is_padding_free,
+        processor=processor,
     )
 
+    dataset_kwargs = {}
+    if processor is not None:
```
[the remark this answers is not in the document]
Abhishek-TAMU: Done.

tuning/data/setup_dataprocessor.py
Abhishek-TAMU: Yes, this is done. Thanks.

tuning/data/setup_dataprocessor.py
Abhishek-TAMU: Done

tuning/data/data_processors.py
Abhishek-TAMU: Done

tuning/data/setup_dataprocessor.py
```diff
@@ -251,13 +292,16 @@ def _process_raw_data_args(
     max_seq_length: int,
     additional_data_handlers: Dict[str, DataHandler] = None,
     is_padding_free: bool = False,
+    processor: AutoProcessor = None,
 ):
 
     # Create a data processor with default processor config
     default_processor_config = DataPreProcessorConfig()
+    default_processor_config.streaming = data_args.use_streaming_dataset
```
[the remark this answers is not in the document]
Abhishek-TAMU: For now I am disabling it use of streaming via arguments.

tuning/sft_trainer.py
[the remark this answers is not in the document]
Abhishek-TAMU: Sure, change is made.

tuning/utils/tokenizer_data_utils.py
```diff
@@ -104,7 +106,38 @@ def tokenizer_and_embedding_resize(
     )
     embedding_size = int(multiple_of * math.ceil(len(tokenizer) / multiple_of))
     num_new_tokens = num_new_tokens + embedding_size - len(tokenizer)
-    model.resize_token_embeddings(embedding_size)
+
+    if isinstance(model, MllamaForConditionalGeneration):
```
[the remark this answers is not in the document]
Abhishek-TAMU: Done.

tuning/data/data_preprocessing_utils.py
```diff
@@ -49,12 +59,18 @@ def get_data_collator(
...
+        processor:
+            Model processor to combine text and image data if using
+            multi-modal vision model.
 
     Returns:
         Callable
             Callable collator to be leveraged by the trainer.
     """
 
+    if processor:
```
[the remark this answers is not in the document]
Abhishek-TAMU: Yea in VisionCollator we are not handling `padding_free` and `packing`. I will add error message.

tuning/utils/collators.py
```diff
@@ -0,0 +1,100 @@
...
+from tuning.utils.utils import try_convert_bytes_dict_to_pil
+
+
+class VisionDataCollator:
+    """
+    A data collator specialized for multi-modal (text + image) inputs.
+    It uses a processor (e.g., LlavaProcessor or MllamaProcessor) to
+    combine text and images into model-ready tensors.
+
+    For padding-free tuning, configure the processor's arguments
```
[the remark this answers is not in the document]
Abhishek-TAMU: Have added the message: `Padding-free tuning is not supported.`

tuning/utils/collators.py
```diff
@@ -0,0 +1,95 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Local
+from tuning.utils.utils import try_convert_bytes_dict_to_pil
+
+
+class VisionDataCollator:
```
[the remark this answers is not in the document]
Abhishek-TAMU: Added.

tests/utils/test_embedding_resize.py
```diff
@@ -197,3 +206,41 @@ def test_resize_with_multiple_of():
...
+    resized_input_embeddings = model.get_input_embeddings()
+    resized_output_embeddings = model.get_output_embeddings()
+    resized_tokenizer_len = len(tokenizer.get_vocab())
+
+    assert resized_tokenizer_len == current_tokenizer_len + 1
+    assert "<unk>" in tokenizer.get_vocab()
+    assert resize_result["num_new_tokens"] == 1
+
+    # 2 new tokens were added: <unk> and <image>
+    assert (
```
dushyantbehl, line 239: @Abhishek-TAMU can you please add 2 lines from your comments here [1](https://github.com/foundation-model-stack/fms-hf-tuning/pull/451#discussion_r2025456404) and [2](https://github.com/foundation-model-stack/fms-hf-tuning/pull/451#discussion_r2025475597) which mention why have we added `2` new tokens.

tuning/data/setup_dataprocessor.py
```diff
@@ -231,6 +233,42 @@ def _get_default_dataset_handlers(data_args, tokenizer_kwargs):
...
+    handler_kwargs1 = {
+        "fn_kwargs": handler_fn_kwargs1,
+        "remove_columns": None,
+    }
+    handlers.append(
+        DataHandlerConfig("apply_tokenizer_chat_template", arguments=handler_kwargs1)
+    )
+
+    # Second data handler configuration
+    fn_kwargs2 = {
```
dushyantbehl, line 255:
```suggestion
    handler_fn_kwargs2 = {
```

and `handler_kwargs2` like above.

tests/test_sft_trainer.py
```diff
@@ -1969,3 +1974,51 @@ def test_handler(element, tokenizer, **kwargs):
             },
         )
         _validate_training(tempdir)
+
+
+def test_vision_data_collator():
```
dushyantbehl, line 1979: Thinking if this should be in `test_data_preprocessing.py`. 

On the same line should we move `utils/collators.py` to `data/collators.py`

dushyantbehl (dismissed):
Thanks @Abhishek-TAMU just last couple of minor changes requested and rest looks good to me. 

Please check the DCO before we can merge it.

tests/test_sft_trainer.py
Abhishek-TAMU: Sounds good.

dushyantbehl (approved):
Thanks a lot @Abhishek-TAMU for diligently fixing all review comments. 

LGTM.

willmj (approved):

PR #452, opened 2025-01-27
fix: Fix fips multiproc bug

dushyantbehl:
### Description of the change

Disable multi proc for python 3.9 till upgrade to python 3.12 is done.

### Was the PR tested

### DIFF
--- a/tuning/data/data_processors.py
+++ b/tuning/data/data_processors.py
@@ -271,8 +271,14 @@ class DataPreProcessor:
                     if kwargs["remove_columns"] == "all":
                         kwargs["remove_columns"] = column_names
 
-                    if "num_proc" not in kwargs:
-                        kwargs["num_proc"] = os.cpu_count()
+                    # The multiprocessing library in Python versions < 3.12 uses MD5,
+                    # which is prohibited on FIPS-enabled clusters due to strict
+                    # security policies. This issue is resolved in Python 3.12,
+                    # but until an upgrade is implemented, we disable multi processinng
+                    # as a temporary workaround.
+
+                    # if "num_proc" not in kwargs:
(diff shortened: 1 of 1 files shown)


tuning/data/data_processors.py
```diff
@@ -271,8 +271,12 @@ def _process_dataset_configs(
                     if kwargs["remove_columns"] == "all":
                         kwargs["remove_columns"] = column_names
 
-                    if "num_proc" not in kwargs:
-                        kwargs["num_proc"] = os.cpu_count()
+                    # Python 3.9 multi processing library uses md5 which is
```
anhuong, line 274: This is not specific to python 3.9 but occurs in versions of python < 3.12 correct? Since the initial error was found running on python 3.11

anhuong: also what is the impact of removing this portion and why is was it needed to set num_proc?

dushyantbehl: @anhuong you are right this should be <3.12 in the comment. 

`num_procs` is specified to run multiple threads to preprocess datasets while we run `map` on them. In our older version of fms-hf-tuning, before data preprocessor v1 the map was done without `num_procs` which means on a single cpu. In the data preprocessor v1 we introduced this to save time by utilising multi processing power of multi threaded cpus. 

Now the FIPS issue comes in because the threads use `md5` for state synchronization in python <3.12 and FIPS doesn't allow that...post 3.12 this has been changed to a different hash function.

dushyantbehl:
@anhuong @willmj are we parking this in favour of upgrade to python 3.12? In that case can we close this.

praveenj83:
Not needed now as the release has been made with python 3.12

PR #455, opened 2025-02-05
fix: space missing from data_formatter_template causing mismatch with response_template

HarikrishnanBalagopal:
### Was the PR tested

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,7 +133,7 @@ Example: Train.json
   },
  ...
 ]`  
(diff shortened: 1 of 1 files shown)


Abhishek-TAMU (approved):
Yea if `response template` is `\n## Label:` then this fix in `data_formatter_template` makes sense. Thank you for catching this. LGTM!

willmj (approved):

PR #456, opened 2025-02-05
fix: Using Sandboxed Environment in handler rendering Jinja template

Abhishek-TAMU:
### Description of the change

Modification as a part of this PR: https://github.com/foundation-model-stack/fms-hf-tuning/pull/438
Includes usage of `Sandboxed Environment` instead of normal `Environment` in rendering Jinja template which is more safer to use as it evaluates untrusted templates and restricts access to unsafe attributes/methods inside the template. 

### Related issue number

Issue: https://github.ibm.com/ai-foundation/watson-fm-stack-tracker/issues/1470

### How to verify the PR

Check relevant test case.

### Was the PR tested

### DIFF
--- a/tests/data/test_data_handlers.py
+++ b/tests/data/test_data_handlers.py
@@ -103,15 +103,23 @@ def test_apply_custom_formatting_template_gives_error_with_wrong_keys():
         )
 
 
-def test_apply_custom_formatting_jinja_template_gives_error_with_wrong_keys():
+@pytest.mark.parametrize(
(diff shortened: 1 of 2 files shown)


ashokponkumar (approved):

willmj (approved):
LGTM

Abhishek-TAMU:
@kmehant Does this PR looks good to merge ?

kmehant (approved):
We can merge, thanks

PR #458, opened 2025-02-07
fix: use main_process_first instead of broadcast_object_list

willmj:
### Description of the change
In the initial version of the data preprocessor, we implemented [this check](https://github.com/foundation-model-stack/fms-hf-tuning/blob/a6e4f9dedd007b264f84c06fe275597f7073bf47/tuning/data/data_processors.py#L334) to ensure that data processing occurs only on a single place `(rank 0)`, preventing unnecessary computation on other nodes. The processed dataset is then broadcasted to other ranks in a multi-node setup.

However, during actual runs with datasets of several gigabytes, we encountered broadcast timeout issues originating as `NCCL` timeouts. Using `accelerator.main_process_first` should resolve this issue.

This assumption works well in a single-node scenario, where it is assumed all processes in the node have access to the same memory location. In a multi-node scenario this assumption might not be true, and may need to be explored further.

Needs to be tested more extensively. Will be tested in [travis](https://v3.travis.ibm.com/github/ai-foundation/fmaas-tuning-docs/branches?serverType=git) on main-process-first-testci branch when token errors are resolved.

Started off by using an `Accelerator` however this might be too heavyweight, and an `Accelerator` is re-initialized in `SFTTrainer` for tuning the model. Having two `Accelerators` running at once could cause problems. So I am looking into using `PartialState` instead, which is [recommended](https://huggingface.co/docs/accelerate/v1.0.0rc1/en/package_reference/state#accelerate.PartialState) to be used for process control, which is what we are doing with the preprocessor.

Previous behavior:
```
  [Rank 1] Not processing data. Setting train_dataset to None
[Rank 1] Preparing to broadcast dataset from rank 0. Current train_dataset: None
[Rank 0] Before processing, dataset_configs: [DataSetConfig(name='training_data', data_paths=['/testing/tuning/input/cc_tone_sft_format_1000_train.json'], builder=None, sampling=None, data_handlers=[DataHandlerConfig(name='apply_dataset_formatting', arguments={'fn_kwargs': {'dataset_text_field': 'output'}, 'batched': False})])]
Map (num_proc=256): 100%|█████████████████████████████████████████████████████████████| 1000/1000 [00:29<00:00, 33.72 examples/s]
[Rank 0] After processing, train_dataset: Dataset({
    features: ['output'],
    num_rows: 1000
})
[Rank 0] Preparing to broadcast dataset from rank 0. Current train_dataset: Dataset({
    features: ['output'],
    num_rows: 1000
})
[Rank 0] After broadcast, received train_dataset: Dataset({
    features: ['output'],
    num_rows: 1000
})
/home/tuning/.local/lib/python3.11/site-packages/transformers/training_args.py:2027: FutureWarning: `--push_to_hub_token` is deprecated and will be removed in version 5 of 🤗 Transformers. Use `--hub_token` instead.
  warnings.warn(
[Rank 1] After broadcast, received train_dataset: Dataset({
    features: ['output'],
    num_rows: 1000
})
 
```
 
After behaviour:

```
[Main Process] Finished processing dataset. train_dataset: Dataset({
    features: ['output'],
    num_rows: 1000
})
[Process 1] Waiting for main process to process the dataset...
[Process 0] Exiting function with train_dataset: Dataset({
    features: ['output'],
    num_rows: 1000
})
/home/tuning/.local/lib/python3.11/site-packages/transformers/training_args.py:2027: FutureWarning: `--push_to_hub_token` is deprecated and will be removed in version 5 of 🤗 Transformers. Use `--hub_token` instead.
  warnings.warn(
[Process 1] Joined after main process. Current train_dataset: Dataset({
    features: ['output'],
    num_rows: 1000
})
[Process 1] Exiting function with train_dataset: Dataset({
    features: ['output'],
    num_rows: 1000
})
```

PartialState (in data preprocessor on CPU):
```
Distributed environment: DistributedType.NO
Num processes: 1
Process index: 0
Local process index: 0
Device: mps
```

PartialState (in data preprocessor on multi GPU):
```
Distributed environment: DistributedType.MULTI_GPU  Backend: nccl
Num processes: 4
Process index: 0
Local process index: 0
Device: cuda:0
Distributed environment: DistributedType.MULTI_GPU  Backend: nccl
Num processes: 4
Process index: 2
Local process index: 2
Device: cuda:2
Distributed environment: DistributedType.MULTI_GPU  Backend: nccl
Num processes: 4
Process index: 1
Local process index: 1
Device: cuda:1
Distributed environment: DistributedType.MULTI_GPU  Backend: nccl
Num processes: 4
Process index: 3
Local process index: 3
Device: cuda:3
```

Accelerator (in trainer):
```
<accelerate.accelerator.Accelerator object at 0x309da9fd0>
```

remaining steps:
- [x] Make sure Accelerator is being initialized correctly since there are multiple opportunities for it to get initialized throughout sft_trainer.py
  - This is being initialized multiple times, and multiple accelerators can cause problems, looking into using PartialState
- [x] e2e tests - successful with [Accelerator](https://v3.travis.ibm.com/github/ai-foundation/fmaas-tuning-docs/builds/34405033)

### Was the PR tested


willmj:
- [ ] Make sure Accelerator is being initialized correctly since there are multiple opportunities for it to get initialized throughout sft_trainer.py

willmj:
Streaming changes + main process first together:
```
100%|██████████| 2200/2200 [25:37<00:00,  1.43it/s]
```
Model location: `/testing/tuning/output/granite31-8b/ft/tone_20250214_1505-streaming-dataconfig-10-datasets/save_model`
F1 score: 
```
"f1": {
"macro": 0.29570348178371664,
"micro": 0.46292134831460674
},
```
Inference:
```
input: ### Text: It was unexpected, and it made me really happy\n\n### Response:
response:  satisfied
stop_reason: 2
```

kmehant:
@willmj can we fastrack this and streaming PRs? Thank you.

Ssukriti:
the pending part in this PR is testing on multi-node, regarding comment 

> This assumption works well in a single-node scenario, where it is assumed all processes in the node have access to the same memory location. In a multi-node scenario this assumption might not be true, and may need to be explored further.

Seems to be well tested on single-node, but we need someone to test it out on multi-node, to see if reported issue is resolved, before we can merge it to main . @ashokponkumar could you please assign that test to someone so we can wrap up this work?

Ssukriti:
if it works with multi-node as well, please approve and merge it. I dont see any changes remaining. Lets start getting PRs merged into main

kmehant:
@willmj Do we even need to do sequencing to allow main process to complete the data process first and then let other processes do the same thing? Trying to understand with a simpler case where you allow all processes do the same thing in any order they wish?

dushyantbehl:
> @willmj Do we even need to do sequencing to allow main process to complete the data process first and then let other processes do the same thing? Trying to understand with a simpler case where you allow all processes do the same thing in any order they wish?
> 
> cc: @dushyantbehl

@kmehant the reason to execute this on main process first is to avoid compute on all processes for the thing which is going to be a repeated work...this is a design decision to not have multiple processes repeat this work...

dushyantbehl:
> the pending part in this PR is testing on multi-node, regarding comment
> 
> > This assumption works well in a single-node scenario, where it is assumed all processes in the node have access to the same memory location. In a multi-node scenario this assumption might not be true, and may need to be explored further.
> 
> Seems to be well tested on single-node, but we need someone to test it out on multi-node, to see if reported issue is resolved, before we can merge it to main . @ashokponkumar could you please assign that test to someone so we can wrap up this work?

@Ssukriti the thing that we should do here is to go ahead with a single node support first and add multi node support later if required calling it out as a limitation in our document directly.

cc @ashokponkumar @willmj

kmehant:
https://github.com/foundation-model-stack/fms-hf-tuning/pull/458#issuecomment-2693569777
This is to reuse map HF datasets caches? Then it makes sense.

tuning/data/data_processors.py
```diff
@@ -349,23 +349,10 @@ def process_dataset_configs(
...
-            # TODO: Check if torch.distributed.barrier() is called in broadcast_object_list()
-            # See https://github.com/pytorch/pytorch/issues/56142
-            # for why the list is shared like this
-            to_share = [train_dataset]
-            torch.distributed.broadcast_object_list(to_share, src=0)
-            train_dataset = to_share[0]
-        else:
-            logger.info("Processing data...")
+        # The main_process_first context ensures that the main process runs first
+        with state.main_process_first():
```
kmehant, line 355:
```suggestion
        with state. local_main_process_first():
```

This will be relevant when each hf cache folder is different for each node, for instance, we point to `/dev/shm`.

dushyantbehl: @willmj +1 can you test by using local main process first and we can ensure this works for both multi and single node.

tuning/data/data_processors.py
```diff
@@ -393,23 +393,10 @@ def process_dataset_configs(
...
-            # Use broadcast_object_list to share the dataset object across ranks
-            # TODO: Check if torch.distributed.barrier() is called in broadcast_object_list()
-            # See https://github.com/pytorch/pytorch/issues/56142
-            # for why the list is shared like this
-            to_share = [train_dataset]
-            torch.distributed.broadcast_object_list(to_share, src=0)
-            train_dataset = to_share[0]
-        else:
-            logger.info("Processing data...")
+        # The local_main_process_first context ensures that the main process runs first per node
```
dushyantbehl, line 398: Please expand the comment here to include our rationale of using local main process first as we want to reuse HF cache and not redo computation all the time.

tuning/data/data_processors.py
```diff
@@ -393,23 +393,10 @@ def process_dataset_configs(
         self, dataset_configs: List[DataSetConfig], **kwargs
     ) -> Union[Dataset, IterableDataset]:
         train_dataset = None
+        state = PartialState()
```
dushyantbehl, line 396: Is partial state always available? do we need to check if it gets initialised properly?

dushyantbehl: Could you also add here the reasoning you mentioned in the PR on why Partial state is preferred and possibly link with the HF docs to mention that its the recommended pattern. 
you can point to the code too https://github.com/huggingface/trl/blob/main/trl/trainer/sft_trainer.py#L367

dushyantbehl (commented):
Please add documentation on the possiblity of NCCL timeouts and how to remediate them using `ddp_timeout` flag to the command. 
For now adding documentation should be good.

PR #466, opened 2025-02-12
feat: Add support for renaming and retaining columns in data preprocessor

dushyantbehl:
### Description of the change

Add two flags to the `data_config` under `dataset` definition. 
1. `rename_columns` which allows users to rename columns by passing a dict
2. `retain_columns` which allows users to specify which columns to retain and drop others.

This is different and outside the `remove_columns` argument to `data handlers` because people might want to use this functionality even without invoking data handlers.

This is especially needed in the case of interleaving multiple datasets with different features.

### Was the PR tested

### DIFF
--- a/docs/advanced-data-preprocessing.md
+++ b/docs/advanced-data-preprocessing.md
@@ -60,6 +60,10 @@ definitions:
         type: float
       builder:
         type: string
+      rename_columns:
+        type: object
+      retain_columns:
+        type: object
       data_paths:
         type: array
         items:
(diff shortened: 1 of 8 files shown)


willmj (approved):
This looks good to me, this might require some investigation with iterable datasets after it is merged in.

willmj:
Should we add one more test in `test_datapreprocessing` to assert column names have been appropriately retained and renamed?

dushyantbehl:
> This looks good to me, this might require some investigation with iterable datasets after it is merged in.

I checked the Iterable datasets API and it does contain [`rename`](https://github.com/huggingface/datasets/blob/fba47587a4f1771ad9f934a5e66f0860231405cd/src/datasets/iterable_dataset.py#L2777) and [`retain`](https://github.com/huggingface/datasets/blob/fba47587a4f1771ad9f934a5e66f0860231405cd/src/datasets/iterable_dataset.py#L2837) APIs but you can add a few tests to your PR for sanity.

Abhishek-TAMU (dismissed):
Apart for this small nit, LGTM! Thanks Dushyant!

willmj (approved):

PR #467, opened 2025-02-12
feat: adding eos token to be made a flag so we don't force it on every handler

dushyantbehl:
### Description of the change

Add a flag `add_eos_token` to every handler which is set to `True` by default but can be disabled.

### Was the PR tested


willmj (dismissed):
lgtm, thanks Dushyant

docs/advanced-data-preprocessing.md
```diff
@@ -204,14 +204,17 @@ Users can also pass any number of `kwargs` arguments required for each data hand
...
     Appends the tokenizer's EOS token to a specified dataset field.
  - `apply_custom_data_formatting_template`:
     Applies a custom template (e.g., Alpaca style) to format dataset elements.
+    By default this handler adds `EOS_TOKEN` which can be disabled by a handler argument, [see](https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/tuning/data/data_handlers.py)
+ - `tokenize_and_apply_input_masking`:
+    Tokenizes input text and applies masking to the labels for causal language modeling tasks, good for input/output datasets.
+    By default this handler adds `EOS_TOKEN` which can be disabled by a handler argument, [see](https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/tuning/data/data_handlers.py) 
  - `apply_custom_jinja_template`:
     Applies a custom jinja template (e.g., Alpaca style) to format dataset elements.
+    By default this handler adds `EOS_TOKEN` which can be disabled by a handler argument, [see](https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/tuning/data/data_handlers.py)
```
Abhishek-TAMU, line 217: I think I might be missing something in this PR. How can user pass the value of `add_eos_token` = False, as for example [here](https://github.com/foundation-model-stack/fms-hf-tuning/blob/a89a4a336487b3bda5bde31f9a7b58a0563e3e02/tuning/data/setup_dataprocessor.py#L196) in `fn_kwargs` we don't pass the value of `add_eos_token` given by the user to the handler. Can I ask what am I missing ? 

Probably a small test case when `add_eos_token` = False in [test_data_preprocessing.py](https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/tests/data/test_data_preprocessing.py) would be appreciated.

dushyantbehl: @Abhishek-TAMU this is only for `data_config` where people can specify it and we disable adding `EOS_TOKEN` to the code. 
This is not for our cli based args flow for where the function you pointed out to is used which for specific instruction masking use case on a single dataset file and hence in that part of the code we making sure its already added is ensured. 

We can try to add a data config test where we set the flag to `false` and  check the behaviour.

Abhishek-TAMU: Okay I see. Yea user can pass it in `fn_kwargs` in data_config.yaml
Feel free to add test case for the same.

Abhishek-TAMU:
Added test case and modified URL of the documentation. 
@dushyantbehl @willmj

tests/artifacts/predefined_data_configs/apply_custom_jinja_template.yaml
```diff
@@ -11,4 +11,5 @@ datasets:
           batched: false
           fn_kwargs:
             dataset_text_field: "dataset_text_field"
-            template: "dataset_template"
\ No newline at end of file
+            template: "dataset_template"
+            add_eos_token: true
```
willmj, line 15: Since it defaults to `true` do we need to add this here?

tests/data/test_data_preprocessing.py
```diff
@@ -748,18 +794,30 @@ def test_process_dataconfig_file(data_config_path, data_path):
         data_args = configs.DataArguments(data_config_path=temp_yaml_file_path)
 
     tokenizer = AutoTokenizer.from_pretrained(MODEL_NAME)
+    tokenizer.add_special_tokens({"eos_token": "</s>"})
     (train_set, _, _) = _process_dataconfig_file(data_args, tokenizer)
     assert isinstance(train_set, Dataset)
     if datasets_name == "text_dataset_input_output_masking":
         column_names = set(["input_ids", "attention_mask", "labels"])
         assert set(train_set.column_names) == column_names
+        print("INFO", train_set[8]["input_ids"], tokenizer.eos_token_id)
```
willmj, line 803: Remove print statement here

tests/data/test_data_preprocessing.py
```diff
@@ -682,39 +682,75 @@ def test_process_data_args_throws_error_where_needed(data_args, packing):
 
 
 @pytest.mark.parametrize(
-    "data_config_path, data_path",
+    "data_config_path, data_path, add_eos_token",
```
willmj, line 685: Should we do a separate unit test for this with `add_eos_token`?

tests/artifacts/predefined_data_configs/apply_custom_jinja_template.yaml
Abhishek-TAMU: I added it here to point this config file in the documentation (As you can see the Url in documentation of this PR) for the user to know that they can use this flag to disable `EOS_TOKEN` from the handler. 
And additionally I am assigning it value `False` in Unit test.

tests/data/test_data_preprocessing.py
Abhishek-TAMU: That's a good idea. Pushed the changes. Thanks!

dushyantbehl (commented):
LGTM .... Thanks @Abhishek-TAMU

willmj (approved):
LGTM! Thanks Abhishek

PR #468, opened 2025-02-13
feat: Packing for pretokenised

dushyantbehl:
### Description of the change

This PR adds support for packing for pretokenized datasets which was added in `transformers>=4.46`

This is based on the PR #448 

### Was the PR tested


kmehant:
1. max_seq_len is too long in your testcase that it was not able to generate not even 1 sample so it was failing
2. you should pass seq2seq collator with padding=False when it is pretokenized but it seem to pick up completiononly collator.

tuning/data/setup_dataprocessor.py
```diff
@@ -396,10 +396,6 @@ def process_dataargs(
         is_padding_free=is_padding_free,
     )
 
-    dataset_kwargs = {}
-    if is_tokenized_dataset:
-        dataset_kwargs["skip_prepare_dataset"] = True
-
```
kmehant, line 402: Any reason we dont want to skip anymore?

dushyantbehl: yes...added explanation here https://github.com/foundation-model-stack/fms-hf-tuning/pull/468#issue-2850878477
We need to remove this check to call `prepare_dataset` and and enable packing for the pretokenized dataset.

kmehant: Just checked, looks like it checks for tokenized with trl and skips prepare.

kmehant (approved):
LGTM

tuning/data/setup_dataprocessor.py
dushyantbehl: This line is what we need - https://github.com/huggingface/trl/blob/a34987956cd5bf08ed7501da2510b9404bede695/trl/trainer/sft_trainer.py#L467

willmj (approved):
LGTM

Abhishek-TAMU (approved):
LGTM

PR #453, opened 2025-01-27
chore: Merge in changes for v2.5.0 release

aluu317:
### Was the PR tested

- [ ] I have added >=1 unit test(s) for every new method I have added.
- [ ] I have ensured all unit tests pass

PR #472, opened 2025-02-18
chore:Merge set of changes for v2.6.0

Abhishek-TAMU:
### Was the PR tested


anhuong:
Two small comments, the first commit should have been in the last release -- upgrading python version -- and then the last commit which you did you should sign. The DCO will likely still fail but good to fix for the ones we can/just committed.

Abhishek-TAMU:
@anhuong The commit `Merge pull request #449 from dushyantbehl/upgrade-python-version` which had original commit ID as [3daf5b](https://github.com/foundation-model-stack/fms-hf-tuning/commit/3daf5bb18b156ef2c2e231122ca3a3591186219c) (PR to main branch of `fms-hf-tuning`), is changed in the last release [PR](https://github.com/foundation-model-stack/fms-hf-tuning/pull/453) to the `release` branch to commit ID `8d5fbf0`. 

Hence due to this,   the changes of this PR is already there in the `release` branch which went in last release with commit ID `8d5fbf0`, but as the same commit has different commit ID in main branch `3daf5b`, it shows up when we compare `main vs release` branch [here](https://github.com/foundation-model-stack/fms-hf-tuning/compare/release...main) though there is no files change.

willmj (approved):
This LGTM, if we don't pull in the duplicate change from `main` now I believe it would appear on every following release since the diff in commits would always be there.

anhuong (approved):

anhuong:
We can merge in the changes and be wary of this going forward. If we have a PR with a set of changes for release, we should not `git merge main` to pull in additional changes from main branch as this changes the commit hashes . This one is further complicated by the fact that the change was a merge commit and not a squash commit which makes it harder to pull in or to fix the release branch. I was just trying on my fork but looks complicated and even cherry-picking the single commit doesn’t bring it over properly

PR #474, opened 2025-02-18
feat: add support for chat template from data config

YashasviChaurasia:
### Description of the change
Adds support for `chat_template`/`chat_template_b64` from data config.
adds fix for "/n" parsing issue for passing chat_template via cli args

Updates test case for chat_template through data config

### Was the PR tested

### DIFF
--- a/tuning/data/setup_dataprocessor.py
+++ b/tuning/data/setup_dataprocessor.py
@@ -75,6 +75,16 @@ def _process_dataconfig_file(
         tokenizer=tokenizer,
         additional_data_handlers=additional_data_handlers,
     )
+
+    if processor.processor_config.chat_template is not None:
+        if tokenizer.chat_template:
+            logger.warning(
+                "replacing existing chat_template %s with data config's chat_template %s",
+                tokenizer.chat_template,
+                processor.processor_config.chat_template,
+            )
+        tokenizer.chat_template = processor.processor_config.chat_template
+
     if processor.processor_config.streaming:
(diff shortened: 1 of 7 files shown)


tuning/data/setup_dataprocessor.py
```diff
@@ -75,6 +75,12 @@ def _process_dataconfig_file(
         tokenizer=tokenizer,
         additional_data_handlers=additional_data_handlers,
     )
+
+    logger.info("Replaces default chat template with user's custom chat template")
+    print(processor.processor_config.chat_template)
```
dushyantbehl, line 80: remove print statements

dushyantbehl (dismissed):
LGTM.

tuning/sft_trainer.py
```diff
@@ -250,6 +250,10 @@ def train(
     )
 
     if data_args.chat_template:
+        # TODO: passing "/n" through cli causes parsing issues,
```
ashokponkumar, line 253: should this be part of this PR or a separate PR?

dushyantbehl: Its fine to fix it as part of this PR as it touches most of the chat template functionality @ashokponkumar 
Its called out as a temp fix via the `TODO` we can get to a better solution later.

ashokponkumar: sure, lets go ahead then.

willmj (approved):
LGTM

PR #475, opened 2025-02-18
refactor: refactor set special tokens function and add unit tests.

Luka-D:
### Description of the change

Making a draft PR for everyone's thoughts on unit tests so far. 

#### Missing Pad Token, LlamaTokenizerFast:
The first test uses a `LlamaTokenizerFast` tokenizer. This tokenizer is only missing a PAD token, however because it is a `LlamaTokenizer`, the function code automatically adds the bos, eos, unk and pad tokens to the special tokens dict. Then, the `<pad>` token is replaced with a `<PAD>` token, because the Llama tokenizer does not have a pad token specified.

#### EOS = PAD:
The second test uses a `GPT2TokenizerFast`tokenizer. This tokenizer is the case where the EOS token = PAD token, both of them are `<|endoftext|>`. So, the pad token in the tokenizer is set to `<PAD>` and the `"pad_token": "<PAD>"` is also added to the special tokens dict.

#### Missing Pad Token:
The third test uses a `GPTNeoXTokenizerFast`tokenizer. This tokenizer is another one that is hardcoded into the function to automatically add just a pad token to the special tokens dict. However, the tokenizer itself is also missing a pad token, so the function then replaces the `<pad>` token with the default `<PAD>` token. 

#### Missing all tokens:
Added in https://github.com/foundation-model-stack/fms-hf-tuning/commit/781ce58a24c1006e271045b47ec3792e79d7d496. This uses the IBM Granite tokenizer and removes all special tokens. The result is that the special tokens dict contains the PAD, EOS, BOS and UNK tokens.

### Related issue number

Related to [Issue #1515](https://github.ibm.com/ai-foundation/watson-fm-stack-tracker/issues/1515)

### How to verify the PR

You can run:
```tox -e py -- tests/utils/test_tokenizer_data_utils.py```

### Was the PR tested

### CODE REVIEW

**Comment by Luka-D** [comment]
With commit https://github.com/foundation-model-stack/fms-hf-tuning/pull/475/commits/781ce58a24c1006e271045b47ec3792e79d7d496, I have added a new unit test for when a tokenizer has no special tokens. I figured it was easier to test a tokenizer that was missing all the tokens, instead of doing 4 unit tests for missing EOS, BOS, PAD etc. I can split it into 4 unit tests if you feel that is a better idea.

I also changed the conditional for the case of EOS = PAD token. Because both tokens are set to `None`, the conditional `tokenizer.pad_token == tokenizer.eos_token` would evaluate to `True`. I am unsure if we want to have the warning appear when both tokens are `None`, since I wouldn't really say they are identical, and the defaults that are added in the code above make them different anyways (PAD is set to `<PAD>` and EOS is set to `</s>`). 
I have added  `if (tokenizer.pad_token is not None and tokenizer.pad_token == tokenizer.eos_token )`to fix this. 

Please let me know your thoughts on this. Thank you

**Comment by Luka-D** [comment]
### Changes with Commit https://github.com/foundation-model-stack/fms-hf-tuning/pull/475/commits/ad3c0bff0cff2f29773c8839cfc513c871938209

Added three unit tests for the tokenizer_and_embedding_resize() function.

#### Unit Test 1:
Tests the resizing when the special tokens dict contains a PAD token, which means the tokenizer is missing one special token. `mulitple_of` is set to 1.

#### Unit Test 2:
Tests the resizing when the special tokens dict contains a PAD, EOS, BOS and Unk token, which means the tokenizer is missing four special tokens. `mulitple_of` is set to 1.

#### Unit Test 3:
Tests the resizing when the special tokens dict contains a PAD, EOS, BOS and Unk token, which means the tokenizer is missing four special tokens. `mulitple_of` is set to 2, this add one to the count of `num_new_tokens` and adds one to the count of `new_embedding_size`.

I'm not sure if these are the tests we would like, please let me know your thoughts and which tests I've missed. Thank you

File: tuning/utils/tokenizer_data_utils.py [hunk 1]
```diff
@@ -14,10 +14,74 @@
...
 
 # Third Party
 import transformers
+from tuning.config import configs
+
+logger = logging.getLogger(__name__)
+
+
+def set_special_tokens_dict(
+    model_args: configs.ModelArguments, tokenizer: transformers.PreTrainedTokenizer
```
aluu317, line 28: nit: I think this utils file doesn't need to import `configs.ModelArguments`. So `model_args` can just be `tokenizer_name_or_path: str`  because that's the only thing from `model_args` we need.

tuning/utils/tokenizer_data_utils.py
```diff
@@ -14,10 +14,74 @@
...
+            special_tokens_dict["eos_token"] = "</s>"
+            special_tokens_dict["unk_token"] = "<unk>"
+            special_tokens_dict["pad_token"] = "<pad>"
+        elif isinstance(
+            tokenizer, (transformers.GPT2Tokenizer, transformers.GPTNeoXTokenizerFast)
+        ):
+            special_tokens_dict["pad_token"] = "<pad>"
+
+    # add special tokens only when a custom tokenizer is not passed
+    if not model_args.tokenizer_name_or_path:
```
aluu317, line 57: unsure why the original code had 2 of the same `if` statement. can you remove one?

Luka-D: Yep, the second if statement has been removed with https://github.com/foundation-model-stack/fms-hf-tuning/pull/475/commits/aee44d7c40303d9fc118e815affe35a4a1de3ecc

tuning/utils/tokenizer_data_utils.py
Luka-D: I changed the first arg to `tokenizer_name_or_path: str` with commit https://github.com/foundation-model-stack/fms-hf-tuning/pull/475/commits/aee44d7c40303d9fc118e815affe35a4a1de3ecc. I agree that it looks better, however I still cannot remove the import statement `from tuning.config import configs` because it is needed for the default tokens, eg. `configs.DEFAULT_PAD_TOKEN`.

tests/utils/test_tokenizer_data_utils.py
```diff
@@ -1,20 +1,126 @@
...
+    )
+    assert special_tokens_dict == {
+        "pad_token": "<PAD>",
+        "eos_token": "</s>",
+        "bos_token": "<s>",
+        "unk_token": "<unk>",
+    }
+
+
+def test_setting_special_tokens_when_path_is_not_none():
```
Luka-D, line 77: I have also added one extra unit test with https://github.com/foundation-model-stack/fms-hf-tuning/pull/475/commits/aee44d7c40303d9fc118e815affe35a4a1de3ecc. It's a small one that just checks to make sure that the `special_tokens_dict` is empty if any actual string is passed as `tokenizer_name_or_path`.

aluu317:
The fmt is failing.. Please run `tox -e fmt` on your branch and commit changes.

First glance, this is awesome, this test file is actually covered 100% so I'd say we're pretty good! https://github.com/foundation-model-stack/fms-hf-tuning/actions/runs/13838569047/job/38719830349#step:4:47

I'll look closely at the unit tests and see if I have any other comments!

Luka-D:
Thanks Angel! 

I have added the unit test descriptions with commit https://github.com/foundation-model-stack/fms-hf-tuning/pull/475/commits/98fff15c325f223339f3ac2c1061b4130ba53496.

aluu317:
Changing this PR label to a `refactor` but otherwise this is good to go! Thank you @Luka-D

aluu317 (approved):

dushyantbehl:
@Luka-D could you update the branch to latest before a final review

Luka-D:
@dushyantbehl Branch has been updated, thank you.

tuning/utils/tokenizer_data_utils.py
```diff
@@ -14,11 +14,76 @@
...
+
+
+def set_special_tokens_dict(
+    tokenizer_name_or_path: str, tokenizer: transformers.PreTrainedTokenizer
+) -> dict:
+    """Creates a special tokens dictionary and sets the special tokens,
+       depending on the tokenizer.
+
+    Args:
+        model_args: configs.ModelArguments.
```
Abhishek-TAMU, line 36: Nit:
```suggestion
        tokenizer_name_or_path: configs.ModelArguments.tokenizer_name_or_path
```

Abhishek-TAMU (commented):
Thanks @Luka-D, for the refactor and for adding unit test cases with different models. 
Apart from a small nit, it LGTM.

Abhishek-TAMU (approved):
LGTM!

tuning/utils/tokenizer_data_utils.py
```diff
@@ -14,11 +14,76 @@
...
+            special_tokens_dict["eos_token"] = "</s>"
+            special_tokens_dict["unk_token"] = "<unk>"
+            special_tokens_dict["pad_token"] = "<pad>"
+        elif isinstance(
+            tokenizer, (transformers.GPT2Tokenizer, transformers.GPTNeoXTokenizerFast)
+        ):
+            special_tokens_dict["pad_token"] = "<pad>"
+
+        # Add special tokens only when a custom tokenizer is not passed
+        # TODO: we need to change this, perhaps follow what open instruct does?
```
dushyantbehl, line 59: @Luka-D may I bring your attention to this...considering we are putting effort into refactoring this change and are already exploring this could you check if OI does things very differently and we should discuss before pushing this PR.

dushyantbehl (changes requested):

tuning/utils/tokenizer_data_utils.py
Luka-D: It seems like the biggest difference between our code and OI's code is when a custom tokenizer is passed in.

In the OI code, they only add a PAD token:
```sh
# NOTE: (Costa) I just commented the `OPTForCausalLM` because we are not likely to use it.
# elif isinstance(tokenizer, GPT2Tokenizer) and isinstance(model, OPTForCausalLM):
#     num_added_tokens = tokenizer.add_special_tokens({"unk_token": "<unk>"})
elif isinstance(tokenizer, transformers.PreTrainedTokenizerFast) and tokenizer.pad_token is None:
   num_added_tokens = tokenizer.add_special_tokens({"pad_token": "<pad>"})
   assert num_added_tokens == 1, "We detected no padding token but add_special_tokens did not add one."
```

Meanwhile, our code currently checks all of the special tokens, and if any of them are missing, it adds the default value of that special token into the dict:
```sh
# Add special tokens only when a custom tokenizer is not passed
        # TODO: we need to change this, perhaps follow what open instruct does?
        if tokenizer.pad_token is None:
            logger.warning("PAD token set to default, missing in tokenizer")
            special_tokens_dict["pad_token"] = configs.DEFAULT_PAD_TOKEN
        if tokenizer.eos_token is None:
            logger.warning("EOS token set to default, missing in tokenizer")
            special_tokens_dict["eos_token"] = configs.DEFAULT_EOS_TOKEN
        if tokenizer.bos_token is None:
            logger.warning("BOS token set to default, missing in tokenizer")
            special_tokens_dict["bos_token"] = configs.DEFAULT_BOS_TOKEN
        if tokenizer.unk_token is None:
            logger.warning("UNK token set to default, missing in tokenizer")
            special_tokens_dict["unk_token"] = configs.DEFAULT_UNK_TOKEN
        if (
            tokenizer.pad_token is not None
            and tokenizer.pad_token == tokenizer.eos_token
        ):
            logger.warning(
                "PAD token set to default, to make it different from eos token"
            )
            if tokenizer.eos_token != configs.DEFAULT_PAD_TOKEN:
                tokenizer.pad_token = configs.DEFAULT_PAD_TOKEN
                special_tokens_dict["pad_token"] = configs.DEFAULT_PAD_TOKEN
            else:
                tokenizer.eos_token = configs.DEFAULT_EOS_TOKEN
                special_tokens_dict["eos_token"] = configs.DEFAULT_EOS_TOKEN
```

dushyantbehl:
@Abhishek-TAMU @Luka-D for any such refactor its better if we utilise the refactored top level API in the smaller scripts like our offline processing one to ensure same functionality is retained in both the scripts
Please use the refactored API here too - 
https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/scripts/offline_data_processing.py

tuning/utils/tokenizer_data_utils.py
dushyantbehl: Thanks!
in that case i would advise to drop the Todo as it seems redundant.

Luka-D:
@Abhishek-TAMU PR has been updated

Abhishek-TAMU:
Thanks Luka. @dushyantbehl can approve it once you address [this](https://github.com/foundation-model-stack/fms-hf-tuning/pull/475#discussion_r2027525599).

tuning/utils/tokenizer_data_utils.py
Luka-D: Should I remove the TODO on line 45 as well?
```# TODO: understand if we need to hardcode these here or just use defaults in model```

**Comment by dushyantbehl** [comment]
> > @Abhishek-TAMU @Luka-D for any such refactor its better if we utilise the refactored top level API in the smaller scripts like our offline processing one to ensure same functionality is retained in both the scripts
> > Please use the refactored API here too -
> > https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/scripts/offline_data_processing.py
> 
> Yea, once this PR is merged, we can make the change in `offline_data_processing.py` too. @Luka-D Feel free to update branch and we can merge this PR then.

@Abhishek-TAMU I really want to avoid [this](https://github.com/foundation-model-stack/fms-hf-tuning/pull/517/files#diff-9f6d56b18e5c81aaa3d14e3ae6528b69fb39644d5d22f30cee5a7325ca92a1bbR135) type of fixes done in two places 

This is a small change so lets go ahead in this PR itself...if @Luka-D doesn't have bandwidth could you please help him.

File: tuning/utils/tokenizer_data_utils.py (hunk 5 above)
Comment (dushyantbehl) [reply tuning/utils/tokenizer_data_utils.py:59]: you can keep the second one for now..I believe someone needs to take a look at all this tokenizer handling at later point and decide how we want to do this...we can leave that to be part of later changes.

File: scripts/offline_data_processing.py [hunk 6]
```diff
@@ -92,36 +93,9 @@ def get_processed_dataset(
         tokenizer.chat_template = data_args.chat_template
 
     # Prepare special tokens dictionary
-    special_tokens_dict = {}
```
Luka-D, line 95: @dushyantbehl @Abhishek-TAMU, I have refactored `offline_data_processing.py` to use the new `set_special_tokens_dict` function instead. All tox unit tests have passed locally.

Abhishek-TAMU: Thanks @Luka-D

tuning/utils/tokenizer_data_utils.py
```diff
@@ -14,11 +14,75 @@
...
 # Third Party
 import transformers
 
+# Local
+from tuning.config import configs
+
+logger = logging.getLogger(__name__)
+
+
+def set_special_tokens_dict(
```
dushyantbehl, line 29: Thanks a lot @Luka-D for refactoring offline processing script

Just a minor nit and rest looks good to me...could this be renamed to `get_special_tokens_dict`?

Luka-D: I think we could go with either name. I think it was `set` before because we were creating a dict where the special tokens depended on the tokenizer, within the code. Now that it is a separate function that returns a dict, `get` could work as well. I'll leave the decision up to the team and I'll change it if we feel `get` is more appropriate.

dushyantbehl: @Abhishek-TAMU any thoughts here? my thought is that this is returning a dict rather than setting it inside the function so suggesting `get` rather than `set`.

Abhishek-TAMU: `get_special_tokens_dict` also sounds good to me.

dushyantbehl: @Luka-D whenever you rename and update the branch please let me know..will be happy to merge it.

Luka-D: @dushyantbehl I have changed it to `get_special_tokens_dict` and have updated the branch. Thank you

dushyantbehl (approved):
LGTM

PR #485, opened 2025-03-04
build(deps): upgrade transformers to 4.49

anhuong:
### Was the PR tested


dushyantbehl (approved):
Approving as all unit test cases pass and this is a time critical feature.

Issue #480, opened 2025-02-27
feat: Include checkpoint safetensor + HF compatible recovery as a default behaviour


kmehant:
## Is your feature request related to a problem? Please describe.

Currently with the use of EP through `--fast_moe` saves checkpoints in PyTorch DCP format. Though `fms-acceleration` has utility CLI script to convert such a checkpoint to safetensors, its useful to include this recovery/conversion process as a default behaviour in `fms-hf-tuning` so that users would not be needing to use additional script every time they train with our stack.

## Describe the solution you'd like

Couple of solutions:
1. Change can be included in `fms-hf-tuning` to perform this as a post checkpointing process preferably running offline without interrupting the main training process. 
2. Change can be included in `fms-acceleration` where it saves safe_tensors and HF reconstruction as part of DCP checkpointing on demand by introducing another flag. 

## Describe alternatives you've considered

Using the checkpoint utils script from `fms-acceleration` in `accelerated-moe` plugin. 

cc: @willmj @fabianlim

fabianlim:
@kmehant this is a trade off between convinience and complexity. I beleive why the new torch dcp does not support full checkpoints anymore, is because of the inefficiency of it. Adding a script to convert it to a single checkpoint is maybe helpful in some instances, but might be also harmful in others

dangxuanhong:
@fabianlim @kmehant I suppose this issue is related to this [one](https://github.ibm.com/ai-foundation/watson-fm-stack-tracker/issues/1640), right? 

If so, what will be our next step or solution? 

Per our experience: we run the manual conversion step and received: RuntimeWarning: 'fms_acceleration_moe.utils.checkpoint_utils' found in sys.modules after import of package 'fms_acceleration_moe.utils', but prior to execution of 'fms_acceleration_moe.utils.checkpoint_utils'; this may result in unpredictable behaviour)

And evaluating the converted model did not show any improved performance after the fine-tuning. So, it's not sure if the conversion had been conducted properly.

fabianlim:
@dangxuanhong I believe that the warning can be ignored. If the conversion had failed, then you should see your model produce rubbish outputs. If you try some sample inputs, and they look ok, then the conversion should have succeeeded, and the lack of performance thereof is related to other factors

PR #486, opened 2025-03-06
feat: support moe hf chkpt

kmehant:
### Description of the change

This is to support automatic HF checkpointing for moe EP training runs using fms-acceleration.

Depends on https://github.com/foundation-model-stack/fms-acceleration/pull/133

### Related issue number

#480

### DIFF
--- a/tuning/config/acceleration_configs/__init__.py
+++ b/tuning/config/acceleration_configs/__init__.py
@@ -15,6 +15,7 @@
 # Local
 from .acceleration_framework_config import AccelerationFrameworkConfig
 from .attention_and_distributed_packing import AttentionAndDistributedPackingConfig
+from .callbacks import get_additional_accel_framework_callbacks
 from .fast_moe import FastMoeConfig
 from .fused_ops_and_kernels import FusedOpsAndKernelsConfig
 from .quantized_lora_config import QuantizedLoraConfig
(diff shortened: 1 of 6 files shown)


dushyantbehl:
@kmehant can you fix the lint error

tuning/config/acceleration_configs/__init__.py
```diff
@@ -18,3 +18,11 @@
 from .fast_moe import FastMoeConfig
 from .fused_ops_and_kernels import FusedOpsAndKernelsConfig
 from .quantized_lora_config import QuantizedLoraConfig
+
+
+def get_more_accel_framework_callbacks(active_plugins, **kwargs):
```
dushyantbehl, line 23: nit: could you place this in `utils` or a different file?

also possibly rename to `get_additional_accel_framework_callbacks`

dushyantbehl (commented):
LGTM

willmj (commented):
Thanks Mehant! This looks good to me, but could we update the [notes section on Fast MoE](https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/README.md#fms-acceleration) to include that this conversion now happens automatically?

willmj:
Also if we could add a small unit test here (maybe similar to the [e2e tests](https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/tests/test_sft_trainer.py#L428) we have in test_sft_trainer) to make sure this conversion is happening correctly and inference can be run on the model, that would be good - but it may be hard to find a small moe model to test with.

kmehant:
@willmj addressed your comments, thaks

willmj (approved):
LGTM, thanks!

PR #487, opened 2025-03-06
feat: Add tokenizer data handler.

dushyantbehl:
### Description of the change

For most people truncation of dataset samples was not being respected hence allow a tokenizer handler for people to have direct control over dataset and truncate samples properly by calling out the handler in data config.

### Was the PR tested

### DIFF
--- a/tuning/data/data_handlers.py
+++ b/tuning/data/data_handlers.py
@@ -15,7 +15,7 @@
 # Definition of some predefined data preprocessing functions that we need.
 
 # Standard
-from typing import Dict, List
+from typing import Dict, List, Union
 import copy
 import re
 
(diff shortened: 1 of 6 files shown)


tuning/data/data_handlers.py
```diff
@@ -257,6 +257,37 @@ def apply_tokenizer_chat_template(
...
+        truncation: Truncation strategy to use, refer the link
+                    (https://huggingface.co/docs/transformers/en/pad_truncation)
+        max_length: Max length to truncate the samples to.
+        kwargs: Any additional kwargs that need to be passed to the tokenizer can be passed as
+                kwargs['tokenizer_kwargs']
+    Returns:
+        tokenized dataset elemenent field "dataset_text_field"
+    """
+    tokenizer_kwargs = kwargs.get("tokenizer_kwargs", {})
+    return tokenizer(
```
ashokponkumar, line 283: Will the tokenizer be initialized for every invocation? is there no way to create a tokenizer object before hand, and reuse it?

dushyantbehl: @ashokponkumar  this is the tokenizer call...we initialize it in the data preprocessor backend its an argument to this function...did you mean tokenizer_kwargs? if so I can make them as part of function args

ashokponkumar (approved):

dushyantbehl:
Merging on critical request. This is independent delta.

Abhishek-TAMU (commented):
Separate handler which just tokenizes input accepting any combination values of argument `truncation` and `padding` operating on `max_seq_len` value is definitely helpful. Thanks @dushyantbehl

Issue #450, opened 2025-01-25
bug: crash on trying to process data when `batched` is set to true


HarikrishnanBalagopal:
This line fails when `batched` is set to true: https://github.com/foundation-model-stack/fms-hf-tuning/blob/f22e2432b8ebbd449144a1546064e872f0888356/tuning/data/data_handlers.py#L96

```
ERROR:sft_trainer.py:multiprocess.pool.RemoteTraceback: 
"""
Traceback (most recent call last):
  File "/home/tuning/.local/lib/python3.11/site-packages/multiprocess/pool.py", line 125, in worker
    result = (True, func(*args, **kwds))
                    ^^^^^^^^^^^^^^^^^^^
  File "/home/tuning/.local/lib/python3.11/site-packages/datasets/utils/py_utils.py", line 678, in _write_generator_to_queue
    for i, result in enumerate(func(**kwargs)):
  File "/home/tuning/.local/lib/python3.11/site-packages/datasets/arrow_dataset.py", line 3558, in _map_single
    batch = apply_function_on_filtered_inputs(
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/tuning/.local/lib/python3.11/site-packages/datasets/arrow_dataset.py", line 3427, in apply_function_on_filtered_inputs
    processed_inputs = function(*fn_args, *additional_args, **fn_kwargs)
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/tuning/.local/lib/python3.11/site-packages/tuning/data/data_handlers.py", line 96, in apply_dataset_formatting
    f"{dataset_text_field}": element[f"{dataset_text_field}"] + tokenizer.eos_token
                             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~
TypeError: can only concatenate list (not "str") to list
"""

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/tuning/.local/lib/python3.11/site-packages/tuning/sft_trainer.py", line 650, in main
    trainer, additional_train_info = train(
                                     ^^^^^^
  File "/home/tuning/.local/lib/python3.11/site-packages/tuning/sft_trainer.py", line 317, in train
    ) = process_dataargs(
        ^^^^^^^^^^^^^^^^^
[...]

PR #488, opened 2025-03-06
feat: Update data handler backend and introduce filter based handlers.

dushyantbehl:
### Description of the change

For now we were supporting only MAP style data handlers but some use cases need support for FILTER type data handlers and to acoomodate this and any further type of handlers this PR adds flexibility to add new type of handlers in the data handler backed. 

Updated data handler backend.
Added explicit batching support. 
Added filter type handler.

### Related issue number

Also solves https://github.com/foundation-model-stack/fms-hf-tuning/issues/450

### Was the PR tested


tuning/data/data_handlers.py
```diff
@@ -322,12 +348,53 @@ def duplicate_columns(
     }
 
 
+def skip_large_text(element: Dict[str, str], column_name: str, max_length: int):
```
willmj, line 351: Add docstring here with explanation of function, args, and return

willmj: Also if `max_length` is referring to tokens or characters

willmj (commented):
Looks pretty much good to me, great testing and backend changes - new handler should include a docstring

tuning/data/data_handlers.py
dushyantbehl: thanks...I missed the docstring...added

willmj (approved):
LGTM

ashokponkumar (approved):

tuning/data/data_handlers.py
```diff
@@ -322,12 +348,66 @@ def duplicate_columns(
...
+                    else this can be characters for text column
+    Returns:
+        Filtered dataset which contains elements with column {column_name}
+                 having length shorter than {max_length}
+    """
+    if column_name not in element or max_length is None:
+        raise ValueError(
+            "Please provide correct column name and max_length to skip large columns"
+        )
+    return len(element[column_name]) < max_length
```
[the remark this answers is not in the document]
Abhishek-TAMU: Also curious if it should be `<= max_length`

tuning/data/data_handlers.py
```diff
@@ -28,6 +29,31 @@
...
 
+class DataHandlerType(Enum):
+    MAP = 1
+    FILTER = 2
+
+
+class DataHandler:
+    op: callable  # the actual handler function
+    handler_type: DataHandlerType  # either map or filter
+    allows_batching: bool  # supports batched mode or not
```
Abhishek-TAMU, line 40: nit: doc string suggestion instead of comments

```
class DataHandler:
"""
  A class to encapsulate a data processing handler.
  
  Attributes:
      op (callable): The actual handler function.
      handler_type (DataHandlerType): Indicates whether the handler is for mapping or filtering.
      allows_batching (bool): Flag to indicate if the handler supports batched operations.
"""
```

tuning/data/data_handlers.py
```diff
@@ -322,12 +348,66 @@ def duplicate_columns(
     }
 
 
+def skip_large_text(element: Dict[str, str], column_name: str, max_length: int):
```
Abhishek-TAMU, line 351: We can also plan to add Filter usage data handlers in [documentation here](https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/docs/advanced-data-preprocessing.md#preexisting-data-handlers) apart from usage of data handlers via MAP and some of the info could be also modified for example first para of [mentioned here](https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/docs/advanced-data-preprocessing.md#how-can-users-specify-data-handlers)

Abhishek-TAMU (commented):
@dushyantbehl This PR is great considering several additions and refactors. 
Didn't mean to slow down the merge process but have some inputs which could be handled Adhoc otherwise if required.

PR #491, opened 2025-03-07
feat: add sum op for trainer controller

kmehant:
### Description of the change

Allow sum ops in trainer controller control rules.


kmehant:
@seshapad can you review this?

seshapad (approved):

Issue #492, opened 2025-03-10
Bug in processing escaped chars in response_template


ChanderG:
## Describe the bug

One would use a response_template of the form "\n### Response:" normally. The first "\n" is included to encode the tokens correctly with context (see [here] (https://huggingface.co/docs/trl/en/sft_trainer#using-tokenids-directly-for-responsetemplate) for some explanation). In the [current code](https://github.com/foundation-model-stack/fms-hf-tuning/blob/c3aa25ccb2048b8d7c09acd30f85c19f83e28db8/tuning/data/data_preprocessing_utils.py#L67), the prefix tokens are stripped.

Some models like Llama have wierd tokenizer behaviour where the text tokenizes differently with a newline. For example:

```
>>> from transformers import AutoTokenizer
>>> tok_l = AutoTokenizer.from_pretrained("meta-llama/Llama-3.1-8B")
>>> msg = "###Input: This is an input. \n###Response:\n This is the output."
>>> tok_l(msg)
{'input_ids': [128000, 14711, 2566, 25, 1115, 374, 459, 1988, 13, 720, 14711, 2647, 512, 1115, 374, 279, 2612, 13], 'attention_mask': [1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1]}
>>> tok_l("\n###Response:")
{'input_ids': [128000, 198, 14711, 2647, 25], 'attention_mask': [1, 1, 1, 1, 1]}
>>> tok_l("\n###Response:\n")
{'input_ids': [128000, 198, 14711, 2647, 512], 'attention_mask': [1, 1, 1, 1, 1]}
```

This means, we need to use response_template with an extra "\n" at the end. In my testing, this was not a problem with Granite, but this is definitely needed with Llama.

But, using this new response template didn't solve the problem. Instead a new error showed up:
```
UserWarning: Could not find response key `[2647,  7338, 77]` in the following instance...
```

This was puzzling, since the expected response key is `[2647, 512]`. Decoding this new token set shows that the input text recieved is "Response:\\n". This is especially puzzling since just printing out the response_template at all points shows the single "\n", but in reality, internally the string stored is actually "\\n". Somehow this problem has happened only for the trailing "\n", perhaps because the first "\n" is anyway thrown away after tokenizing.

[...]

PR #493, opened 2025-03-10
fix: (data pre-process) un-escape templates and strings passed on cli correctly

ChanderG:
Fix for #492

### Related issue number

#492 

### Was the PR tested


ChanderG:
I have tested the change locally with the original dataset that caused the problem.

willmj (approved):
LGTM

PR #495, opened 2025-03-11
feat: Add tools and documents usage in chat template

dushyantbehl:
### Description of the change

Cetain chat templates like [Graite 3.1 Instruct](https://huggingface.co/ibm-granite/granite-3.1-8b-instruct/blob/main/tokenizer_config.json#L188) have different way to processing the element and need to process a sub field of the data element. 

Further they require usage of `tools` and `documents` which can be enabled via our existing API. 
This code performs just that and enables users to pass in additional data arguments to the data handler to enable processing sub field of a data sample.

### Was the PR tested


willmj (commented):
Change looks good to me, but can we add a unit test for this before merging?

dushyantbehl:
@willmj @Abhishek-TAMU added the test case as well.
Feel free to look at the PR again.

willmj (approved):
LGTM

Abhishek-TAMU (approved):
LGTM!

PR #497, opened 2025-03-13
fix:Avoid passing dataset_text_field in data arguments when already passed in data_config yaml

Abhishek-TAMU:
### Description of the change

Currently when we use `data_config` YAML file to apply `Data Handlers` which needs `fn_kwargs` as `dataset_text_field`, we also needed to pass the same `dataset_text_field` via command line arguments. 

Now with the change in this PR, if `dataset_text_field` passed via arguments is None, then it picks up the `dataset_text_field` passed in `data_config` YAML.

Other checks: If user passes multiple datasets in `data_config` YAML and each datasets have different `dataset_text_field` then it will give warning and pick up `dataset_text_field` of the first dataset.

Related Slack Discussion: https://ibm-research.slack.com/archives/C0698T0FW3S/p1740598883877809

### Related issue number

Issue: https://github.ibm.com/ai-foundation/watson-fm-stack-tracker/issues/1657

### Was the PR tested

### DIFF
--- a/tuning/data/setup_dataprocessor.py
+++ b/tuning/data/setup_dataprocessor.py
@@ -33,6 +33,7 @@ from tuning.data.data_config import (
 from tuning.data.data_handlers import DataHandler
 from tuning.data.data_preprocessing_utils import get_data_collator
 from tuning.data.data_processors import get_datapreprocessor
+from tuning.utils.utils import get_dataset_text_field_config
 
 logger = logging.getLogger(__name__)
 
@@ -97,7 +98,13 @@ def _process_dataconfig_file(
             )
     train_dataset = processor.process_dataset_configs(data_config.datasets)
 
-    return (train_dataset, None, data_args.dataset_text_field)
+    dataset_text_field = data_args.dataset_text_field
+    if dataset_text_field is None:
+        try:
+            dataset_text_field = get_dataset_text_field_config(data_config)
+        except KeyError:
+            logger.warning("dataset_text_field not found in data_config arguments")
+    return (train_dataset, None, dataset_text_field)
 
 
 # Data Format 1: Pretokenized Data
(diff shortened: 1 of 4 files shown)


tuning/data/setup_dataprocessor.py
```diff
@@ -97,7 +98,13 @@ def _process_dataconfig_file(
             )
     train_dataset = processor.process_dataset_configs(data_config.datasets)
 
-    return (train_dataset, None, data_args.dataset_text_field)
+    dataset_text_field = data_args.dataset_text_field
+    if dataset_text_field is None:
```
Abhishek-TAMU, line 102: Probable question @dushyantbehl :
Do we need this check `if data_args.dataset_text_field is None:` or should we overwrite data_args.dataset_text_field with what is passed in `data_config` as `dataset_text_field` and if not `dataset_text_field` is not passed in `data_config` then overwrite it to None ?

Abhishek-TAMU:
**Question**: Which place do you think documentation of this would be suitable to warn user of not passing multiple and different `dataset_text_field` for different datasets in `data_config.yaml` ? @dushyantbehl @willmj

tuning/data/setup_dataprocessor.py
willmj: Would be interested to hear Dushyant's opinion on this, in my opinion we should be overwriting `data_args.dataset_text_field` if `dataset_text_field` is set in the `data_config` in case someone accidentally is passing the wrong `dataset_text_field` in `data_args`

tuning/data/setup_dataprocessor.py
```diff
@@ -97,7 +98,13 @@ def _process_dataconfig_file(
...
+        try:
+            dataset_text_field = get_dataset_text_field_config(data_config)
+        except KeyError:
```
willmj, line 105: Can KeyError be raised? Or do we want something like (this would have to be changed if we decide to go with overwrite)
```suggestion
        if dataset_text_field is None:
```

willmj (commented):
Unit tests look good, couple comments

dushyantbehl:
@Abhishek-TAMU closing the PR as we will not take this change in.

PR #503, opened 2025-03-18
fix: save model dir hf moe checkpoint

willmj:
### Description of the change
Building on PR #486, saving `hf_converted_checkpoint` to `save_model_dir` if it exists.

### Was the PR tested

### DIFF
--- a/tuning/config/acceleration_configs/fast_moe.py
+++ b/tuning/config/acceleration_configs/fast_moe.py
@@ -126,7 +126,9 @@ def get_callbacks(**kwargs):
                     checkpoint(checkpoint_dir)
 
                     # If final save directory is provided, save the model there
-                    if getattr(self, "save_model_dir", None):
+                    if getattr(self, "save_model_dir", None) and state.global_step + 1 == state.max_steps:
+                        if not os.path.exists(self.save_model_dir):
+                            checkpoint(self.save_model_dir)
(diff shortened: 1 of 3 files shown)


tuning/config/acceleration_configs/fast_moe.py
```diff
@@ -98,29 +97,42 @@ def checkpoint():
...
+                    if (
+                        getattr(self, "save_model_dir", None)
+                        and state.global_step == state.max_steps
+                    ):
+                        if not os.path.exists(self.save_model_dir):
+                            os.mkdir(self.save_model_dir)
+                        checkpoint(checkpoint_dir, self.save_model_dir)
 
         callbacks.append(
             ConvertAndSaveHFCheckpointAtEverySave(
```
kmehant, line 138: Can we pass `save_model_dir` here and pick it up in `__init__` ?

kmehant (commented):
@willmj apart from one comment I have, rest all looks good to me. Can you also confirm if you are able to pass both the unit tests for this feature on your end? Since I dont think these would run in our GitHub Action runners.

tuning/config/acceleration_configs/fast_moe.py
willmj: Fixed, good catch!

PR #507, opened 2025-03-20
build(deps): install mamba_ssm from package instead of github

anhuong:
### Description of the change

Installing mamba_ssm from github blocks PyPi release due to error
```
[33mWARNING [0m Error during upload. Retry with the --verbose option for more details. 
2025-03-20T15:47:26.0222991Z [31mERROR   [0m HTTPError: 400 Bad Request from https://upload.pypi.org/legacy/        
          Can't have direct dependency: mamba_ssm[causal-conv1d]@                
          git+https://github.com/state-spaces/mamba.git ; extra == "mamba". See  
          https://packaging.python.org/specifications/core-metadata for more     
          information.      
```

Thus we are installing from package instead of github.

### Was the PR tested


aluu317 (approved):
Thank you!!!

PR #508, opened 2025-03-20
build(deps): install mamba_ssm from package instead of github

anhuong:
### Description of the change

Doing patch release of 2.7.1 with mamba dependency fix in order to publish pypi package.

### Was the PR tested


anhuong:
need to cherry-pick commit from main branch

willmj (approved):
LGTM

PR #513, opened 2025-03-26
feat: Add ALoRA support

kgreenewald:
### Description of the change

Support for activated lora. Right now points to alora code on a public repo https://github.com/IBM/activated-lora . Also uses a data collator by default due to the nature of alora, and adds a new argument for the invocation string. Also note that peft=0.14.0 seems to be necessary to avoid errors with the alora code.

### Related issue number

https://github.ibm.com/ai-foundation/watson-fm-stack-tracker/issues/1629

### Was the PR tested
I tested it with the following command
python tuning/sft_trainer.py \
--model_name_or_path ibm-granite/granite-3.1-2b-base \
--training_data_path /proj/dmfexp/statllm/users/kgreenewald/Thermometer/TuningStack/fms-hf-tuning-alora/tests/artifacts/testdata/json/twitter_complaints_input_output.json \
--output_dir ./output \
--num_train_epochs 1 \
--per_device_train_batch_size 4 \
---learning_rate 1e-4 \
--response_template "\n### Label:" \
--peft_method "alora" \
--save_strategy=no \
--r 8 \
--use_flash_attn false \
--lora_dropout 0.05 \
--lora_alpha 16 \
--invocation_string "\n### Label:" \
--target_modules q_proj k_proj v_proj

output:

{'loss': 0.0, 'grad_norm': 0.0, 'learning_rate': 0.0, 'epoch': 1.0}                                                         
{'train_runtime': 25.4343, 'train_samples_per_second': 1.966, 'train_steps_per_second': 0.511, 'train_loss': 0.0, 'epoch': 1.0}

### DIFF
--- a/README.md
+++ b/README.md
@@ -702,7 +702,7 @@ Activated LoRA (aLoRA) is a new low rank adapter architecture that allows for re
 
 [Github](https://github.com/IBM/activated-lora)
 
-*Usage* Usage is very similar to standard LoRA, with the key difference that an invocation_string must be specified so that the model knows when to turn on i.e "activate" the adapter weights. The model will scan any input strings (during training or at test time) for this invocation_string, and activate the adapter weights 1 token after the start of the sequence. If there are multiple instances of the invocation_string in the same input, it will activate at the last such instance.
+**Usage** Usage is very similar to standard LoRA, with the key difference that an invocation_string must be specified so that the model knows when to turn on i.e "activate" the adapter weights. The model will scan any input strings (during training or at test time) for this invocation_string, and activate the adapter weights 1 token after the start of the sequence. If there are multiple instances of the invocation_string in the same input, it will activate at the last such instance.
 
 **Note** Often (not always) aLoRA requires higher rank (r) than LoRA. r=32 can be a good starting point for challenging tasks.
 
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@
   - [Tips on Parameters to Set](#tips-on-parameters-to-set)
 - [Tuning Techniques](#tuning-techniques)
   - [LoRA Tuning Example](#lora-tuning-example)
-  - [Activated LoRA Tuning Example](#alora-tuning-example)
+  - [Activated LoRA Tuning Example](#activated-lora-tuning-example)
   - [GPTQ-LoRA with AutoGPTQ Tuning Example](#gptq-lora-with-autogptq-tuning-example)
   - [Fine Tuning](#fine-tuning)
   - [FMS Acceleration](#fms-acceleration)
--- a/README.md
+++ b/README.md
@@ -9,6 +9,7 @@
   - [Tips on Parameters to Set](#tips-on-parameters-to-set)
 - [Tuning Techniques](#tuning-techniques)
   - [LoRA Tuning Example](#lora-tuning-example)
+  - [Activated LoRA Tuning Example](#alora-tuning-example)
   - [GPTQ-LoRA with AutoGPTQ Tuning Example](#gptq-lora-with-autogptq-tuning-example)
   - [Fine Tuning](#fine-tuning)
   - [FMS Acceleration](#fms-acceleration)
--- a/README.md
+++ b/README.md
@@ -452,7 +452,7 @@ To summarize you can pick either python for single-GPU jobs or use accelerate la
 
 ### Tips on Parameters to Set
 
-#### Saving checkpoints while training
+#### Saving checkpoints while training (does not apply to Activated LoRA)
 
 By default, [`save_strategy`](tuning/config/configs.py) is set to `"epoch"` in the TrainingArguments. This means that checkpoints will be saved on each epoch. This can also be set to `"steps"` to save on every `"save_steps"` or `"no"` to not save any checkpoints.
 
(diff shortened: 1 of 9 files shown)


kgreenewald:
> @kgreenewald Thank you for bringing in this! Can you sign your commits? `git commit -s -m "<message goes here>"`

Certainly - is there a way to fix this retroactively, or should I just keep this in mind for the future?

kgreenewald:
> @kgreenewald Thank you for bringing in this! Can you sign your commits? `git commit -s -m "<message goes here>"`

I think I was retroactively able to sign these, let me know if it worked

kmehant:
@kgreenewald All commits from your commit lineage have signs except the recent once, you might want to amend the commit and also sign it. So that DCO passes. Thanks

kgreenewald:
> @kgreenewald All commits from your commit lineage have signs except the recent once, you might want to amend the commit and also sign it. So that DCO passes. Thanks

Thanks. Fixed, I think.

kgreenewald:
> @kgreenewald the commit [4d45f57](https://github.com/foundation-model-stack/fms-hf-tuning/commit/4d45f5713c7d7660461b9c73b273b42e33a6b9c0) is not signed :)

Not sure what happened since it said it was signed on my end, but rerunning rebase seems to have fixed & DCO is now passed.

error.log
```diff
@@ -0,0 +1 @@
+Exception raised during training. This may be a problem with your input: name 'aLoRAConfig' is not definedUnhandled exception during training: FlashAttention2 has been toggled on, but it cannot be used due to the following error: the package flash_attn seems to be not installed. Please refer to the documentation of https://huggingface.co/docs/transformers/perf_infer_gpu_one#flashattention-2 to install Flash Attention 2.Unhandled exception during training: name 'aLoRAConfig' is not defined
```
kmehant, line 1: Shall we remove this file?

tuning/config/peft_config.py
```diff
@@ -58,6 +58,50 @@ class LoraConfig:
...
+            completely match or "
+            'end with one of the strings. If the value is ["all-linear"], \
+            then LORA selects all linear and Conv1D '
+            "modules except for the output layer."
+        },
+    )
+    bias = "none"
+    lora_dropout: float = 0.05
+    activation_prompt: str = None
+
```
kmehant, line 104: Can we reuse LoraConfig dataclass which is in this file instead of creating a new dataclass for aLoRA? You can simply add activation_prompt parameter to LoraConfig dataclass. WDYT?

error.log
kgreenewald: removed!

tuning/config/peft_config.py
kgreenewald: The shared parent class for most of these is PeftConfig, which is a PEFT class

kmehant: Thanks for letting me know, however, if there are common attributes with some other peft dataclass, they seem to subclass from that class. E.g. AdaLoraConfig. https://github.com/huggingface/peft/blob/8c8b529b3153596514b319c834ad3aa32274236a/src/peft/tuners/adalora/config.py#L24

Should we also take similar approach in your case?

tuning/sft_trainer.py
```diff
@@ -37,8 +37,10 @@
 )
 from transformers.trainer_utils import get_last_checkpoint
 from transformers.utils import is_accelerate_available
-from trl import SFTConfig, SFTTrainer
+from trl import SFTConfig, SFTTrainer, DataCollatorForCompletionOnlyLM
 import transformers
+from alora.peft_model_alora import aLoRAPeftModelForCausalLM
+from alora.config import aLoraConfig
```
kmehant, line 43: I think you might want to import `tuning.config` where you defined your dataclass isn't it?

tuning/sft_trainer.py
```diff
@@ -375,16 +377,32 @@ def train(
...
+        trainer = SFTTrainer(
+            model=model,
+            tokenizer=tokenizer,
+            train_dataset=formatted_train_dataset,
+            eval_dataset=formatted_validation_dataset,
+            data_collator=data_collator,
+            args=training_args,
+            callbacks=trainer_callbacks,
+            peft_config=peft_config,
+        )
```
kmehant, line 405:
```suggestion
    if isinstance(peft_config, aLoraConfig):
        tokenizer.padding_side='right'
        response_token_ids = (tokenizer(peft_config.invocation_string, return_tensors="pt", add_special_tokens=False))['input_ids']
        model = aLoRAPeftModelForCausalLM(model, peft_config, response_token_ids = response_token_ids)
        data_collator = DataCollatorForCompletionOnlyLM(invocation_string, tokenizer=tokenizer)
        peft_config = None
    trainer = SFTTrainer(
        model=model,
        tokenizer=tokenizer,
        train_dataset=formatted_train_dataset,
        eval_dataset=formatted_validation_dataset,
        data_collator=data_collator,
        args=training_args,
        callbacks=trainer_callbacks,
        peft_config=peft_config,
    )
```

How about this?

tuning/sft_trainer.py
```diff
@@ -375,16 +377,32 @@ def train(
...
-        train_dataset=formatted_train_dataset,
-        eval_dataset=formatted_validation_dataset,
-        data_collator=data_collator,
-        args=training_args,
-        callbacks=trainer_callbacks,
-        peft_config=peft_config,
-    )
+    # activated LoRA
+    if isinstance(peft_config, aLoraConfig):
+        tokenizer.padding_side='right'
```
kmehant, line 382: Can we log here that we modifying the padding_side to right?

tuning/sft_trainer.py
kgreenewald: I guess this is a question for you - is there a reason that you have your own LoraConfig in tuning.config, as opposed to the one PEFT has (peft.LoraConfig)? 

I've updated my repo to subclass peft.LoraConfig, I could also subclass your tuning.config.LoraConfig if that makes sense. Alternately, I could just usethe aLoraConfig from my repo (as is currently implemented).

tuning/config/peft_config.py
kgreenewald: subclass - sounds good. I've updated my repo to do this, see my other comment below re this repo.

tuning/sft_trainer.py
kgreenewald: If you don't do this, you will actually get a warning I think? It is best practice to always use         tokenizer.padding_side='right' for training loras, and then go back to the default tokenizer.padding_side='left' at test time.

In other words, since the tokenizer is getting loaded in from the file (not an argument) and is not returned to anything, is there a reason to log this?

pyproject.toml
```diff
@@ -35,10 +35,11 @@ dependencies = [
...
 "trl>=0.13,<0.15",
-"peft>=0.8.0,<0.14",
+"peft>=0.8.0,<=0.14",
 "protobuf>=5.28.0,<6.0.0",
 "datasets>=2.15.0,<4.0",
 "simpleeval>=0.9.13,<2.0",
+"alora @ git+ssh://git@github.com/IBM/activated-lora.git@main",
 ]
 
 [project.optional-dependencies]
```
kmehant, line 45: Can we make alora an optional dependency? This would help us custom include it in the builds based on the user needs?

tuning/sft_trainer.py
kmehant: > I've updated my repo to subclass peft.LoraConfig, I could also subclass your tuning.config.LoraConfig if that makes sense. Alternately, I could just usethe aLoraConfig from my repo (as is currently implemented).


Thank you, I am okay on using the `aLoraConfig` from your repo

tuning/sft_trainer.py
kmehant: > If you don't do this, you will actually get a warning I think? It is best practice to always use tokenizer.padding_side='right' for training loras, and then go back to the default tokenizer.padding_side='left' at test time.


Thank you for letting me know. It would be good to log specific hardcoded actions in the code, this would lead to better transparency to the user. Something like

```
logging.info("tokenizer's padding side is explicitly modified to 'right' for aLoRA usage. May some addtional reasoning can go here...")
```

tuning/sft_trainer.py
```diff
@@ -586,9 +606,14 @@ def parse_arguments(parser, json_config=None):
 
         peft_method = additional.peft_method
         exp_metadata = additional.exp_metadata
-
+        invocation_string = additional.invocation_string
+        if peft_method == "alora":  
+            if invocation_string is None:
+                error("invocation_string needed for aLoRA")
```
kmehant, line 612: Is `error()` a inbuilt Python function? Can we use `raise SomeExceptionClass("")` instead? In this case, it can be 
```
raise ValueError("invocation_string is not passed required for aLoRA usage")
```

tuning/sft_trainer.py
kmehant, line 43: Since we want to ship it as an optional dependency, can we import these conditionally only if peft_method is alora?

kmehant (commented):
@kgreenewald Thanks a lot for the work and patience in composing this PR.

Added some final set of comments for you. I will hand off to one of tuning squad to have an another look on your PR.

Meanwhile, do you have cycles to add 1 unit test? You can copy this test case - https://github.com/foundation-model-stack/fms-hf-tuning/blob/dc77c6360f82a934441527ba4ea5e2fd1562ef23/tests/test_sft_trainer.py#L669 and modify it for aLoRA. You can add that to the same file.

pyproject.toml
```diff
@@ -35,10 +35,11 @@ dependencies = [
 "tokenizers>=0.13.3,<1.0",
 "tqdm>=4.66.2,<5.0",
 "trl>=0.13,<0.15",
-"peft>=0.8.0,<0.14",
+"peft>=0.8.0,<=0.14",
 "protobuf>=5.28.0,<6.0.0",
 "datasets>=2.15.0,<4.0",
 "simpleeval>=0.9.13,<2.0",
+"alora @ git+ssh://git@github.com/IBM/activated-lora.git@main",
```
HarikrishnanBalagopal, line 42: Please don't use the `main` branch. Pin the version using a branch/release/tag, etc.

anhuong: If you continue to use a github repo, this will fail to push to PyPi on the next release of fms-hf-tuning. Pypi doesn't allow using github repos as dependencies

anhuong:
I'm not sure why unit tests aren't running on this PR, we want them to run. Previously we have seen some unit tests fail with peft 0.14 as seen here https://github.com/foundation-model-stack/fms-hf-tuning/actions/runs/13978920484/job/39139509875?pr=505

tuning/sft_trainer.py
kgreenewald: In further experimentation, its not necessary to have paddingside=right for aLoRA, so I've actually removed this line.

tuning/sft_trainer.py
kgreenewald: Incorporated

pyproject.toml
kgreenewald: Putting as optional under name 
 activated-lora = ...

tuning/sft_trainer.py
kgreenewald: I believe it should work, I set it up with an if statement here and putting the alora import inside "try:" logic elsewhere.

pyproject.toml
kgreenewald: Made a special branch for this, for now. @anhuong  Going forward - we will work to directly contribute it to PEFT, failing that, I assume getting that into pip is the way  (never done this)?

tuning/sft_trainer.py
kgreenewald: Done

kgreenewald:
> @kgreenewald Thanks a lot for the work and patience in composing this PR 🎉. Can you rebase your PR and fix the conflicts?
> 
> Added some final set of comments for you. I will hand off to one of tuning squad to have an another look on your PR.
> 
> Meanwhile, do you have cycles to add 1 unit test? You can copy this test case -
> 
> https://github.com/foundation-model-stack/fms-hf-tuning/blob/dc77c6360f82a934441527ba4ea5e2fd1562ef23/tests/test_sft_trainer.py#L669
> and modify it for aLoRA. You can add that to the same file.

Thanks @kmehant ! I've gone through everything and added this unit test. updating the actual PR next

kgreenewald:
@kmehant @anhuong I've updated everything, let me know if things look good from your end. Thanks!

tests/test_sft_trainer.py
```diff
@@ -695,6 +717,44 @@ def test_run_causallm_lora_and_inference(request, target_modules, expected):
         assert len(output_inference) > 0
         assert "Simply put, the theory of relativity states that" in output_inference
```
kmehant, line 719: @kgreenewald Thanks for adding the unit test. You would need to decorate it with a condition to run only when alora package is locally available. 
1. Define a utility function. example. https://github.com/foundation-model-stack/fms-hf-tuning/blob/b3744ca0d5031c6e3a406190e0de8bdbe0a6900f/tuning/utils/import_utils.py#L22
2. Decore the unit test to run only if package available. example. https://github.com/foundation-model-stack/fms-hf-tuning/blob/b3744ca0d5031c6e3a406190e0de8bdbe0a6900f/tests/test_sft_trainer.py#L1453-L1456

kmehant: Apart from this, I vouch unit tests pass at my end as well, thank you!

tuning/utils/config_utils.py
```diff
@@ -79,7 +89,23 @@ def get_hf_peft_config(task_type, tuning_config, tokenizer_name_or_path):
...
+        from alora.config import aLoraConfig
+        if isinstance(tuning_config, aLoraConfig):
+            USE_ALORA = True
+    except ImportError:
+        pass
+    if USE_ALORA:
+        #alora_config = asdict(tuning_config)
+        #if alora_config["target_modules"] == ["all-linear"]:
+        #    alora_config["target_modules"] = "all-linear"
+        #hf_peft_config = aLoraConfig(**alora_config)
```
kmehant, line 103: Shall we clean up these comments?

tuning/sft_trainer.py
```diff
@@ -126,6 +127,20 @@ def train(
...
+        from alora.peft_model_alora import aLoRAPeftModelForCausalLM
+        from alora.config import aLoraConfig
+        if isinstance(peft_config, aLoraConfig):
+            USE_ALORA = True
+            if train_args.save_strategy!="no": 
+                logger.warning("Setting train_args.save_strategy to 'no' for aLoRA. Model will be saved at end of training.")
+                ALORA_SAVE_END = True
+                train_args.save_strategy = "no"
+            else:
+                ALORA_SAVE_END = False
```
kmehant, line 141: Apologies, if I have missed rationale behind this, why we wish to save when save_strategy is explicitly set to "no"? Specifically for aLoRA?

tuning/sft_trainer.py
```diff
@@ -558,6 +603,14 @@ def parse_arguments(parser, json_config=None):
...
             Extra tracker metadata.
     """
+    if peft_method == "alora": 
+        try:
+            from alora.config import aLoraConfig
+        except ImportError:
+            raise ImportError(
+                "The alora package is required for this operation. "
+                "Please install it from https://github.com/IBM/activated-lora."
+            )
```
kmehant, line 613: This is too early, since, `peft_method` variable is not yet materialized at this point of the code. You might want to add it anywhere after line `631`

tuning/sft_trainer.py
kgreenewald: It is actually the opposite -- if the save_strategy is *not* = no, we want to save things at the end since we aren't able to save in the loop. SFTTrainer isn't able to save aLoRA if you don't set save_strategy=no in the argument. In the docs, we'll say you should use save_strategy='no' for aLoRA.

tuning/sft_trainer.py
kgreenewald: Thanks!

tuning/utils/config_utils.py
kgreenewald: yes

kgreenewald (commented):
updated! See comments

pyproject.toml
anhuong: Sorry for missing this comment, yes getting the package into pypi so we can install via pip would be the best way forward - https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/

kgreenewald: @anhuong @kmehant it is in pypi now, updated here to do pip install alora

kgreenewald:
> @kgreenewald waiting for the tests to pass, can we fix the failing tests? failing tests seem to be related to alora. Thank you.
@kmehant 
Fixed, alora tests now all working on my end. I initially also saw failing tests due to peft=0.14 upgrade. They seem to have all stemmed from a single line (padding=False) in the data processor file. As of peft=0.14, they are much more strict about making sure tensors line up so this was throwing errors. For peft=0.13, in situations like this, padding would happen anyway, but in 0.14 it throws an error. Changing this one line to True seems to have fixed.

kmehant (approved):
Thanks for this feature.

scripts/run_inference.py
```diff
@@ -279,12 +307,19 @@ def run(
...
-        )
+        if alora_offsets is None:
+            peft_outputs = self.peft_model.generate(
+                input_ids=input_ids, max_new_tokens=max_new_tokens
+            )
+        else:  # pass in alora_offsets needed for alora model
+            peft_outputs = self.peft_model.generate(
+                input_ids=input_ids,
+                max_new_tokens=max_new_tokens,
+                alora_offsets=alora_offsets,
```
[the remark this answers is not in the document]
kgreenewald: This is an optional argument to the method "run" which this is inside. For aLoRA inference, the user always needs to pass in this alora_offsets with the inputs (as done in the test & consistent with the PEFT implementation). I've added a comment here.

Ive also added an inference example to the README.

tuning/utils/import_utils.py
```diff
@@ -32,3 +32,18 @@ def is_fms_accelerate_available(
...
+
+
+def is_alora_available(
+    plugins: Union[str, List[str]] = None, package_name: str = "alora"
+):
+    names = [package_name]
+    if plugins is not None:
+        if isinstance(plugins, str):
+            plugins = [plugins]
+        names.extend([package_name + "_" + x for x in plugins])
```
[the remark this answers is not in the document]
kgreenewald: Done!

tuning/sft_trainer.py
```diff
@@ -126,8 +127,28 @@ def train(
...
+            USE_ALORA = True
+            if train_args.save_strategy != "no":
+                logger.warning(
+                    "Setting train_args.save_strategy to 'no' for aLoRA."
+                    "Model will be saved at end of training."
+                )
+                ALORA_SAVE_END = True
+                train_args.save_strategy = "no"
+            else:
+                ALORA_SAVE_END = False
```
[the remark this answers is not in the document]
kgreenewald: Done

kgreenewald:
@anhuong @kmehant Requested changes implemented, thanks!

anhuong:
@kgreenewald Please delete the `mykey.asc` and `mypubkey.asc` files.

README.md
```diff
@@ -700,6 +701,191 @@ post_process_vLLM_adapters_new_tokens(
...
+    "lora_alpha": 16,
+    "target_modules": ["q_proj", "k_proj", "v_proj"]
+}
+```
+
+Notice the `target_modules` are the names of the modules to apply the adapter to.
+- If this is specified, only the modules with the specified names will be replaced. When passing a list of strings, either an exact match will be performed or it is checked if the name of the module ends with any of the passed strings. If this is specified as `all-linear`, then all linear/Conv1D modules are chosen, excluding the output layer. 
+- If this is not specified, modules will be chosen according to the model architecture. If the architecture is not known, an error will be raised — in this case, you should specify the target modules manually. See [HuggingFace docs](https://huggingface.co/docs/peft/en/package_reference/lora#peft.LoraConfig) for more details.
+
+#### How to get list of aLoRA target_modules of a model
```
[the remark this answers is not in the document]
kgreenewald: Done

kgreenewald:
@anhuong Re the inference flag stuff - I now understand your point! I was able to restructure to move the "alora_offsets" logic inside of .run , so now only need "use_alora" when loading the model. At .run, it then retrieves this flag from the model object.

anhuong (approved):
Thank you so much for this feature Kristjan!

PR #520, opened 2025-04-07
feat: [experimental] enable sum loss trainer

dushyantbehl:
### Description of the change

Try to enable loss reduction to sum with hf trainer.

### Was the PR tested

### DIFF
--- a/tuning/config/configs.py
+++ b/tuning/config/configs.py
@@ -211,7 +211,10 @@ class TrainingArguments(transformers.TrainingArguments):
     enable_reduce_loss_sum: bool = field(
         default=False,
         metadata={
-            "help": "This function enables reduce_loss = sum in OpenInstruct style."
+            "help": "Pass `True` to enable use of sum loss reduction on the loss function. \
+                Please note this feature is experimental and not fully supported. \
+                One Known limitation of this function is PEFT PT so its disabled \
+                for all PEFT runs by the library internally."
         },
     )
 
--- a/tuning/config/configs.py
+++ b/tuning/config/configs.py
@@ -208,6 +208,12 @@ class TrainingArguments(transformers.TrainingArguments):
             Other possible values are 'debug', 'info', 'warning', 'error' and 'critical'"
         },
     )
+    enable_reduce_loss_sum: bool = field(
+        default=False,
+        metadata={
+            "help": "This function enables reduce_loss = sum in OpenInstruct style."
+        },
+    )
 
 
 @dataclass
--- a/tuning/sft_trainer.py
+++ b/tuning/sft_trainer.py
@@ -360,7 +360,7 @@ def train(
     # Note this check has to be done post trainer initialization
     if train_args.enable_reduce_loss_sum and _is_peft_model(trainer.model):
         raise ValueError(
-            "❌ Loaded model is PEFT model but both PEFT and sum loss should not be set together."
+            "❌ Loaded model is PEFT model but both PEFT and sum loss is not supported yet."
             + "Set --enable_reduce_loss_sum to false and retry"
         )
 
--- a/tuning/sft_trainer.py
+++ b/tuning/sft_trainer.py
@@ -53,6 +53,7 @@ from tuning.data.data_handlers import DataHandler
 from tuning.data.setup_dataprocessor import process_dataargs
 from tuning.trackers.tracker_factory import FILE_LOGGING_TRACKER, get_tracker
 from tuning.trainercontroller import TrainerControllerCallback
+from tuning.trainers.sum_loss_sft_trainer import SumLossSFTTrainer
 from tuning.utils.config_utils import get_hf_peft_config, get_json_config
 from tuning.utils.data_type_utils import get_torch_dtype
 from tuning.utils.error_logging import (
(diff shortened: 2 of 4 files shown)


tuning/config/configs.py
```diff
@@ -208,6 +208,17 @@ class TrainingArguments(transformers.TrainingArguments):
...
     )
+    enable_reduce_loss_sum: bool = field(
+        default=False,
+        metadata={
+            "help": "Pass `True` to enable use of sum loss reduction on the loss function. \
+                Switches out the compute_loss in SFTTrainer with that from internal one in\
+                tuning.trainers.sum_loss_sft_trainer::SumLossSFTTrainer. \
+                Please note this feature is experimental and not fully supported. \
+                One Known limitation of this function is PEFT PT so its disabled \
+                for all PEFT runs by the library internally."
```
ashokponkumar, line 219:
```suggestion
            "help": "Pass `True` to enable use of sum loss reduction on the loss function. \
                Please note this feature is experimental and not fully supported. \
                PEFT is not supported with this flag yet."
```

tuning/sft_trainer.py
```diff
@@ -339,16 +341,40 @@ def train(
...
-        train_dataset=formatted_train_dataset,
-        eval_dataset=formatted_validation_dataset,
-        data_collator=data_collator,
-        args=training_args,
-        callbacks=trainer_callbacks,
-        peft_config=peft_config,
-    )
+    trainer = None
+    if train_args.enable_reduce_loss_sum:
+        trainer = SumLossSFTTrainer(
```
ashokponkumar, line 346: If the parameters are same across both, would it make sense to have a variable to capture the class to use, and do the initialisation call only once?

tuning/sft_trainer.py
```diff
@@ -339,16 +341,40 @@ def train(
...
+            eval_dataset=formatted_validation_dataset,
+            data_collator=data_collator,
+            args=training_args,
+            callbacks=trainer_callbacks,
+            peft_config=peft_config,
+        )
+
+        # Note this check has to be done post trainer initialization
+        if _is_peft_model(trainer.model):
+            logger.warning(
```
ashokponkumar, line 359: Shouldn't we rather fail if both enable_reduce_loss_sum and peft flags are enabled?

kmehant (commented):
I am skeptical in introducing a new class, 

I wanted to make sure if we have seen these two flows

Causal Loss flow in transformers

1. https://github.com/huggingface/transformers/blob/08e3217bafddc5d11ce0e7369bcfaaabe5501ba5/src/transformers/loss/loss_utils.py#L41
2. https://github.com/huggingface/transformers/blob/08e3217bafddc5d11ce0e7369bcfaaabe5501ba5/src/transformers/loss/loss_utils.py#L27-L38 (loss reduction is chosen based on num_items_in_batch)

Custom loss over outputs and labels after forward pass through `compute_loss_func`. You might want to see the flow of self.compute_loss_func in https://github.com/huggingface/transformers/blob/08e3217bafddc5d11ce0e7369bcfaaabe5501ba5/src/transformers/trainer.py#L3778 (loss reduction can be applied in a custom func and passed to existing trainer)

Let me know.

dushyantbehl:
> I am skeptical in introducing a new class,

@kmehant what is the skepticism behind it can you elaborate

> I wanted to make sure if we have seen these two flows
> 
> Causal Loss flow in transformers
> 
> 1. https://github.com/huggingface/transformers/blob/08e3217bafddc5d11ce0e7369bcfaaabe5501ba5/src/transformers/loss/loss_utils.py#L41
> 2. https://github.com/huggingface/transformers/blob/08e3217bafddc5d11ce0e7369bcfaaabe5501ba5/src/transformers/loss/loss_utils.py#L27-L38

I am aware of the trainer loss flow and if you see [this](https://github.com/huggingface/transformers/blob/08e3217bafddc5d11ce0e7369bcfaaabe5501ba5/src/transformers/loss/loss_utils.py#L27-L38) the function alread performs an averaging over `num_items_in_batch`

> Custom loss over outputs and labels after forward pass through `compute_loss_func`. You might want to see the flow of self.compute_loss_func in https://github.com/huggingface/transformers/blob/08e3217bafddc5d11ce0e7369bcfaaabe5501ba5/src/transformers/trainer.py#L3778
> 

Also custom `compute_loss_func` doesn't work here because when computing full loss outside requires you to use state like `model embedding` size and was not suitable for when I tried it. 

Maybe we could revisit the flow at later point in time but did not seem like a HARD problem to me because [HF themselves call out ](https://github.com/huggingface/transformers/blob/08e3217bafddc5d11ce0e7369bcfaaabe5501ba5/src/transformers/trainer.py#L3782)that you should

`Subclass and override for custom behavior`

tuning/sft_trainer.py
dushyantbehl: its up to us to decide weather to fail or continue with a user message..if we decide to fail the above flow makes more sense to me.

https://github.com/foundation-model-stack/fms-hf-tuning/pull/520#discussion_r2034832245

tuning/sft_trainer.py
dushyantbehl: you would have to initialize a class anyway to know if the model is PEFT or not 
https://github.com/foundation-model-stack/fms-hf-tuning/pull/520/files#diff-7d8f6f8626340337d03a2f04fec2840d2ede32b0c7af2640eed9759af45bb3cfR357

tuning/sft_trainer.py
```diff
@@ -350,6 +357,13 @@ def train(
         peft_config=peft_config,
     )
 
+    # Note this check has to be done post trainer initialization
+    if train_args.enable_reduce_loss_sum and _is_peft_model(trainer.model):
+        raise ValueError(
+            "❌ Loaded model is PEFT model but both PEFT and sum loss should not be set together."
```
ashokponkumar, line 363:
```suggestion
            "❌ Loaded model is PEFT model but both PEFT and sum loss is not supported yet."
```

ashokponkumar (dismissed):

PR #523, opened 2025-04-08
feat: Enable LoRA checkpoint utils for ScatterMoE

willmj:
### Was the PR tested

### DIFF
--- a/build/accelerate_launch.py
+++ b/build/accelerate_launch.py
@@ -145,13 +145,16 @@ def main():
                 post_process_vLLM_adapters_new_tokens(
                     save_model_dir, save_model_dir, num_added_tokens
                 )
-                hf_converted_checkpoint = os.path.join(save_model_dir, "hf_converted_checkpoint")
-                if os.path.exists(
-                    os.path.join(hf_converted_checkpoint, "adapter_model.safetensors")
-                ):
-                    post_process_vLLM_adapters_new_tokens(
-                        hf_converted_checkpoint, hf_converted_checkpoint, num_added_tokens
-                    )
+
+            hf_converted_checkpoint = os.path.join(
+                save_model_dir, "hf_converted_checkpoint"
+            )
+            if os.path.exists(
+                os.path.join(hf_converted_checkpoint, "adapter_model.safetensors")
+            ):
+                post_process_vLLM_adapters_new_tokens(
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -902,7 +902,13 @@ Notes:
...
           - if True, it is Scatter MoE Kernels with experts sharded based on the top level sharding protocol (e.g. FSDP).
           - if False, Scatter MoE Kernels with complete replication of experts across ranks.
-    - `world_size` must be divisible by the `ep_degree`
+      - lora tuning with ScatterMoE is supported, but because of inference restrictions on vLLM/vanilla PEFT, experts should not be trained as `target_modules` for models being tuned with ScatterMoE. Users have control over which `target_modules` they wish to train:
+          - Passing `all-linear` to adapter layers will include the router, which is a linear layer, and all attn layers. This **will not** train the expert layers.
+          - To train only attention layers, specify target modules specifically (i.e `target_modules: ["q_proj", "v_proj", "o_proj", "k_proj"]`).
+          - To train expert layers, specify `input_linear` and `output_linear` in target modules along with `router` (i.e `target_modules: ["q_proj", "v_proj", "o_proj", "k_proj", "router", "input_linear", "output_linear"]`). If you specify these layers, inference with vLLM/vanilla HF PEFT **is not possible**.
```
kmehant, line 908:
```suggestion
          - To train expert layers, specify `input_linear` and `output_linear` in target modules along with `router` (i.e `target_modules: ["q_proj", "v_proj", "o_proj", "k_proj", "router", "input_linear", "output_linear"]`). If you specify these layers, inference with vLLM/vanilla HF PEFT **is not currently supported.**.
```

README.md
```diff
@@ -902,7 +902,13 @@ Notes:
...
+      - When lora tuning with ScatterMoE, the values `--fast_moe 1` or `--fast_moe True` are not expected to work, as FSDP must be enabled when lora tuning. Run either `--fast_moe False` or `--fast-moe x>1`.
```
kmehant, line 909: Didnt get your point quite yet here. `--fast_moe True` disables expert parallel however, experts are sharded by FSDP.  So FSDP is active here.

BTW, `--fast_moe 1` and `--fast_moe False` Both have the same effect isn't it? In both the settings, all experts are replicated and deferred from FSDP however, other layers are under FSDP sharding. 

May be if you are confortable with a support matrix table, lets do that and pin point case by case.

PR #526, opened 2025-04-10
fix: multipack and streaming are incompatible

willmj:
### Description of the change
Raises an error if multipack is passed with streaming

### Related issue number
[Context](https://github.com/foundation-model-stack/fms-acceleration/issues/87#issuecomment-2382869825)

### Was the PR tested

### DIFF
--- a/tests/data/test_data_preprocessing.py
+++ b/tests/data/test_data_preprocessing.py
@@ -831,6 +831,56 @@ def test_process_dataconfig_file_with_streaming_no_max_steps_errors(
     with pytest.raises(ValueError):
         (train_set, _, _) = _process_dataconfig_file(data_args, TRAIN_ARGS, tokenizer)
 
+@pytest.mark.parametrize(
+    "data_config_path, data_path",
+    [
+        (
+            DATA_CONFIG_YAML_STREAMING_INPUT_OUTPUT,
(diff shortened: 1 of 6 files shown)


tests/data/test_data_preprocessing.py
```diff
@@ -832,6 +833,67 @@ def test_process_dataconfig_file_with_streaming_no_max_steps_errors(
...
+@pytest.mark.parametrize(
+    "data_config_path, data_path",
+    [
+        (
+            DATA_CONFIG_YAML_STREAMING_INPUT_OUTPUT,
+            TWITTER_COMPLAINTS_DATA_INPUT_OUTPUT_JSON,
+        ),
+    ],
+)
+def test_process_dataconfig_file_with_streaming_and_multipack_error(
```
dushyantbehl, line 845:
```suggestion
def test_process_dataconfig_file_with_streaming_and_multipack_throws_error(
```

tuning/sft_trainer.py
```diff
@@ -306,6 +308,7 @@ def train(
         tokenizer,
         train_args,
         additional_data_handlers,
+        is_multipack=is_multipack,
         is_padding_free=is_padding_free,
```
dushyantbehl, line 312: do we have any test which use padding free? will they be affected since you added before padding free? 
could you move this below padding free in that case..

dushyantbehl (approved):
LGTM

PR #528, opened 2025-04-14
fix: change the default model_name_or_path from "facebook/opt-125m" to None

HarikrishnanBalagopal:
### Description of the change

Change the default model_name_or_path from "facebook/opt-125m" to None

This is especially relevant when using the offline dataset processing script where it's easy to forget to specify the model_name_or_path

### Was the PR tested

### DIFF
--- a/tuning/config/configs.py
+++ b/tuning/config/configs.py
@@ -35,7 +35,7 @@ DEFAULT_UNK_TOKEN = "<unk>"
 
 @dataclass
 class ModelArguments:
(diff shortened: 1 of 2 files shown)


kmehant:
@HarikrishnanBalagopal Can you raise a ValueError in train function in `sft_trainer.py` file if model_args.model_name_or_path is None?

kmehant (approved):

dushyantbehl (approved):
LGTM

PR #530, opened 2025-04-17
feat: Enable LoRA saving only for non MoE linear layers training with kernels. 

willmj:
### Description of the change
A more limited version of #523 to enable LoRA but explicitly block off router linear and expert layers.

### How to verify the PR
image
image

### Was the PR tested


README.md
```diff
@@ -902,6 +902,9 @@ Notes:
       - When a boolean is passed, the expert parallel degree defaults to 1 and further the behaviour would be as follows:
           - if True, it is Scatter MoE Kernels with experts sharded based on the top level sharding protocol (e.g. FSDP).
           - if False, Scatter MoE Kernels with complete replication of experts across ranks.
+      - lora tuning with ScatterMoE is supported, but because of inference restrictions on vLLM/vanilla PEFT, the expert layers and router linear layer should not be trained as `target_modules` for models being tuned with ScatterMoE. Users have control over which `target_modules` they wish to train:
+          - At this time, only attention layers are trainable when using LoRA with scatterMoE. Until support for the router linear layer is added in, target modules must be specified explicitly (i.e `target_modules: ["q_proj", "v_proj", "o_proj", "k_proj"]`) instead of passing `target_modules: ["all-linear"]`.
+      - FSDP must be used when lora tuning with `--fast_moe`
```
kmehant, line 907: I think this is true even for non lora usecase when moe kernels / EP is in use. So lets move this point above.

build/accelerate_launch.py
```diff
@@ -159,11 +170,30 @@ def main():
...
+                        if os.path.exists(
+                            os.path.join(
+                                hf_converted_checkpoint, "adapter_model.safetensors"
+                            )
+                        ):
+                            post_process_vLLM_adapters_new_tokens(
+                                hf_converted_checkpoint,
+                                hf_converted_checkpoint,
+                                num_added_tokens,
+                            )
```
kmehant, line 196: We can simplify this code as in you can prepare checkpoint_dir with  `output_dir, name` or `output_dir, name, hf_converted_checkpoint` by checking if `hf_converted_checkpoint` exists (`os.path.exists`) and then move to below code with the prepared checkpoint_dir

```
                        if os.path.exists(
                            os.path.join(checkpoint_dir, "adapter_model.safetensors")
                        ):
                            post_process_vLLM_adapters_new_tokens(
                                checkpoint_dir,
                                checkpoint_dir,
                                num_added_tokens,
                            )
```

tests/test_sft_trainer.py
```diff
@@ -1447,6 +1447,49 @@ def test_run_moe_ft_and_inference_ep1_kernels(dataset_path, ep_degree):
...
+        train_args = copy.deepcopy(TRAIN_ARGS)
+        train_args.output_dir = tempdir
+        lora_args = copy.deepcopy(PEFT_LORA_ARGS)
+        lora_args.r = 16
+        lora_args.target_modules = [
+            "q_proj",
+            "v_proj",
+            "o_proj",
+            "k_proj",
+        ]  # Router doesn't work with LoRA test inference
```
kmehant, line 1476: Can we test parameterize target modules as well so that we can run the same test for multiple target_modules inputs?

possible inputs to try

```
all-linear
```

```
q_proj
```

```
q_proj, k_proj
```

```
q_proj, k_proj, v_proj
```

```
q_proj, k_proj, v_proj, o_proj
```

tests/test_sft_trainer.py
```diff
@@ -1447,6 +1447,49 @@ def test_run_moe_ft_and_inference_ep1_kernels(dataset_path, ep_degree):
...
+        fast_moe_config = FastMoeConfig(fast_moe=FastMoe(ep_degree=False))
```
kmehant, line 1477: Lets also parameterise ep_degree to try for ep_degree true

tests/test_sft_trainer.py
```diff
@@ -1447,6 +1447,49 @@ def test_run_moe_ft_and_inference_ep1_kernels(dataset_path, ep_degree):
...
+            train_args,
+            lora_args,
+            fast_moe_config=fast_moe_config,
+        )
+        _test_run_inference(
+            checkpoint_path=os.path.join(
+                _get_checkpoint_path(tempdir), "hf_converted_checkpoint"
+            ),
+            base_model_name_or_path="ibm-granite/granite-3.1-1b-a400m-base",
+        )
```
kmehant, line 1490: Nice!

tuning/config/acceleration_configs/fast_moe.py
```diff
@@ -121,10 +122,29 @@ def checkpoint(checkpoint_dir, save_dir):
...
                         )
-                        # Save model config files
-                        self.trainer.model.config.save_pretrained(
-                            hf_converted_output_dir
-                        )
+
+                        # Unwrap FSDP module
+                        model = self.trainer.model
+                        if hasattr(model, "module"):
+                            model = model.module
```
kmehant, line 129: is this needed for non lora use case? since after this step it would directly reach the below block

```
                        else:
                            model.config.save_pretrained(hf_converted_output_dir)
```

tuning/sft_trainer.py
```diff
@@ -155,6 +155,23 @@ def train(
                 "Trainer should not perform packing when using `--padding_free`"
             )
 
+    if fast_moe_config is not None:
+        # Checking for unsupported modules with Scatter MoE for LoRA
+        restricted_modules = ["all-linear", "output_linear", "input_linear", "router"]
```
kmehant, line 160: Lets not hard code `"output_linear", "input_linear", "router"` since these vary model to model. I understand this is the best attempt, however, lets not do this. Rather, lets only restrict `all-linear` however, raise a warning that modules that are part of moe model if specific can lead to unexpected behaviour or unsuccessful tuning. And then place the recommendation as well saying restrict to linear layers outside moe module like the attention layer (q_proj, k_proj ..) etch. WDYT?

kmehant:
@willmj thanks for diligently taking my inputs, after addressing my comments, I request you to run all tests related to moe kernels on your local setup and paste screenshots here for record. Thanks, appreciate this PR!

tuning/config/acceleration_configs/fast_moe.py
willmj: It's needed to determine if there is a peft config for the following if statement, but in fine-tuning case this if statement should be false

tuning/sft_trainer.py
```diff
@@ -168,6 +168,40 @@ def train(
...
+                for module in restricted_modules
+            )
+        ):
+            raise ValueError(
+                "`--fast_moe` with LoRA does not currently support `all-linear`, as "
+                "target modules at this time. Please explicitly specify target "
+                "modules when using `--fast_moe` with LoRA."
+            )
+        # If other common non-linear modules, raise warning
+        restrained_modules = ["input_linear", "output_linear", "router"]
```
fabianlim, line 189: this is a potential hardcode

willmj: Made the warning more general, what do you think?

tuning/sft_trainer.py
```diff
@@ -390,6 +387,16 @@ def train(
         model, (peft_config,) = framework.augmentation(
             model, train_args, modifiable_args=(peft_config,)
         )
+        # For LoRa ScatterMoE, if expert layers are included, disable grad
+        if peft_config is not None:
```
fabianlim, line 391: I think its better to just do an instance check using `ScatterMoE` and then freeze everything inside.

fabianlim: maybe you should put a comment that this is a workaround, and in the future where ScatterMoE loras can be tuned, this code needs to be removed.

fabianlim (approved):
Since @kmehant is not around I approve it. but TBH im not that familiar with the recent updates in this repo. But overall looks ok. Just one more comment

fabianlim (approved):

PR #534, opened 2025-04-19
feat: allow specifying the chat template as base64 to avoid weird escaping and templating issues

HarikrishnanBalagopal:
### Was the PR tested


dushyantbehl (changes requested):
@HarikrishnanBalagopal Please add a unit test for the base64 chat template and also say somewhere that this is a undocumented feature...i.e. in the schema you should try to add `chat_template_base64` but add a note that this is something users should use if they know what they are doing. 

Any way to check if the base64 decode happened correctly? or throw some error if it didn't result in a chat template?

tuning/data/data_config.py
```diff
@@ -153,6 +154,18 @@ def _validate_dataprocessor_config(dataprocessor_config) -> DataPreProcessorConf
         chat_template = kwargs["chat_template"]
         assert isinstance(chat_template, str), "chat_template should be a string"
         c.chat_template = chat_template
+    elif "chat_template_base64" in kwargs:
+        chat_template_base64 = kwargs["chat_template_base64"]
+        assert isinstance(
+            chat_template_base64, str
+        ), "chat_template_base64 should be a string"
+        logger.warning(
```
dushyantbehl, line 162: you should also look to print the chat tempalte along side the warning...what was decoded one so users know what was picked up.

tuning/data/data_config.py
```diff
@@ -153,6 +154,18 @@ def _validate_dataprocessor_config(dataprocessor_config) -> DataPreProcessorConf
...
+            "You are using the 'chat_template_base64' field. "
+            + "Please use the 'chat_template' field instead for better readability."
+        )
+        chat_template_bytes = b64decode(chat_template_base64)
+        chat_template = chat_template_bytes.decode("utf-8")
```
dushyantbehl, line 167: Always `utf-8`?

tuning/data/data_config.py
HarikrishnanBalagopal: > ...what was decoded one so users know what was picked up.
ok but it's not a string then we can't decode it right?

tuning/data/data_config.py
HarikrishnanBalagopal: yes, the `chat_template` is always a valid utf-8 string.

HarikrishnanBalagopal:
> and also say somewhere that this is a undocumented feature...i.e. in the schema you should try to add `chat_template_base64` but add a note that this is something users should use if they know what they are doing.

Then it become documented wouldn't it?

HarikrishnanBalagopal:
> @HarikrishnanBalagopal Please add a unit test for the base64 chat template and also say somewhere that this is a undocumented feature...i.e. in the schema you should try to add `chat_template_base64` but add a note that this is something users should use if they know what they are doing.
> 
> Any way to check if the base64 decode happened correctly? or throw some error if it didn't result in a chat template?

I have added tests

dushyantbehl (changes requested):
Minor changes requested

dushyantbehl:
@HarikrishnanBalagopal requested minor name changes and post them please also fix the format error

tests/artifacts/predefined_data_configs/__init__.py
```diff
@@ -43,6 +43,20 @@
...
+)
+DATA_CONFIG_MULTITURN_CHAT_TOKENIZE_AND_MASKING_DATA_INVALID_BASE64_HANDLER = (
+    os.path.join(
+        PREDEFINED_DATA_CONFIGS,
+        "mt_data_granite_3_1B_tokenize_and_mask_invalid_base64_handler.yaml",
+    )
+)
+DATA_CONFIG_MULTITURN_CHAT_TOKENIZE_AND_MASKING_DATA_BASE64_HANDLER_EXPECTED_CHAT_TEMPLATE = os.path.join(
+    PREDEFINED_DATA_CONFIGS,
+    "mt_data_granite_3_1B_tokenize_and_mask_base64_handler_expected_chat_template.txt",
```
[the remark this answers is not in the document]
HarikrishnanBalagopal: done

HarikrishnanBalagopal:
> @HarikrishnanBalagopal requested minor name changes and post them please also fix the format error

done

dushyantbehl (approved):
LGTM

PR #536, opened 2025-04-24
fix: yaml parsing converts floats like 1.0 into integer 1 causing assertion failure

HarikrishnanBalagopal:
### Description of the change

Yaml parsing converts floats like 1.0 into integer 1 causing assertion failure of sampling ratio. This patch fixes the assertion condition.

### Was the PR tested


kmehant (approved):
Thanks for the update.

tuning/data/data_config.py
```diff
@@ -100,6 +100,9 @@ def _validate_dataset_config(dataset_config) -> DataSetConfig:
         c.builder = builder
     if "sampling" in kwargs and kwargs["sampling"] is not None:
         ratio = kwargs["sampling"]
+        if isinstance(ratio, int):
+            # YAML parsing causes 1.0 (float) to be cast to 1 (int) causing assertion error
+            ratio = float(ratio)
         assert isinstance(ratio, float) and (
```
dushyantbehl, line 106: Requesting rewrite of the logic to this 

```
       assert isinstance(ratio, (float, int)) and (
            0 <= float(ratio) <= 1.0
        ), f"sampling ratio: {ratio} should be float and in range [0.0,1.0]"
        c.sampling = float(ratio)
```

dushyantbehl (changes requested):
Please update PR description with the error you encountered and the reason to fix this.

tuning/data/data_config.py
HarikrishnanBalagopal: done

PR #538, opened 2025-04-25
ci: use --no-build-isolation flag for mamba

aluu317:
### Description of the change
Use `no-build-isolation` so mamba's build requirements do not interfere with our library's during image build process

### Was the PR tested


willmj (approved):

dushyantbehl (approved):
Looks Good. Thanks @aluu317 @willmj

PR #540, opened 2025-04-25
build(deps): upgrade trl

willmj:
### Description of the change
Upgrade trl, and to comply with trl we use default collator `DataCollatorForLanguageModeling` for pre-tokenized dataset instead of `DataCollatorForSeq2Seq`, removing unit tests to check for Seq2Seq until we support Seq2Seq on our side.

### Was the PR tested


Abhishek-TAMU (approved):
LGTM.

PR #544, opened 2025-04-29
fix: incorrect check on fast moe activation

kmehant:
Incorrect check leading to unnecessary lora checks to non fast moe runs. Introduced in https://github.com/foundation-model-stack/fms-hf-tuning/pull/530


dushyantbehl (dismissed):
LGTM

dushyantbehl (approved):
LGTM

Issue #142, opened 2024-04-25
feat: support for robust benchmarking of fms-hf-tuning


VassilisVassiliadis:
## Is your feature request related to a problem? Please describe.

When running benchmarks using fms-hf-tuning we need to:

1. record exceptions such as GPU Out Of Memory, transient RuntimeError and NCCL errors, etc
2. ensure robust capture of system metrics
3. annotate AIM run information with additional information regarding the benchmarking in order to make the AIM data searchable via benchmarking labels e.g. benchmark experiment name, benchmark identifier

However fms-hf-tuning does not support these features hence requiring maintenance of external forks of fms-hf-tuning that provide them. Specifically
 
1. fms-hf-tuning does not surface the various exceptions in a way that is easily identifiable by an automated process I.e. currently requires grepping stderr. 
2. system metrics are sent to an AIM server. However we have observed stability issues with storing and retrieving the data from AIM leading to lost data. In general for benchmarking we need a highly robust way to ensure the metrics are captured
3. There is no way to supply external metadata for insertion to AIM for a fms-hf-tuning run 

## Describe the solution you'd like

Here is a simplified view of the current state of the `sft_trainer.py` script:

```python
def train(
        model_args,
        data_args,
        training_args,
        tune_config,
        trainer_controller_args,
    ):
    callbacks = [SomeCallback()]

    if aim_available:
        callbacks.append(get_aimstack_callback())

    callbacks.extend(other_callbacks)

    sft_trainer.train(callbacks=callbacks)

def main():
   (
        model_args,
        data_args,
        training_args,
        tune_config,
        trainer_controller_args,
    ) = parser.parse_args_into_dataclasses(return_remaining_strings=True)

   train(
        model_args,
        data_args,
        training_args,
        tune_config,
        trainer_controller_args,
    )
```

We propose the following

```python
class BenchmarkingFriendlyAIMCallBack(AimCallBack):
    def on_train_begin(self, args, state, control, model=None, **kwargs):
        super().on_train_begin(args, state, control, model, **kwargs)
        # annotate self.run with custom metadata so that we can search for this run on AIM
    
    def on_train_end(self, args, state, control, **kwargs):
        if state.is_world_process_zero:
            # Dump aim metrics to a file such that we can collect this data without actually spinning up an AIM server
            ...
        super().on_train_end(args=args, state=state, control=control, **kwargs)

def parse_args():
    return (
        model_args,
        data_args,
        training_args,
        tune_config,
[...]

PR #545, opened 2025-04-29
fix: fastmoe lora `all-linear` excluding `router`

willmj:
### Description of the change
Goes with [#142](https://github.com/foundation-model-stack/fms-acceleration/pull/142) to allow `all-linear` for ScatterMoE. Without the fms-acceleration PR to filter it out, router was included since it is a linear layer, but will not be able to run vLLM inference if it is included. With the fms-acceleration PR, `all-linear` can be passed because `router` will always be filtered out

### Was the PR tested

### DIFF
--- a/tuning/sft_trainer.py
+++ b/tuning/sft_trainer.py
@@ -188,7 +188,11 @@ def train(
                 "modules when using `--fast_moe` with LoRA."
             )
         # If other common non-linear modules, raise warning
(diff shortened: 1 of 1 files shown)


anhuong (approved):
LGTM, allowing all-linear now but depends on this PR https://github.com/foundation-model-stack/fms-acceleration/pull/142 being merged first

PR #546, opened 2025-04-29
docs: add addt vision model support

anhuong:
### Description of the change

Adding models that we have not tested but are the same architecture as models we have tested and are different sizes.

### Was the PR tested

### DIFF
--- a/README.md
+++ b/README.md
@@ -909,7 +909,7 @@ For information on supported dataset formats and how to tune a vision-language m
 
 ### Supported vision model
 
-Note that vision models are supported starting with `fms-hf-tuning` v2.8.0 or later.
+Note that vision models are supported starting with `fms-hf-tuning` v2.8.1 or later.
 
 - Legend:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -909,6 +909,8 @@ For information on supported dataset formats and how to tune a vision-language m
 
 ### Supported vision model
 
+Note that vision models are supported starting with `fms-hf-tuning` v2.8.0 or later.
```
Abhishek-TAMU, line 912: We can update the [version](https://github.com/foundation-model-stack/fms-hf-tuning/releases/tag/v2.8.1) to `v2.8.1`.

Abhishek-TAMU (commented):
Looks good to me apart from small nit. 

Also just an observation. Model with vision architecture `LlavaNextForConditionalGeneration` has different Text model architecture:

`Llava Mistral 1.6-7B`: [MistralForCausalLM](https://huggingface.co/llava-hf/llava-v1.6-mistral-7b-hf/blob/main/config.json#L34)
`Llava 1.6-34B`: [LlamaForCausalLM](https://huggingface.co/llava-hf/llava-v1.6-34b-hf/blob/main/config.json#L34)
`Llava next-72B`: [Qwen2ForCausalLM](https://huggingface.co/llava-hf/llava-next-72b-hf/blob/main/config.json#L34)

But I guess having different Text model architecture shouldn't be an issue when tuning these models.

README.md
anhuong: Good catch, and good point on the different text models. I have tested on llava1.6-34b but hit into OOM issues and since we have run LlamaForCausalLM, I will leave this model and remove the Llava-next one since we haven't tested on Qwen2ForCausalLM

Abhishek-TAMU (approved):
LGTM.

PR #549, opened 2025-05-07
docs: add instructions on how to correctly specify the chat template

HarikrishnanBalagopal:
### Description of the change

Add docs to avoid common pitfalls while specifying the chat template.

### Was the PR tested


dushyantbehl (approved):
Looks good. Thanks @HarikrishnanBalagopal

PR #557, opened 2025-06-04
fix: decouple and update concatenate_dataset functionality from load_dataset

YashasviChaurasia:
### Description of the change

This PR refactors the dataset loading logic by **decoupling the concatenation functionality** from the `load_dataset` method from DataPreProcessor Class 

A new dedicated wrapper method for the Hugging Face `datasets.concatenate_datasets` API is written, providing capabilities to support concatenation of:

- **`DatasetDict` and `IterableDatasetDict` objects**, with the requirement that all dictionaries share the **same split keys**.
- In case of Multiple Datapaths,  Datasets are of **same type** (i.e. one of  `Dataset`, `DatasetDict`, `IterableDataset`, `IterableDatasetDict`).

### How to verify the PR

- Use existing test cases loading multiple datasets as lists of `DatasetDict` or `IterableDatasetDict`.
- Verify concatenation raises errors if split keys mismatch or dataset types differ.
- Confirm existing dataset loading flows remain unaffected.

### Was the PR tested

### DIFF
--- a/tests/data/test_data_preprocessing.py
+++ b/tests/data/test_data_preprocessing.py
@@ -19,7 +19,7 @@ import os
 import tempfile
 
 # Third Party
-from datasets import Dataset, IterableDataset
+from datasets import Dataset, DatasetDict, IterableDataset
 from PIL import Image
 from transformers import AutoProcessor, AutoTokenizer, DataCollatorForSeq2Seq
 from trl import DataCollatorForCompletionOnlyLM
@@ -84,6 +84,7 @@ from tuning.data.setup_dataprocessor import (
     is_pretokenized_dataset,
     process_dataargs,
 )
+from tuning.data.utils import try_concatenate_datasets
 
 
 @pytest.mark.parametrize(
@@ -776,6 +777,48 @@ def test_process_dataconfig_file_with_streaming(data_config_path, data_path):
         assert set(["input_ids", "labels"]).issubset(set(train_set.column_names))
     elif datasets_name == "apply_custom_data_template":
         assert formatted_dataset_field in set(train_set.column_names)
+
+
+def test_concatenate_dict_with_multi_keys():
+    """
+    Ensure that concatenated datasets are formatted and validated correctly.
+    Ensures the returned dataset has proper concatenation
+
+    Details for Concatenation Operation of dictionary with different keys
+        data                        => { "train": Values }
+        data_dict1                  => { "train": Values, "train2": Values }
+        data_dict2                  => { "train": Values, "train2": Values, "train3": Values }
+        ------------------------------------------------------------------------------------------
+        concatenated_dataset        => { "train": Values*3, "train2": Values*2, "train3": Values }
+    """
+
+    data_paths = TWITTER_COMPLAINTS_DATA_INPUT_OUTPUT_JSON
+    data = datasets.load_dataset("json", data_files=[data_paths])
+    data_streaming = datasets.load_dataset(
+        "json", data_files=[data_paths], streaming=True
+    )
+
+    data_dict1 = DatasetDict()
+    data_dict1["train"] = data["train"]
+    data_dict1["train2"] = data["train"]
+
+    data_dict2 = DatasetDict()
+    data_dict2["train"] = data["train"]
+    data_dict2["train2"] = data["train"]
+    data_dict2["train3"] = data["train"]
+
+    concatenated_dataset = try_concatenate_datasets([data, data_dict1, data_dict2])
+
+    # Check if the datasets are concatenated correctly
+    assert (
+        len(concatenated_dataset) == 3
+        and concatenated_dataset["train"].num_rows == data["train"].num_rows * 3
+        and concatenated_dataset["train2"].num_rows == data["train"].num_rows * 2
+        and concatenated_dataset["train3"].num_rows == data["train"].num_rows
+    )
+    # Assert ValueError on concatenation of mixed dataset types (only same types supported)
+    with pytest.raises(ValueError):
+        try_concatenate_datasets([data, data_streaming])
 
 
 @pytest.mark.parametrize(
--- a/tuning/data/data_processors.py
+++ b/tuning/data/data_processors.py
@@ -35,6 +35,7 @@ from tuning.data.utils import (
     get_loader_for_filepath,
     maybe_align_datasets,
     resolve_iterable_dataset_features,
+    try_concatenate_datasets,
 )
 
 logger = logging.getLogger(__name__)
@@ -101,7 +102,7 @@ class DataPreProcessor:
         self,
         datasetconfig: DataSetConfig,
         streaming: bool,
-        splitName: str,
+        splitName: str = None,
         datafile: str = None,
         **kwargs,
     ):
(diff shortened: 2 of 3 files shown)


tuning/data/data_processors.py
```diff
@@ -398,6 +382,68 @@ def _execute_data_handlers(
               {",".join([e.name for e in DataHandlerType])}'
         )
 
+    def _concatenate_dict_datasets(
```
dushyantbehl, line 385: rename to `_concatenate_datasets_dict`

tuning/data/data_processors.py
```diff
@@ -398,6 +382,68 @@ def _execute_data_handlers(
...
+
+    def _concatenate_simple_datasets(
+        self, all_datasets: List[Union[Dataset, IterableDataset]]
+    ) -> Union[Dataset, IterableDataset]:
+        maybe_align_datasets(all_datasets)
+        concatenated = datasets.concatenate_datasets(all_datasets)
+        if isinstance(concatenated, IterableDataset):
+            concatenated = resolve_iterable_dataset_features(concatenated)
+
+        logger.info(
```
dushyantbehl, line 414: remove this log.

tuning/data/data_processors.py
dushyantbehl: move this function to utils file and remove self from here.

tuning/data/data_processors.py
```diff
@@ -398,6 +382,68 @@ def _execute_data_handlers(
...
+    ) -> Union[DatasetDict, IterableDatasetDict]:
+        all_keys = [set(d.keys()) for d in all_datasets]
+        first_keys = all_keys[0]
+        if not all(keys == first_keys for keys in all_keys):
+            raise ValueError(
+                f"DatasetDicts do not have matching keys. Found keys: {all_keys}"
+            )
+
+        concatenated_dict = {}
+        for key in first_keys:
```
dushyantbehl, line 396: simplify the loop 

```
return {
    key: _concatenate_datasets([d[key] for d in all_datasets])
    for key in first_keys
}
```

tuning/data/data_processors.py
```diff
@@ -398,6 +382,68 @@ def _execute_data_handlers(
...
+            key_datasets = [d[key] for d in all_datasets]
+            maybe_align_datasets(key_datasets)
+            concatenated = datasets.concatenate_datasets(key_datasets)
+            if isinstance(concatenated, IterableDataset):
+                concatenated = resolve_iterable_dataset_features(concatenated)
+            concatenated_dict[key] = concatenated
+
+        return DatasetDict(concatenated_dict)
+
+    def _concatenate_simple_datasets(
```
dushyantbehl, line 406: rename `_concatenate_datasets`

dushyantbehl (changes requested):
Also add unit tests.

tuning/data/data_processors.py
YashasviChaurasia: `_concatenate_datasets` is already used as the name of the wrapper function

YashasviChaurasia: does `_concatenate_flat_datasets` sound good?

kmehant:
rejected

dushyantbehl (changes requested):

tuning/data/utils.py
```diff
@@ -211,3 +219,69 @@ def try_convert_image_to_rgb(image):
...
+    if len(all_datasets) == 1:
+        return all_datasets[0]
+
+    try:
+        # Case 1: Flat datasets
+        if all(isinstance(d, (Dataset, IterableDataset)) for d in all_datasets):
+            return _concatenate_datasets(all_datasets)
+
+        # Case 2: Dict-style datasets
+        if all(isinstance(d, (DatasetDict, IterableDatasetDict)) for d in all_datasets):
```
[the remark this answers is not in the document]
YashasviChaurasia: Yeah the simpler code does not ensure the key frequency, in my case I would figure out unique and common keys before concatenation and only concatenate the common keys while would simply put the unique key as it is without calling concatenation on it as it is not required as that key only occurs in a single dataset hence saving on some function calls.. preventing redundant operations

Also the present version of code ensures the correct type of object is returned as per the input.. as we are ensuring all dataset objects should be of same type using `all` over `any`
Helpful explanation might be found [in the discussion below](https://github.com/foundation-model-stack/fms-hf-tuning/pull/557#discussion_r2136062179)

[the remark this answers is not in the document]
dushyantbehl: can you add a simple check in the base function `_concatenate_dataset` to return if `all_datasets` is a single object?

With that you won't need to perform the count check

tuning/data/utils.py
```diff
@@ -211,3 +219,69 @@ def try_convert_image_to_rgb(image):
...
+
+            # Merge common keys (appear in more than one dataset)
+            for key, count in key_count.items():
+                key_datasets = [d[key] for d in all_datasets if key in d]
+                if count > 1:
+                    merged_dict[key] = _concatenate_datasets(key_datasets)
+                else:
+                    merged_dict[key] = key_datasets[0]  # unique, take as-is
+
+            if all(isinstance(d, IterableDatasetDict) for d in all_datasets):
```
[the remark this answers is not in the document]
dushyantbehl: sure so the test with mixing datasets should also throw an error? maybe worth adding a unit test which shows the scenarios where the error is thrown.

tests/data/test_data_preprocessing.py
```diff
@@ -778,6 +779,46 @@ def test_process_dataconfig_file_with_streaming(data_config_path, data_path):
...
+    data_dict1["train2"] = data["train"]
+
+    data_dict2 = DatasetDict()
+    data_dict2["train"] = data["train"]
+    data_dict2["train2"] = data["train"]
+    data_dict2["train3"] = data["train"]
+
+    concatenated_dataset = try_concatenate_datasets([data, data_dict1, data_dict2])
+
+    assert (
```
dushyantbehl, line 812:
```suggestion
    # Check if the datasets are concatenated correctly
    assert (
```

tests/data/test_data_preprocessing.py
```diff
@@ -778,6 +779,46 @@ def test_process_dataconfig_file_with_streaming(data_config_path, data_path):
...
+        len(concatenated_dataset) == 3
+        and concatenated_dataset["train"].num_rows == data["train"].num_rows * 3
+        and concatenated_dataset["train2"].num_rows == data["train"].num_rows * 2
+        and concatenated_dataset["train3"].num_rows == data["train"].num_rows
+    )
+    # Assert ValueError on concatenation of mixed dataset types (only same types supported)
+    assert pytest.raises(ValueError, try_concatenate_datasets, [data, data_streaming])
```
dushyantbehl, line 819:
```suggestion
    with pytest.raises(ValueError):
        try_concatenate_datasets([data, data_streaming])
```

dushyantbehl: `pytest.raises` is a context manager and the assert will not work as intended if called inside single line.

dushyantbehl (changes requested):
Minor correction required.

dushyantbehl (approved):
Looks good to me.

PR #559, opened 2025-06-06
fix: saving logic for alora

kmehant:
Irrespective of the save_strategy, we only save once at the end of the training as per @kgreenewald. However, in such case we dont need save_model_dir to be defined and model is saved at the output_dir and global step checkpoint dir. In case save_model_dir is defined then the model is saved at the save_model_dir location. PR also removes hard coding of the checkpoint dir.

### DIFF
--- a/tests/test_sft_trainer.py
+++ b/tests/test_sft_trainer.py
@@ -762,6 +762,7 @@ def test_run_causallm_alora_and_inference(request, target_modules, expected):
             "Simply put, the theory of relativity states that \n" + invocation_string,
             max_new_tokens=50,
         )
+        print(output_inference)
         assert len(output_inference) > 0
         assert "Simply put, the theory of relativity states that \n" in output_inference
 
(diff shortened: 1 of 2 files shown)


kmehant:
Screenshot 2025-06-07 at 1 09 05 AM

aLoRA tests

kgreenewald:
> @ashokponkumar have cleaned up the saving part now we leverage inbuilt transformers saving mechanism.
> 
> This requires merge of [IBM/activated-lora#15](https://github.com/IBM/activated-lora/pull/15) to alora package.
> 
> Related tests pass:
> Screenshot 2025-06-10 at 2 48 16 AM

This should be live now as alora==0.3.0

ashokponkumar (approved):

ashokponkumar:
@kmehant Do we want to update the pyproject.toml too?

ashokponkumar:
@dushyantbehl can we merge this?

PR #560, opened 2025-06-08
feat: add train_test_split functionality via dataconfig

YashasviChaurasia:
## Description of the Change

This PR introduces the `_split_dataset` method to support flexible dataset splitting for both `Dataset`/`DatasetDict` and `IterableDataset`/`IterableDatasetDict`.

Key changes:
- Added `_split_dataset()` to support train/validation split using both `train_size` and `test_size`.
- Supports validation of split ratios and throws appropriate errors for invalid configurations.
- Streaming datasets only support full train or full validation splits.
- `_process_dataset_configs()` updated to call `_split_dataset()` when a `split` is specified in the dataset config.
- Ensures backward compatibility by leaving datasets unchanged if no split is specified.

## Related PR Number

Related to PR #557 
This PR assumes support from the above PR for concatenation !

## How to Verify the PR

1. Provide one or more `DataSetConfig` objects with `split={"train": x, "validation": y}`.
2. Call the `_process_dataset_configs()` method.
3. Verify that:
   - The dataset is correctly split when the `split` config is defined.
   - The data flow remains unchanged if no `split` is defined.
   - Streaming datasets only allow 100% train or 100% validation splits.
   - Appropriate errors are raised for unsupported or invalid configurations.
   
## Assumptions

The following assumptions were made while implementing the `_split_dataset` and dataset processing logic:

- **Single vs Multiple Datasets**:
  - `data_args` supports only a **single dataset**.
  - `data_configs` supports **multiple datasets**, each with individual configuration.

- **Sampling Support**:
  - Sampling (with probabilities) is **only supported** when using `data_configs`.
  - `data_args` does **not support** sampling or dataset interleaving.

- **Validation Handling**:
  - `data_args` can include a separate **validation dataset path**.
  - When using `data_configs`, validation data can be obtained either from a separate dataset or via **splitting** using the `split` field.

- **Train-Test Splitting**:
  - Splitting into train and validation using `train_test_split` is supported **only via `data_configs`** with the `split` dictionary.
  - `data_args` does **not support train-validation splitting** via `split` ratios.

- **Streaming Dataset Handling**:
  - Partial splits are **not supported** for streaming datasets.
  - Only full assignment to `train` or `validation` is supported for streaming datasets.

## Was the PR Tested?

- [x] I have added at least one unit test for every new method.
- [x] I have verified that all unit tests pass.

## The image build process depends on `peft` package upgrade 

The below screenshot provides a successful build with the changes in `pyproject.toml` so this check can be skipped.

image


tuning/data/data_processors.py
```diff
@@ -282,46 +282,38 @@ def __execute_filter_data_handler(self, raw_datasets, handler, **kwargs):
...
+            if not isinstance(raw_datasets, (IterableDatasetDict, IterableDataset)):
+                kwargs["desc"] = handler.desc
 
-        # IterableDatasets doesn't support any description
-        if not isinstance(raw_datasets, (IterableDatasetDict or IterableDataset)):
-            kwargs["desc"] = handler.desc
+            processed_splits[split_name] = split.map(handler.op, **kwargs)
 
-        return raw_datasets.map(handler.op, **kwargs)
+        return DatasetDict(processed_splits)
```
dushyantbehl, line 314: you could be changing an iterable dict to a dataset dict...please correct this line 

you should be returned `processed_ds`

tuning/data/data_processors.py
```diff
@@ -282,46 +282,38 @@ def __execute_filter_data_handler(self, raw_datasets, handler, **kwargs):
...
-                    unable to set `remove_columns` to all.\n \
-                    Please explicitly specify the column list or \
-                    use remove/select data handlers.",
-                    datasetName,
-                )
-            else:
-                kwargs["remove_columns"] = column_names
+    def __execute_map_data_handler(self, raw_datasets, handler, datasetName, **kwargs):
+        """Apply handler.op to all splits in raw_datasets instead of just a single split."""
+        processed_splits = {}
```
dushyantbehl, line 287:
```suggestion
        processed_ds = IterableDatasetDict() if isinstance(IterableDatasetDict, raw_datasets) else DatasetDict()
```

tuning/data/data_processors.py
```diff
@@ -282,46 +282,38 @@ def __execute_filter_data_handler(self, raw_datasets, handler, **kwargs):
...
 
-        if "fn_kwargs" not in kwargs:
-            kwargs["fn_kwargs"] = {}
+            if "fn_kwargs" not in kwargs:
+                kwargs["fn_kwargs"] = {}
+            kwargs["fn_kwargs"]["tokenizer"] = self.tokenizer
 
-        kwargs["fn_kwargs"]["tokenizer"] = self.tokenizer
-        kwargs["fn_kwargs"]["column_names"] = column_names
+            if not isinstance(raw_datasets, (IterableDatasetDict, IterableDataset)):
```
dushyantbehl, line 309: This has to be a `IterableDataset` now as it cannot be `dict` in any scenario after you pull out the splits.

tuning/data/data_processors.py
```diff
@@ -282,46 +282,38 @@ def __execute_filter_data_handler(self, raw_datasets, handler, **kwargs):
...
-                    Please explicitly specify the column list or \
-                    use remove/select data handlers.",
-                    datasetName,
-                )
-            else:
-                kwargs["remove_columns"] = column_names
+    def __execute_map_data_handler(self, raw_datasets, handler, datasetName, **kwargs):
+        """Apply handler.op to all splits in raw_datasets instead of just a single split."""
+        processed_splits = {}
+        for split_name, split in raw_datasets.items():
```
dushyantbehl, line 288: can you use enumerate here? also possibly simplify 
```
for split_name, ds in...

```

tuning/data/data_processors.py
```diff
@@ -282,46 +282,38 @@ def __execute_filter_data_handler(self, raw_datasets, handler, **kwargs):
...
+            column_names = getattr(split, "column_names", None)
```
dushyantbehl, line 289: move this line inside the `if kwargs["remove_columns"] == "all":` loop as this doesn't have any use outside

tuning/data/data_processors.py
```diff
@@ -383,11 +375,128 @@ def _execute_data_handlers(
...
 
+    def _split_dataset(
+        self,
+        dataset_config: DataSetConfig,
+        loaded_dataset: Union[
+            Dataset, DatasetDict, IterableDataset, IterableDatasetDict
+        ],
+    ) -> Union[DatasetDict, IterableDatasetDict]:
+
+        seed = self.processor_config.sampling_seed
```
dushyantbehl, line 386: I am not a big fan of piggybacking on top of `sampling_seed` 

if the `sampling_seed` is just a seed we use then you can rename DataPreProcessorConfig.sampling_seed to `DataPreProcessorConfig.seed`

dushyantbehl (changes requested):

tuning/data/data_processors.py
YashasviChaurasia: Sure before, the split support we only used the sampling seed for the interleave function.
but now as we use it for splitting too, I'll rename it!

tuning/data/data_processors.py
YashasviChaurasia: the `raw_datasets` variable is not being changed during the loop, and it still can be dictionary!

dushyantbehl: then we can base this check on `ds` I assume or create these few arguments once for all the splits as you won't have to again.

dushyantbehl (commented):
If we use just `train` and `validataion` split...shouldn't we just restrict all processing to these two?

tuning/data/data_processors.py
```diff
@@ -282,46 +282,43 @@ def __execute_filter_data_handler(self, raw_datasets, handler, **kwargs):
...
 
-        # IterableDatasets doesn't support any description
-        if not isinstance(raw_datasets, (IterableDatasetDict or IterableDataset)):
-            kwargs["desc"] = handler.desc
+        for split_name, ds in raw_datasets.items():
+
+            if kwargs["remove_columns"] == "all":
+                column_names = getattr(ds, "column_names", None)
+                if column_names is None:
+                    logger.warning(
```
dushyantbehl, line 307: This is not related to the PR changes but on a second thought can you make this an error and raise it 
this might end up doing damage to the user expercation if remove_columns is not adhered to.

tuning/data/data_processors.py
```diff
@@ -383,11 +380,128 @@ def _execute_data_handlers(
...
+        if total_split > 1.0 or total_split == 0:
+            raise ValueError(f"Sum of splits must be in (0,1]. Got {total_split}")
+
+        # Todo : Add support for using multiple existing splits during spliting
+        # Problem : DatasetDict might have multiple splits already, which to use?
+        # One or more than one, or existing splitName other than train
+        # Currently Supports "train" by default
+        use_existing_splits = ["train"]  # For DatasetDict
+
+        if streaming:
```
dushyantbehl, line 410: why do we need to check `if streaming` here? 
We can base all our checks on type of dataset

tuning/data/data_processors.py
```diff
@@ -383,11 +380,128 @@ def _execute_data_handlers(
...
+            )
+
+            return DatasetDict(
+                {
+                    "train": split_ds["train"],
+                    "validation": split_ds["test"],
+                }
+            )
+
+        raise TypeError(
```
dushyantbehl, line 492: why do we need this check here? is there a reason you expect the types to mismatch?

tuning/data/data_processors.py
```diff
@@ -383,11 +380,128 @@ def _execute_data_handlers(
...
+        use_existing_splits = ["train"]  # For DatasetDict
+
+        if streaming:
+            # For streaming IterableDataset / IterableDatasetDict
+            # Simple stub: either full train or full validation, no partial splits
+            # For example, if train_split >= 1.0, return all as train,
+            # else if validation_split >= 1.0, all as validation
+            # Partial Splits not Supported yet!
+
+            if isinstance(loaded_dataset, IterableDataset):
```
dushyantbehl, line 417: Can you remove the heavy code duplication between this and the below function...it seems to be same logic in `DatasetDict` as well.

dushyantbehl (changes requested):
Please address comments.

tuning/data/data_processors.py
```diff
@@ -383,11 +380,128 @@ def _execute_data_handlers(
...
+    ) -> Union[DatasetDict, IterableDatasetDict]:
+
+        seed = self.processor_config.seed
+        streaming = self.processor_config.streaming
+
+        train_split = dataset_config.split.get("train")
+        validation_split = dataset_config.split.get("validation")
+
+        if train_split is None or validation_split is None:
+            raise ValueError("Both 'train' and 'validation' splits must be specified!")
```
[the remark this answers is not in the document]
YashasviChaurasia: We intentionally require both train and validation keys to be explicitly set in dataset_config.split to avoid assumptions that the sum of the split ratios should be 1.
Altho we can infer validation split from train split for the corner case you have specified but it might not be always helpful incase the user only specifies train 0.3 we cannot assume validation split then as they might not sum to 1.

Hence I thought it would be better to explicitly ask the user

tuning/data/data_processors.py
YashasviChaurasia: Yeah we could have done it that way too but in my opinion it would be better to use streaming flag from data config as it would be better to decide intent of the  user from the config rather than from the type of Dataset the function receives.. altho both will work just fine but in my opinion I think we should let it be as it would be for better readability and design..

tuning/data/data_processors.py
YashasviChaurasia: this was to ensure we catch if we receive something other than the following formats!
maybe due to some incomplete load or something!
more like safety catch!
I can remove this if required but it seems it can be good to have..
like just in case you receive None type from load_dataset if it fails or anything other than the required Dataset 
types!

docs/advanced-data-preprocessing.md
```diff
@@ -184,6 +185,45 @@ We also allow users to pass a [`seed`](https://huggingface.co/docs/datasets/v3.2
...
 
+### Dataset Splitting
+
+In addition to sampling and mixing, our library supports **dataset splitting**, which allows users to split a dataset into training and validation splits using the `split` field in the dataset config.
+
+This is especially useful when users want to split a single dataset (or multiple datasets) internally instead of supplying separate files for training and validation.
+
+#### How to Use
+
+To split a dataset, simply pro
```
dushyantbehl, line 196: This line needs to be corrected

docs/advanced-data-preprocessing.md
```diff
@@ -184,6 +185,45 @@ We also allow users to pass a [`seed`](https://huggingface.co/docs/datasets/v3.2
...
+The `split` field in each dataset config allows users to internally divide a dataset into `train` and `validation` sets using fractional ratios.
+
+To use it, specify both `train` and `validation` values under the `split` key. Example:
```
dushyantbehl, line 199:
```suggestion
To use it, specify both `train` and `validation` ratios values under the `split` key for each dataset. Example:
```

docs/advanced-data-preprocessing.md
```diff
@@ -184,6 +185,45 @@ We also allow users to pass a [`seed`](https://huggingface.co/docs/datasets/v3.2
...
+- If no `split` is defined, the dataset is returned unchanged.
+
+**Streaming Datasets (`IterableDataset`, `IterableDatasetDict`)**:
+- Only supports full splits:
+  - Either `train: 1.0, validation: 0.0`
+  - Or `train: 0.0, validation: 1.0`
+- Partial splits like `train: 0.8, validation: 0.2` are not supported and will raise a `NotImplementedError`.
+- If no `split` is defined, the dataset is returned unchanged.
+- Streaming behavior must be explicitly enabled via `dataprocessor.streaming: true`.
+
```
dushyantbehl, line 226: Add a line in bold that users need to explicitly set `eval_strategy` link to HF argument to something valid for eval to run and simply splitting a dataset will not result in eval and will likely result in an error. 

Also set a default config parameter in our training args `eval_strategy="no"`

tuning/data/data_processors.py
```diff
@@ -383,11 +377,70 @@ def _execute_data_handlers(
...
+        dataset: Union[Dataset, DatasetDict, IterableDataset, IterableDatasetDict],
+    ) -> Union[DatasetDict, IterableDatasetDict]:
+
+        seed = self.processor_config.seed
+
+        train_size = dataset_config.split.get("train")
+        validation_size = dataset_config.split.get("validation")
+
+        if train_size is None or validation_size is None:
+            raise ValueError("Both 'train' and 'validation' splits must be specified!")
```
dushyantbehl, line 392:
```suggestion
            raise ValueError("Both 'train' and 'validation' splits must be specified for a dataset in data_config")
```

tuning/data/data_processors.py
```diff
@@ -458,39 +513,52 @@ def _process_dataset_configs(
...
-            seed = self.processor_config.sampling_seed
+            seed = self.processor_config.seed
             logger.info(
                 "Interleaving datasets: strategy[%s] seed[%d] probabilities[%s]",
                 strategy,
                 seed,
                 str(sampling_probabilities),
             )
+            # Default sampling for "train" split, concatenates remaining splits
+            split_to_interleave = ["train"]
```
dushyantbehl, line 524:
```suggestion
            splits_to_interleave = ["train"]
```

tuning/data/data_processors.py
```diff
@@ -458,39 +513,52 @@ def _process_dataset_configs(
...
 
         train_dataset = final_datasets.get("train", None)
+        eval_dataset = None
+        if "validation" in final_datasets:
+            eval_dataset = final_datasets.get("validation", None)
+            if isinstance(eval_dataset, IterableDataset):
+                eval_dataset = resolve_iterable_dataset_features(eval_dataset)
 
         # Just a failsafe in case this is required later.
         if isinstance(train_dataset, IterableDataset):
```
dushyantbehl, line 552: This code should be below line 544

tuning/data/setup_dataprocessor.py
```diff
@@ -312,6 +324,11 @@ def _process_raw_data_args(
     is_eval_dataset_present = False
     if data_args.validation_data_path:
         is_eval_dataset_present = True
+    if data_args.data_config_path is not None:
```
dushyantbehl, line 327: can you move this check to the top?

dushyantbehl (changes requested):
Minor changes request rest looks good.

dushyantbehl (approved):
LGTM!

PR #575, opened 2025-06-26
data pre-processor: add to_json flag to jinja handler

ChanderG:
add a new flag to cast the output of the jinja handler to json


dushyantbehl (changes requested):
@ChanderG as discussed internally can we convert this to a separate data handler which can take any column(s) and then convert them to base format?

Also have you had any thought on how the final element should look like? do we want columns for all keys in the json or whole json should be under a single column...
It would be good if we can create a generic piece here which could fit many potential upcoming preference style data engagements.

cc @ashokponkumar

PR #581, opened 2025-07-11
feat: add error handling for split dataset feat

YashasviChaurasia:
### Description of the change

Added error handling for the dataset splitting and streaming configurations in the `process_dataargs` flow. Specifically:
- Raise `ValueError` if `eval_strategy` is set to anything other than `"no"` but no evaluation dataset is provided.
- Raise `ValueError` if `padding_free` is enabled while streaming is also enabled, which is not supported.

### Related issue number

N/A

### How to verify the PR

- Try setting `eval_strategy="epoch"` or `"steps"` without providing an eval set → should raise an error.
- Try enabling `padding_free=huggingface` while using streaming dataset → should raise an error.

### Was the PR tested

- [x] I have added ≥1 unit test for each new case I handled.
- [x] I have ensured all unit tests pass.

added screenshot for Error handling when train_set is None using the attached dataconfig:
```
dataprocessor:
    type: default
    sampling_stopping_strategy: first_exhausted
    seed: 66
datasets:
  - name: dataset_1
    split:
      train: 0.0
      validation: 1.0
    sampling: 0.3
    data_paths:
      - "FILE_PATH"
    data_handlers:
      - name: tokenize_and_apply_input_masking
        arguments:
          remove_columns: all
          batched: false
          fn_kwargs:
            input_column_name: input
            output_column_name: output
  - name: dataset_2
    split:
      train: 0.0
      validation: 1.0
    sampling: 0.4
    data_paths:
      - "FILE_PATH"
    data_handlers:
      - name: tokenize_and_apply_input_masking
        arguments:
          remove_columns: all
          batched: false
          fn_kwargs:
            input_column_name: input
            output_column_name: output
  - name: dataset_3
    split:
      train: 0.0
      validation: 1.0
    sampling: 0.3
    data_paths:
      - "FILE_PATH"
    data_handlers:
      - name: tokenize_and_apply_input_masking
        arguments:
          remove_columns: all
          batched: false
          fn_kwargs:
            input_column_name: input
            output_column_name: output
```
image

### DIFF
--- a/tuning/data/setup_dataprocessor.py
+++ b/tuning/data/setup_dataprocessor.py
@@ -70,6 +70,7 @@ def process_dataconfig_file(
     additional_data_handlers: Dict[str, DataHandler] = None,
     processor: AutoProcessor = None,
     is_multipack: bool = False,
+    is_padding_free: bool = False,
 ):
     """
     Args:
@@ -114,6 +115,14 @@ def process_dataconfig_file(
     )
 
     if data_processor.processor_config.streaming:
+        if is_padding_free:
+            logging.error(
+                "`padding_free` is not supported when streaming is enabled.",
+            )
+            raise ValueError(
+                "`--padding_free` is not allowed when `streaming=True`. "
+                "Please remove the `padding_free` argument from your configuration."
+            )
         if train_args.max_steps < 1:
             logging.error(
                 "ValueError: `--max_steps` must be set when streaming is set in data "
@@ -549,6 +558,7 @@ def process_dataargs(
             additional_data_handlers,
             processor,
             is_multipack,
+            is_padding_free,
         )
     else:
         train_dataset, eval_dataset, dataset_text_field = _process_raw_data_args(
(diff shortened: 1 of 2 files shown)


tuning/data/setup_dataprocessor.py
```diff
@@ -561,6 +570,12 @@ def process_dataargs(
             processor,
         )
 
+    if getattr(train_args, "eval_strategy", "no") != "no" and eval_dataset is None:
```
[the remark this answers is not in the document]
YashasviChaurasia: to Avoid AttributeError if the field doesn't exist

tuning/data/setup_dataprocessor.py
```diff
@@ -114,6 +114,15 @@ def process_dataconfig_file(
     )
 
     if data_processor.processor_config.streaming:
+        if getattr(train_args, "padding_free", None):
```
dushyantbehl, line 117: you have to use is_padding_free
its passed here https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/tuning/data/setup_dataprocessor.py#L491
there's no padding_free in train_args so your code has a bug

dushyantbehl (changes requested):

tuning/data/setup_dataprocessor.py
YashasviChaurasia: actually it won't be a bug ig as if its not present it wont throw an error, but if it is passed it ensures the if condition.

dushyantbehl (approved):

PR #592, opened 2025-08-04
fix: Disable caching in transformers via `use_cache` flag to avoid unncessary memory overhead

romitjain:
### Description of the change

Sets the `use_cache=False`, disabling the model's KV cache while fine-tuning the model.

### Related issue number

Solves https://github.com/huggingface/transformers/issues/39795 for our stack.

### How to verify the PR
--

### Was the PR tested

Since this was a bug fix, I have tested multiple settings of the flags to verify that the memory usage has dropped for FSDP+LoRA use case.

In my toy example, with 4 GPUs on an 8b model with BF16 datatype, memory dropped from 28GB per GPU to 12GB per GPU.

### DIFF
--- a/tuning/config/configs.py
+++ b/tuning/config/configs.py
@@ -61,16 +61,6 @@ class ModelArguments:
                 tokenizer classes."
         },
     )
-    use_cache: bool = field(
-        default=False,
-        metadata={
-            "help": "This enables/disables the KV cache for model generation."
-            "This should be set to False in most cases, unless you are doing a finetuning"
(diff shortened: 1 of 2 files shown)


dushyantbehl (changes requested):
@romitjain since the caching is generic for models we might wanna change the vision model loaders in [this line](https://github.com/foundation-model-stack/fms-hf-tuning/blob/main/tuning/sft_trainer.py#L272) of the code as well as part of this PR.

PR #599, opened 2025-08-26
chore: add changes to support Granite 4 models

YashasviChaurasia:
### Description of the change
updates required dependencies to fix mamba-ssm incompatibility with various package versions.

### Was the PR tested

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,7 +34,7 @@ dependencies = [
 "sentencepiece>=0.1.99,<0.3",
 "tokenizers>=0.13.3,<1.0",
 "tqdm>=4.66.2,<5.0",
-"trl>=0.13,<0.18",
+"trl>=0.13,<0.20",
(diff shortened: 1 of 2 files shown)


build/Dockerfile
```diff
@@ -151,6 +151,9 @@ RUN if [[ -z "${WHEEL_VERSION}" ]]; \
     fi && \
     ls /tmp/*.whl >/tmp/bdist_name
 
+# Ensures to always build mamba_ssm from source
+ENV PIP_NO_BINARY=mamba-ssm,mamba_ssm
```
kmehant, line 155: Lets make sure that selective_scan binaries also built which may get ignored.

```
FORCE_BUILD=TRUE
SKIP_CUDA_BUILD=FALSE
```
is used during `pip install .`

YashasviChaurasia: image

kmehant: looks good

PR #600, opened 2025-08-28
fix: Rollback trl to a version which doesn't overwrite labels and attention masks.

dushyantbehl:
### Description of the change

This PR rolbacks trl to v16.1 to solve https://github.ibm.com/ai-foundation/watson-fm-stack-tracker/issues/1913

### Was the PR tested

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,7 +34,7 @@ dependencies = [
 "sentencepiece>=0.1.99,<0.3",
 "tokenizers>=0.13.3,<1.0",
 "tqdm>=4.66.2,<5.0",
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -34,7 +34,7 @@ dependencies = [
 "sentencepiece>=0.1.99,<0.3",
 "tokenizers>=0.13.3,<1.0",
 "tqdm>=4.66.2,<5.0",
-"trl>=0.13,<0.18",
+"trl>=0.13,<0.17",
```
ashokponkumar, line 37: If we are anyway returning the right collator, should we still revert the version?

PR #603, opened 2025-09-01
feat: Support gpt-oss class of models with flash attention support

dushyantbehl:
### Description of the change

1. Adds code changes required to support the new GPT-OSS models.
2. Adds a new dockerfile which is based on nvcr for dev mode and supports flash attention 3. 
3. Updates most of the required package versions. 
4. Adds a new quantization class for Mxfp4 which dequantizes the model before training.

### Was the PR tested


tuning/config/peft_config.py
```diff
@@ -14,16 +14,38 @@
...
+
 
 @dataclass
 class LoraConfig:
     """
     This is the configuration class to store the configuration of a [`LoraModel`].
 
     Args:
-        r (`int`):
+        lora_r (`int`):
```
dushyantbehl, line 48: @ashokponkumar would you welcome this change along with this PR? if not i can revert it

pyproject.toml
```diff
@@ -27,19 +27,19 @@ classifiers=[
...
 "sentencepiece>=0.1.99,<0.3",
-"tokenizers>=0.13.3,<1.0",
+"tokenizers<=0.22",
 "tqdm>=4.66.2,<5.0",
-"trl>=0.13,<0.20",
-"peft>=0.15.0,<=0.15.2",
-"protobuf>=5.28.0,<6.0.0",
-"datasets>=3.5.0,<4.0",
+"trl>=0.19.1,<0.20.0",
+"peft<=0.17.0",
```
ashokponkumar, line 38: don't we want the `>=0.15` filter for peft?

pyproject.toml
```diff
@@ -27,19 +27,19 @@ classifiers=[
...
+"datasets<=4.0.0",
```
ashokponkumar, line 39: Dont' we want `>=3.5.0` filter?

ashokponkumar (commented):
Nice. Some suggestions.

pyproject.toml
dushyantbehl: sure we can add...

dushyantbehl: we need peft higher version will set accordingly

build/nvcr.Dockerfile
```diff
@@ -0,0 +1,98 @@
...
+# Ensures to always build mamba_ssm from source
+ENV PIP_NO_BINARY=mamba-ssm,mamba_ssm
+
+RUN python -m pip install --upgrade pip
+
+# upgrade torch as the base layer contains only torch 2.7
+RUN pip install --upgrade --force-reinstall torch torchaudio torchvision --index-url https://download.pytorch.org/whl/cu128
+
+# Install main package + flash attention
+RUN git clone --branch ${SOURCE_BRANCH} --depth 1 https://github.com/foundation-model-stack/fms-hf-tuning.git ${SOURCE_DIR}
```
ashokponkumar, line 50: It should not be using git clone, it should be copy from local, since it is from a local build.

build/nvcr.Dockerfile
```diff
@@ -0,0 +1,98 @@
...
+# Set Triton environment variables for qLoRA
+ENV TRITON_HOME="/tmp/triton_home"
+ENV TRITON_DUMP_DIR="/tmp/triton_dump_dir"
+ENV TRITON_CACHE_DIR="/tmp/triton_cache_dir"
+ENV TRITON_OVERRIDE_DIR="/tmp/triton_override_dir"
+
+WORKDIR $WORKDIR
+
+# this is just a dev image so this is okay.
+CMD ["sleep inifinity"]
```
ashokponkumar, line 98:
```suggestion
CMD [ "python", "/app/accelerate_launch.py" ]
```

Why are we using sleep infinity here?

ashokponkumar (dismissed):

build/nvcr.Dockerfile
dushyantbehl: This is a dev image and the command is usually overwritten by the openshift helm chart 

This command you mention below is rarely used so removed it

ashokponkumar (approved):

PR #607, opened 2025-09-15
fix: add default optim arg in training arg

YashasviChaurasia:
### Description of the change
The change defaults the optimizer argument to `adamw_torch` as default for fms-hf-tuning.

The default for hf trainer is `adamw_torch_fused` which wont be compatible with mixed precision, given that we push for mixed precision ="bf16" for performance, I propose we should make `adamw_torch` as default for our stack..
[hf optim](https://huggingface.co/docs/transformers/en/main_classes/trainer#transformers.TrainingArguments.optim)


dushyantbehl (approved):
LGTM

PR #608, opened 2025-09-15
fix: update fms-accel

YashasviChaurasia:
Updates fms-accel optional-dependencies to build from Main as it addresses the following issues:
https://github.com/foundation-model-stack/fms-acceleration/pull/148


pyproject.toml
```diff
@@ -48,7 +48,7 @@ flash-attn = ["flash-attn>=2.8.3"]
 aim = ["aim>=3.19.0,<4.0"]
 mlflow = ["mlflow"]
 clearml = ["clearml==2.0.0"]
-fms-accel = ["fms-acceleration>=0.6"]
+fms-accel = ["fms-acceleration @ git+https://github.com/foundation-model-stack/fms-acceleration.git#subdirectory=plugins/framework"]
```
dushyantbehl, line 51: can you change this to reflect the version we are picking finally instead of main.

dushyantbehl (approved):
LGTM

dushyantbehl (changes requested):
Requesting fms-acceleration tests to be enabled

YashasviChaurasia:
@dushyantbehl 
I can see the tests in `fms-hf-tuning/tests/acceleration` already running with the current Test/build runs..
Is it the same tests you had wanted to be enabled ? if so they are already being tested on..
image

PR #614, opened 2025-09-27
feat: add ckpt conversion script fp32-bf16

YashasviChaurasia:
### Description of the change
fms-hf saves both intermediate and end checkpoints based on settings like (save_strategy, save_model_dir) using HF APIs.
For trainings done on multi node setup with mixed precision (which is default in granite dot build due to better performance) are saved in fp32 mode. This means the checkpoint saved is large in number and even becomes bigger for intermediate checkpoints due to optimizer state.

This PR will add a script to convert model checkpoints from `fp32` to `bf16`, this script can be then used as and when required by to convert checkpoints..

The script supports converting:
- `.pt` / `.pth` files
- single `.safetensors` files
- Hugging Face checkpoint directories (with sharded `.safetensors`)

Additional details:
- Skips optimizer state conversion to preserve FP32 for training stability  
- Preserves non-tensor objects (e.g., JSON configs) and non-FP32 tensors  
- Copies configs, tokenizers, and other non-model files unchanged  

Usage:
```
# Convert .pt/.pth → .pt/.pth
python convert_fp32_to_bf16.py model_fp32.pt model_bf16.pt

# Convert .safetensors → .safetensors
python convert_fp32_to_bf16.py model_fp32.safetensors model_bf16.safetensors

# Convert Hugging Face dir (sharded safetensors) → new dir
python convert_fp32_to_bf16.py ./model_fp32_dir ./model_bf16_dir
```
### How to verify the PR

image
Model checkpoint fine tuned with mixed-precision.  

---

image
Model checkpoint after conversion using the script

### Was the PR tested

### DIFF
--- a/scripts/convert_fp32_to_bf16.py
+++ b/scripts/convert_fp32_to_bf16.py
@@ -1,0 +1,125 @@
+#!/usr/bin/env python3
+# Convert FP32 -> BF16 for .pt/.pth, single .safetensors, or a dir of .safetensors.
+# Skips optimizer states to preserve FP32.
+
+# Standard
+from pathlib import Path
+from typing import Any
+import argparse
+import shutil
+
+# Third Party
+import torch
+
+try:
+    # Third Party
+    from safetensors.torch import safe_open, save_file
+
+    HAS_SAFETENSORS = True
+except ImportError:
+    HAS_SAFETENSORS = False
+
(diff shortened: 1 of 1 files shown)


ashokponkumar:
Can we make it generic as say checkpoint_utils.py or something, and features like removing optimizer files etc.

ashokponkumar (approved):

PR #615, opened 2025-09-29
feat: Allow chat template to be specified via a path in data config.

YashasviChaurasia:
### Description of the change
HF Models now allow users to load chat template via separate chat_template.jinja file which is unlike the chat template inside tokenizer config file. This means users who launch experiments via fms-hf-tuning on models where they want to update chat template need to copy paste it to the data config.

The PR would add support for `chat_template_path` in `data_config.yaml` so users can specify path to chat_template.jinja.
Fixes a few typos along the way.

### Was the PR tested

### DIFF
--- a/tuning/data/data_config.py
+++ b/tuning/data/data_config.py
@@ -150,29 +150,39 @@ def _validate_dataprocessor_config(dataprocessor_config) -> DataPreProcessorConf
         assert isinstance(streaming, bool), f"streaming: {streaming} should be a bool"
         c.streaming = streaming
 
-    provided_chat = "chat_template" in kwargs and kwargs["chat_template"] is not None
-    provided_path = (
+    is_chat_template_present = (
+        "chat_template" in kwargs and kwargs["chat_template"] is not None
+    )
+    is_chat_template_path_present = (
         "chat_template_path" in kwargs and kwargs["chat_template_path"] is not None
     )
-    provided_b64 = (
+    is_chat_template_b64_present = (
         "chat_template_base64" in kwargs and kwargs["chat_template_base64"] is not None
     )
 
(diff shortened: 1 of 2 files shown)


tuning/data/data_config.py
```diff
@@ -148,28 +149,84 @@ def _validate_dataprocessor_config(dataprocessor_config) -> DataPreProcessorConf
         streaming = kwargs["streaming"]
         assert isinstance(streaming, bool), f"streaming: {streaming} should be a bool"
         c.streaming = streaming
-    if "chat_template" in kwargs:
+
+    provided_chat = "chat_template" in kwargs and kwargs["chat_template"] is not None
```
dushyantbehl, line 153: Minor nit...can we rename these variables to

`is_chat_template_present`
`is_chat_template_path_present`
`is_chat_template_b64_present`

tuning/data/data_config.py
```diff
@@ -148,28 +149,84 @@ def _validate_dataprocessor_config(dataprocessor_config) -> DataPreProcessorConf
...
+    provided_path = (
+        "chat_template_path" in kwargs and kwargs["chat_template_path"] is not None
+    )
+    provided_b64 = (
+        "chat_template_base64" in kwargs and kwargs["chat_template_base64"] is not None
+    )
+
+    provided_count = sum([provided_chat, provided_path, provided_b64])
+    if provided_count > 1:
```
dushyantbehl, line 162: Make this single line as the `provided_count` is not used anywhere else.

dushyantbehl (changes requested):
LGTM baaring minor changes.

dushyantbehl (approved):
LGTM.

PR #617, opened 2025-10-08
feat: Adopt resumption feature of online data mixing

kmehant:
## Changes

In this PR, we move construction of the `resume_from_checkpoint` to the top of the function so that it can used in odm_config. Further depends on merge of https://github.com/foundation-model-stack/fms-acceleration/pull/155

### DIFF
--- a/tuning/sft_trainer.py
+++ b/tuning/sft_trainer.py
@@ -129,11 +129,29 @@ def train(
         logger_name="sft_trainer_train", level=train_args.log_level
     )
 
+    resume_from_checkpoint = None
+    # Check if resume flag is not passed (None), or if flag is true and
+    # output_dir has checkpoints then get last checkpoint from output_dir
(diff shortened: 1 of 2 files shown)


tuning/sft_trainer.py
```diff
@@ -129,12 +129,36 @@ def train(
         logger_name="sft_trainer_train", level=train_args.log_level
     )
 
+    resume_from_checkpoint = None
+    if train_args.output_dir:
+        os.makedirs(train_args.output_dir, exist_ok=True)
```
dushyantbehl, line 134: Why do we need to make the output directory doesn't the trainer make it already?

kmehant: related to this failing test - https://github.com/foundation-model-stack/fms-hf-tuning/actions/runs/18341924401

PR #620, opened 2025-10-16
feat: ensure fms-acceleration callbacks run before TrainerController

YashasviChaurasia:
### Description of the change
This PR fixes the callback registration order in sft_trainer.py to ensure that fms-acceleration callbacks are registered before the TrainerController, essentially reordering the callbacks correctly.
This ensures `on_save()` from fms-acceleration executes first in order by registering fms-acceleration callbacks before `TrainerControllerCallback`

This is a required change because the fms-acceleration callbacks handle low-level checkpoint preparation — such as unwrapping model wrappers, merging adapter weights, and finalizing the model state before saving, when registering them before the TrainerControllerCallback, their on_save() runs first, ensuring checkpoints are complete and consistent.

### Was the PR tested

### DIFF
--- a/tuning/sft_trainer.py
+++ b/tuning/sft_trainer.py
@@ -246,15 +246,6 @@ def train(
             trainer_callbacks.append(cb)
             trackers.append(t)
 
-    # Now add trainer controller callbacks if requested
-    if (trainer_controller_args is not None) and (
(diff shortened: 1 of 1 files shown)


dushyantbehl (changes requested):
Requesting clarification on slack on the design before merging

PR #626, opened 2025-11-03
fix: directly save final ckpt in save_model_dir

YashasviChaurasia:
### Description of the change
With this PR the the fms_hf_tuning stack will directly save final ckpt in save_model_dir rather than nesting it further within the `/{save_model_dir}/hf_converted_checkpoint`

When using `SHARDED_STATE_DICT`, each intermediate checkpoint produces distributed shard files that must be converted into Hugging Face–compatible format.
which are saved within
```
{output_dir}/checkpoint-<step>/hf_converted_checkpoint/
```

but for the final checkpoint nesting it again under `hf_converted_checkpoint` is unnecessary and hence the change..

### How to verify the PR
Following is the screenshot of the job with the PR changes.

1. The changes ensures to save hf converted intermediate checkpoint within `hf_converted_checkpoint` subdir
2. While ensuring the final checkpoint directly has hf compatible format without any sub-dir..
image

### Was the PR tested

### DIFF
--- a/tuning/config/acceleration_configs/fast_moe.py
+++ b/tuning/config/acceleration_configs/fast_moe.py
@@ -96,16 +96,21 @@ def get_callbacks(**kwargs):
                 Also saves the final model in save_model_dir if provided.
                 """
 
-                def checkpoint(checkpoint_dir, save_dir):
-                    hf_converted_output_dir = os.path.join(
-                        save_dir, "hf_converted_checkpoint"
-                    )
-                    if os.path.exists(hf_converted_output_dir):
+                def checkpoint(checkpoint_dir, save_dir, is_intermediate: bool = True):
+                    if is_intermediate:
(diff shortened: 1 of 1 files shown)


dushyantbehl (changes requested):
Rather than doing this can we pass the save_dir directly with `hf_converted_output_dir` attached if intermediate else not if this is final from the call to this function?

This would reduce the code and have a clean function.

YashasviChaurasia:
> Can you explain what you mean by ` but I had thought that it could easily lead to inconsistent directory handling.`

With inconsistent I had meant passing different paths outside the checkpoint function, the current version contains all the logic within the checkpoint function and simply abstracts the intermediate vs final checkpoint logic within the function

dushyantbehl (approved):
LGTM

PR #640, opened 2025-11-21
fix: typo in log message about validation dataaset

HarikrishnanBalagopal:
### Was the PR tested


dushyantbehl:
Thanks @HarikrishnanBalagopal

PR #643, opened 2025-11-24
fix: add global step number in hf moe chpt

YashasviChaurasia:
### Description of the change
add global step to checkpoint path for moe models finetuned with sharded_state.  
Example: `hf_converted_checkpoint` -> `safetensors-1234`

### Was the PR tested

### DIFF
--- a/tests/artifacts/language_models/maykeye-tinyllama-v0/config.json
+++ b/tests/artifacts/language_models/maykeye-tinyllama-v0/config.json
@@ -14,7 +14,7 @@
   "num_hidden_layers": 8,
   "pad_token_id": 0,
   "rms_norm_eps": 1e-06,
-  "tie_word_embeddings": false,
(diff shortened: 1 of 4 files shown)


tuning/config/acceleration_configs/fast_moe.py
```diff
@@ -99,7 +99,7 @@ def on_save(
                 def checkpoint(checkpoint_dir, save_dir, is_intermediate: bool = True):
                     if is_intermediate:
                         hf_converted_output_dir = os.path.join(
-                            save_dir, "hf_converted_checkpoint"
+                            save_dir, f"hf_converted_checkpoint-{state.global_step}"
```
dushyantbehl: Can you rename this to `safetensors_converted_checkpoint-{state.global_step}`

if this would require big changes to llmb we can ignore

ashokponkumar (approved):

PR #644, opened 2025-11-25
fix: add fms-recommender wrapper

YashasviChaurasia:
### Description of the change
The PR adds [tuning-config-recommender](https://github.com/foundation-model-stack/tuning-config-recommender) functionality to fms-hf-tuning simplifying the fine-tuning experience for the end user.
 
 This contributes `fms-recommender.py` script which simplifies the tuning experience for the user and helps them avoid minor/major pitfalls which can hinder their tuning experience..
 With the help of `tuning-config-recommender` now we can correct erroneous launch commands or get the recommender to provide suggestions for various flags used while fine-tuning.
 
 The _`fms-recommender.py`_ script has a `--preview` flag which would preview the suggested command and not execute it..
 Note: by default without the `--preview` flag the suggested command is executed !
 
Following is a sample of running the command in preview mode:
 
 ```
python fms-recommender.py --preview tuning.sft_trainer \
--model_name_or_path ibm-granite/granite-4.0-h-350m \
--training_data_path tatsu-lab/alpaca \
--tuning_strategy full
 ```
### Preview command:
image

### Generates a valid `data_config` file with the datasets, chat-template, data_handlers, etc.
image

### Generates a valid `accelerate_config`
image

#### Note:
Currently the [tuning-config-recommender](https://github.com/foundation-model-stack/tuning-config-recommender) is undergoing review cycles before its first release and hence the pyproject.yaml is not linked to a release version..
for now the pyproject would install via ~`main` branch which might not work the same way as shown above~ due to pending essential PR's for FMSAdapter at : [PR's](https://github.com/foundation-model-stack/tuning-config-recommender/pulls) are undergoing review!

The above commands can be replicated via installing recommender from the `main` branch

### Was the PR tested

### DIFF
--- a/scripts/convert_to_hf_checkpoint.py
+++ b/scripts/convert_to_hf_checkpoint.py
@@ -1,13 +1,15 @@
 #!/usr/bin/env python3
+# Standard
+from pathlib import Path
 import argparse
 import os
-from pathlib import Path
 
+# Third Party
+from fms_acceleration_moe.utils import (  # pylint: disable=import-error
+    recover_safetensors_from_dcp,
+)
 from huggingface_hub import snapshot_download
 from transformers import AutoConfig, AutoTokenizer
-
-from fms_acceleration_moe.utils import recover_safetensors_from_dcp
-
 
 HF_CACHE = "/workspace/.hf"
 os.environ.setdefault("HF_HOME", HF_CACHE)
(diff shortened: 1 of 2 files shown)


YashasviChaurasia:
updated pyproject to use `main` branch until first release.
major changes were merged into main for tuning-config-recommender.

PR #651, opened 2026-01-13
feat: Added dataset split support

### LINKED ISSUE #645: Choose dataset split to use in data_config

Currently we can only choose `train` split for our dataset using fms-hf..
Would request ability to load other splits of our dataset as main split.

key in data_config which holds value to the split name to load for the dataset.

## Additional context

`tuning-config-recommender` has ability to decide and check data object keys from other splits to decide on various fields like `conversation_columns` but the output of the tuning-config-recommender is not useful as fms-hf does not yet support loading dataset from other splits.

romitjain:
### Description of the change

This PR adds

1. Support for custom dataset split name - this is useful when loading datasets directly from HuggingFace
2. Support for disabling datset shuffling - this can be useful when you are using a pre-shuffled or pre-processed dataset where you want to maintain the order of the samples
3. Sets seed for accelerate

### Related issue number

Closes #645 

### How to verify the PR

```
pytest tests/test_sft_trainer.py::test_run_chat_style_ft_using_custom_split_nam
```

### Was the PR tested

- [X] I have added >=1 unit test(s) for every new method I have added.
- [X] I have ensured all unit tests pass

### DIFF
--- a/tuning/data/data_processors.py
+++ b/tuning/data/data_processors.py
@@ -112,9 +112,9 @@ class DataPreProcessor:
         if (not datafile) and (not datasetconfig):
             raise ValueError("Either datafile or datasetconfig must be set")
 
-        effective_split = (
-            splitName or datasetconfig.dataset_split_name if datasetconfig else None
-        )
+        effective_split = splitName
+        if datasetconfig and datasetconfig.dataset_split_name:
+            effective_split = datasetconfig.dataset_split_name
 
         def _load_dataset(
             data_path=None,


tuning/sft_trainer.py
```diff
@@ -501,6 +502,9 @@ def train(
         peft_config=peft_config,
     )
 
+    # Set seed for accelerate processes
```
dushyantbehl: Thanks for adding this.

tuning/data/data_processors.py
```diff
@@ -471,7 +472,9 @@ def _prepare_processed_datasets(
             logger.info("Loading the dataset - %s", d.name)
 
             # In future the streaming etc go as kwargs of this function
-            loaded_dataset = self.load_dataset(d, self.processor_config.streaming)
+            loaded_dataset = self.load_dataset(
+                d, self.processor_config.streaming, splitName=d.dataset_split_name
```
dushyantbehl: We are already passing the dataset config so do we need to pass the split name separately?

dushyantbehl (commented):
Thanks @romitjain requesting minor changes and clarification but largely looks good to me.

tuning/data/data_processors.py
dushyantbehl: can you remove the splitName parameter from test cases and use your way of passing it inside data config...that could be made standard now. 
Feel free to update the load_dataset signature and that should be the final change we need.

dushyantbehl (changes requested):
Thanks for the documentation and getting all the tests to run. 

minor change requested in - https://github.com/foundation-model-stack/fms-hf-tuning/pull/651/changes#r2767607777

Also please sign the commit, DCO check is failing.

tuning/data/data_processors.py
romitjain: Sure, will do that

PR #652, opened 2026-01-19
fix: only trigger on_save if the output checkpoint path actually contains files

HarikrishnanBalagopal:
### Description of the change

Trigger `on_save` event only if anything was actually saved in the output checkpoint directory.
This simplifies the use of trainer_controller for triggering subsequent steps (like artifact upload)

### Related issue number

Internal issue in our pipeline

### Was the PR tested

### DIFF
--- a/tuning/sft_trainer.py
+++ b/tuning/sft_trainer.py
@@ -576,10 +576,36 @@ def save(path: str, trainer: SFTTrainer, tc_callback, log_level="WARNING", args=
 
     logger.info("Saving tuned model to path: %s", path)
     trainer.save_model(path)
+    actually_saved = False
     if tc_callback and args:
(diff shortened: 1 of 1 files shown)


HarikrishnanBalagopal:
> @HarikrishnanBalagopal I don't think this is the right fix. If save_ model is called, and files are not there, its a problem. Trainer should be fixed rather.
> 
> @dushyantbehl @YashasviChaurasia Can you look at the situations where this occurs?

@ashokponkumar The original issue has already been fixed in the `main` branch and we are now using the image built from `main` branch commit `b3a8a78d5a1e81635a974934cb1b08a3a5361b3e` that contains the fix.

This is just an additional/sanity check on top since `on_save` should not be called if nothing was saved.

